PR #4, opened 2019-07-01
Test docker-compose files on Travis

tijmenb:
This adds a travis config file and test task to the Makefile. This allows us to run tests for every pull request.

I wanted to run this on our own CI, but we're using an old version of docker/docker-compose that won't accept our newer docker-compose syntax.

https://trello.com/c/B0TSiqy5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,14 +19,15 @@ pull:
 	done
 
 build:
-	govuk-docker build
+	bin/govuk-docker build
 
 setup:
 	for repo in $(REPOS); do \
 		make -f $$repo/Makefile; \
 	done
-	govuk-docker run whitehall-e2e rake taxonomy:populate_end_to_end_test_data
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,14 +19,19 @@ pull:
 	done
 
 build:
-	govuk-docker build
+	bin/govuk-docker build
```
benthorner: Should we make this consistent in all the project specific Makefiles?

kevindew:
You should be able to bump the version of compose on CI. It's only pinned to 1.17.1 as (if I remember right) it needed a version specified to update: https://github.com/alphagov/govuk-puppet/commit/1fe7f22e5dadd52f62c9919132a99f0587a4ba88

Makefile
tijmenb: Yeah that would be good. @barrucadu can you sneak that into https://github.com/alphagov/govuk-docker/pull/9?

PR #6, opened 2019-07-01
Allow specifying which repos to clone/pull

barrucadu:
Variables can be specified on the command line like so:

    make clone REPOS=email-alert-api

And that will only clone email-alert-api.  Multiple values can be
specified with spaces:

    make clone REPOS="email-alert-api content-publisher"


benthorner:
I've run this locally, but it doesn't seem to work when no variable is specified i.e. `make pull` does nothing.

PR #7, opened 2019-07-01
Added: run dig to check name resolution

VitalieMogoreanu:


README.md
```diff
@@ -91,6 +91,12 @@ sudo brew services restart dnsmasq
 
 To check if the new config has been applied, you can run `scutil --dns` to check that `dev.gov.uk` appears in the list.
 
+To check name resolution run `dig app.dev.gov.uk @127.0.0.1`. The response has to include answer section
```
vanitabarrett:
```suggestion
To check name resolution run `dig app.dev.gov.uk @127.0.0.1`. The response has to include the following answer section:
```

benthorner:
Might also be worth clarifying why we need both the `scutils` check and the `dig` check?

README.md
```diff
@@ -70,17 +70,20 @@ of  `/etc/resolver/dev.gov.uk`.
 cp /etc/resolver/dev.gov.uk ~/dev.gov.uk
 ```
 
-Then create or append to the following and restart dnsmasq. If you've been using
-the vagrant based dev vm, you'll need to replace `/etc/resolver/dev.gov.uk`..
```
vanitabarrett: @VitalieMogoreanu I think we want to keep this line about replacing the file if you've been using the VM?

PR #8, opened 2019-07-01
Add Email Alert API

karlbaker02:
This PR adds the Email Alert API as a Docker image and runnable via docker-compose.

Screen Shot 2019-07-02 at 10 36 56

Screen Shot 2019-07-02 at 10 37 13

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ The following apps are supported by govuk-docker to some extent.
    - ✅ content-publisher
    - ✅ content-store
    - ✅ content-tagger
+   - ✅ email-alert-api
    - ✅ government-frontend
    - ✅ govspeak
    - ✅ govuk-developer-docs
--- a/services/email-alert-api/Dockerfile
+++ b/services/email-alert-api/Dockerfile
@@ -1,0 +1,8 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim && apt-get clean
+
+RUN useradd -m build
+USER build
+
+ENV EDITOR=vim
--- a/services/email-alert-api/Makefile
+++ b/services/email-alert-api/Makefile
@@ -1,0 +1,3 @@
+email-alert-api: ../email-alert-api
+	bin/govuk-docker run email-alert-api-default bundle
+	bin/govuk-docker run email-alert-api-default sh -c 'rake db:migrate 2>/dev/null || rake db:setup'
--- a/services/email-alert-api/docker-compose.yml
+++ b/services/email-alert-api/docker-compose.yml
@@ -1,0 +1,46 @@
+version: '3.7'
+
+x-email-alert-api: &email-alert-api
+  build:
+    context: .
+    dockerfile: email-alert-api/Dockerfile
+  image: email-alert-api
+  volumes:
+    - ..:/govuk:delegated
+    - bundle:/usr/local/bundle
+  working_dir: /govuk/email-alert-api
+
+services:
+  email-alert-api-default:
+    <<: *email-alert-api
+    depends_on:
+      - postgres
+      - redis
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres/email-alert-api"
+      TEST_DATABASE_URL: "postgresql://postgres@postgres/email-alert-api-test"
+      REDIS_URL: redis://redis
+
+  email-alert-api-backend: &email-alert-api-backend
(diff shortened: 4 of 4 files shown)


email-alert-api/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM ruby:2.6.3
+
```
benthorner: For consistency, it would be good to have `vim` installed here and to set the EDITOR env var - the usage is to enable the `bundle open` command that would normally be available.

email-alert-api/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+      REDIS_URL: redis://redis
+
+  email-alert-api-backend: &email-alert-api-backend
+    <<: *email-alert-api
+    depends_on:
+      - postgres
+      - redis
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres/email-alert-api"
+      TEST_DATABASE_URL: "postgresql://postgres@postgres/email-alert-api-test"
```
benthorner: Oh, forgot that we shouldn't need to set this in other stacks, since we don't need a test DB.

karlbaker02: Good catch

karlbaker02:
For testing the backend stack, do you mean some kind of visual indication of it working as Leena has in her PR? I can definitely do this for the healthcheck / main entry page / potentially an API endpoint given some existing seed data 🙂

benthorner:
@karlbaker02 yeah, something like that. It means someone can look at the PR and quickly get an idea of the level of support/testing we did when a particular stack was added for an app. The Support app is one example where the more you try and use it, the more dependencies you uncover - doesn't mean we should spend time fully testing each app, but it does help to demo what we have tested.

In this particular PR, we've also got a worker, so another thing that would be good to do is to find a way to get the app to make use of it when running the E2E stack. If we can't do that with this app running in isolation, then it might be better to leave it out until we have a cross-app E2E stack that can make it do something, where we can see that it does indeed work.

karlbaker02:
@barrucadu I'm going to remove e2e as I don't think this can be properly tested without data for lists  / subscriptions etc

barrucadu:
The bad command is actually in the `backend` stack

email-alert-api/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential vim
```
kevindew: I don't think we have build-essential in any of these any more since: https://github.com/alphagov/govuk-docker/pull/11

email-alert-api/Makefile
```diff
@@ -0,0 +1,4 @@
+email-alert-api: ../email-alert-api
+	bin/govuk-docker run email-alert-api-default bundle
+	bin/govuk-docker run email-alert-api-default sh -c 'rake db:migrate 2>/dev/null || rake db:setup'
+	bin/govuk-docker run email-alert-api-default bundle exec rake
```
kevindew: I don't think we want rake here as that would run the test suite. I think we can just remove that line as we just need the former 2.

email-alert-api/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    depends_on:
+      - postgres
+      - redis
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres/email-alert-api"
+      REDIS_URL: redis://redis
+      VIRTUAL_HOST: email-alert-api.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
```
kevindew: We don't need HOST, VIRTUAL_HOST, or ports for the worker since it's not a web process.

email-alert-api/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    command: bundle exec sidekiq -C ./config/sidekiq.yml
```
kevindew: I'd suggest we still have an e2e entry for consistency, however I think the only additional dependency that would have to backend is email-alert-api-worker.

kevindew:
Oh @benthorner I see I contradicted your comment about e2e. Shouldn't we still have one so we have something with a dependency on the worker?

karlbaker02:
@barrucadu I added this project before everything was moved into `services` - this has now been fixed. @benthorner @kevindew any updates on what to do about e2e?

PR #9, opened 2019-07-01
Allow running setup task for individual apps and their dependencies

barrucadu:
Setting up all the apps is slow, so we want to be able to just set up some of them.  It's also good to be able to set up all the dependent apps, so if you set up content-publisher and then start it, it doesn't immediately stop working because you haven't installed gems for the publishing-api (eg).

So now to get everything you need for content-publisher, and nothing else, you can run

    make APPS=content-publisher

This makes a fairly substantial change to the way the project makefiles are structured, pulling them all into the top-level makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ GOVUK_ROOT_DIR="${HOME}/govuk"
 
 APPS ?= $(shell ls */Makefile | xargs -L 1 dirname)
 
-default: clone build setup clean
+default: build setup clean
 
 clone: $(addprefix ../,$(APPS))
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 GOVUK_ROOT_DIR="${HOME}/govuk"
 
+.PHONY: clone pull build setup clean $(shell ls */Makefile | xargs -L 1 dirname)
+
 APPS ?= $(shell ls */Makefile | xargs -L 1 dirname)
 
 default: build setup clean
@@ -16,7 +18,7 @@ pull:
 build:
 	bin/govuk-docker build
 
-setup: $(addsuffix _setup,$(APPS))
+setup: $(APPS)
 
 clean:
 	bin/govuk-docker stop
@@ -27,9 +29,9 @@ test:
 	# in the YAML files, or incompatible features are used.
 	bin/govuk-docker config
 
-../%: % %/Makefile
-	if [ ! -d "${GOVUK_ROOT_DIR}/$<" ]; then \
-		echo "$<" && git clone "git@github.com:alphagov/$<.git" "${GOVUK_ROOT_DIR}/$<"; \
+../%: %/Makefile
+	if [ ! -d "${GOVUK_ROOT_DIR}/$(subst /Makefile,,$<)" ]; then \
+		echo "$(subst /Makefile,,$<)" && git clone "git@github.com:alphagov/$(subst /Makefile,,$<).git" "${GOVUK_ROOT_DIR}/$(subst /Makefile,,$<)"; \
 	fi
 
 include $(shell ls */Makefile)
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
 GOVUK_ROOT_DIR="${HOME}/govuk"
 
-.PHONY: clone pull build setup clean $(shell ls */Makefile | xargs -L 1 dirname)
+.PHONY: clone pull build clean test $(shell ls */Makefile | xargs -L 1 dirname)
 
 APPS ?= $(shell ls */Makefile | xargs -L 1 dirname)
 
-default: build setup clean
+default:
+	@echo "Run 'make build' to bootstrap govuk-docker"
+	@echo "Or 'make APP-NAME' to set up an app"
 
 clone: $(addprefix ../,$(APPS))
 
@@ -17,8 +19,6 @@ pull:
 
 build:
 	bin/govuk-docker build
-
-setup: $(APPS)
 
 clean:
 	bin/govuk-docker stop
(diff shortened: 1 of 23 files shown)


Makefile
```diff
@@ -21,12 +21,11 @@ pull:
 build:
 	govuk-docker build
 
-setup:
-	for repo in $(REPOS); do \
-		make -f $$repo/Makefile; \
-	done
+setup: $(addsuffix _setup,$(REPOS))
 	govuk-docker run whitehall-e2e rake taxonomy:populate_end_to_end_test_data
```
barrucadu: This command fails if `REPOS` has been specified and doesn't contain everything - should it be a separate task?

benthorner:
Looks good - it might be nice to stay without the '_setup' suffix for a while, just for simplicity until we know we need something else.

benthorner:
I realised a missed something yesterday: the way the rules are now could mean we do setup multiple times for things like Asset Manager. Do we actually need two rules for each service: one to make the thing in isolation, and one to do it with dependencies?

barrucadu:
Is doing setup multiple times a problem?  It's not like it'll install all the gems again.

benthorner:
Yeah - it's expensive to run each govuk-docker command e.g. ones that try to setup the DB. I think it should be pretty easy to have two rules to avoid any repetition.

barrucadu:
Could we create a hidden file indicating that setup has been done for an app or something?  Having two sets of tasks means you need to know which to use.

barrucadu:
We decided to remove the top-level setup task because we're not sure why someone would now want to set-up everything, and it's easier to just remove it than to implement a work-around to avoid re-running app setups.

benthorner:
This is looking good. A couple of suggestions that could make it better:

   - Could we explain a bit what the .PHONY stuff is and the ../% rule. Might just be me, but looks tricky to understand.
   - I think we should remove the 'build' and 'test' steps, as the former will happen automatically and the latter just repeats the former.
 
More of a stretch thing, but I'm also wondering if `make app-name` would be more helpful if it did a pull if the repo already exists?

barrucadu:
> Could we explain a bit what the .PHONY stuff is and the ../% rule. Might just be me, but looks tricky to understand.

Ok, I'll open another PR with a couple of comments

> I think we should remove the 'build' and 'test' steps, as the former will happen automatically and the latter just repeats the former.

Does attempting to use an image which hasn't been built yet cause it to be built?  If so, yes, we can remove `build`.  I'm not sure `test` is redundant though, as it's used in the CI.

> More of a stretch thing, but I'm also wondering if `make app-name` would be more helpful if it did a pull if the repo already exists?

I don't think I'd like code I have locally to be updated unless I specifically ask for that.

benthorner:
Thanks @barrucadu :-). Build is indeed automatic, and since we're no longer looking to setup everything, we don't need to build everything either. I think I'm just not understanding what the '.PHONY' bit is for - we still need the task itself, I agree.

Thinking about the 'pull' idea a bit more, I think I agree with you - doing a global pull is a bit dodgy too, as it sometimes interacts badly with any local changes.

barrucadu:
`.PHONY` is so this doesn't happen:

```
bash-3.2$ touch clean
bash-3.2$ make clean
make: `clean' is up to date.
```

Though it seems it's not actually needed for the apps, I guess directories are handled differently:

```
bash-3.2$ file content-publisher
content-publisher: directory
bash-3.2$ make content-publisher
if [ ! -d ""/Users/michaelswalker/govuk"/publishing-api" ]; then \
                echo "publishing-api" && git clone "git@github.com:alphagov/publishing-api.git" ""/Users/michaelswalker/govuk"/publishing-api"; \
        fi
if [ ! -d ""/Users/michaelswalker/govuk"/content-store" ]; then \
                echo "content-store" && git clone "git@github.com:alphagov/content-store.git" ""/Users/michaelswalker/govuk"/content-store"; \
        fi
if [ ! -d ""/Users/michaelswalker/govuk"/router-api" ]; then \
                echo "router-api" && git clone "git@github.com:alphagov/router-api.git" ""/Users/michaelswalker/govuk"/router-api"; \
        fi
bin/govuk-docker run router-default sh -c "BINARY=router make build"
+ docker-compose -f bin/../docker-compose.yml -f bin/../asset-manager/docker-compose.yml -f bin/../content-data-admin/docker-compose.yml -f bin/../content-publisher/docker-compose.yml -f bin/../content-store/docker-compose.yml -f bin/../content-tagger/docker-compose.yml -f bin/../government-frontend/docker-compose.yml -f bin/../govspeak/docker-compose.yml -f bin/../govuk-developer-docs/docker-compose.yml -f bin/../govuk-lint/docker-compose.yml -f bin/../govuk_app_config/docker-compose.yml -f bin/../govuk_publishing_components/docker-compose.yml -f bin/../miller-columns-element/docker-compose.yml -f bin/../nginx-proxy/docker-compose.yml -f bin/../plek/docker-compose.yml -f bin/../publishing-api/docker-compose.yml -f bin/../router-api/docker-compose.yml -f bin/../router/docker-compose.yml -f bin/../signon/docker-compose.yml -f bin/../static/docker-compose.yml -f bin/../support-api/docker-compose.yml -f bin/../support/docker-compose.yml -f bin/../whitehall/docker-compose.yml run router-default sh -c 'BINARY=router make build'
^C
```

benthorner:
I'm guessing the need for this is related to the '../%' rule - is that what's doing 'touch clean' kind of things?

barrucadu:
No, it seems it's not actually needed for the apps at all (though `.PHONY: clone pull build clean test` is still useful for that reason), I was just overzealous in adding it.

PR #14, opened 2019-07-02
Remove 'fleshclam' command from Dockerfile

benthorner:

Previously we ran 'freshclam' to update virus definitions for Asset
Manager, but this takes a while a doesn't seem to have any impact on
the day-to-day tasks. This removes the command from the Dockerfile -
we may want to support updating virus definitions more regularly and
easily if this is a day-to-day concern.

### DIFF
--- a/asset-manager/Dockerfile
+++ b/asset-manager/Dockerfile
@@ -1,6 +1,5 @@
 FROM ruby:2.6.3
 RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y clamav vim
-RUN freshclam
 RUN ln -sf /usr/bin/clamscan /usr/bin/govuk_clamscan
 
 RUN useradd -m build


benthorner:
@kevindew perhaps worth double-checking this is unnecessary now that you've got that final Asset Manager test working?

kevindew:
@benthorner asset manager will still be calling out to govuk_clamscan - but don't think there is any need for it to be up to date, unless Clam will moan if it gets too far out of date. Though given how it's built and docker layers there isn't anything to keep it up date, even with the freshclam command.

One option I recall that was being considered with end-to-end tests and might be better in a dev environment is to set govuk_clamscan to just return an exit code of 0 whatever the input. I have a feeling we have code for that somewhere but I can't remember where.

benthorner:
@kevindew I think it's here: https://github.com/alphagov/publishing-e2e-tests/blob/master/docker-compose.yml#L885. But I'm cautious about making the dev environment unrealistic in this way, although that's just my superficial opinion that a dev on Asset Manager would want to be working with a real virus scanner.

kevindew:
> @kevindew I think it's here: https://github.com/alphagov/publishing-e2e-tests/blob/master/docker-compose.yml#L885. But I'm cautious about making the dev environment unrealistic in this way, although that's just my superficial opinion that a dev on Asset Manager would want to be working with a real virus scanner.

Yup, understandable - pros and cons either way.

PR #17, opened 2019-07-02
Add link-checker-api

sihugh:
Screenshot 2019-07-02 10 22 59

This also adds the link checker api worker as a dependency of the backend app which seems to work
pretty well.

running:

 `curl -s http://link-checker-api.dev.gov.uk/check\?uri\=https%3A%2F%2Fwww.gov.uk%2F`

gives an initial response of:

> {"uri":"https://www.gov.uk/","status":"pending","checked":null,"problem_summary":null,"errors":[],"warnings":[],"suggested_fix":null}%

once the worker has run, repeating the request gives the response:

> {"uri":"https://www.gov.uk/","status":"ok","checked":"2019-07-02T10:04:44Z","problem_summary":null,"errors":[],"warnings":[],"suggested_fix":null}%

We don't think we need an e2e option right now.
https://trello.com/c/CrqFTwwu/35-set-up-link-checker-api

### DIFF
--- a/link-checker-api/Dockerfile
+++ b/link-checker-api/Dockerfile
@@ -1,0 +1,8 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim && apt-get clean
+
+RUN useradd -m build
+USER build
+
+ENV EDITOR=vim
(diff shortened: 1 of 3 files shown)


sihugh:
We've updated the makefile format and tweaked the dependencies. 👍

PR #18, opened 2019-07-02
Add travel-advice-publisher

brucebolt:
Adds the travel advice publisher app

Screen Shot 2019-07-02 at 10 46 35

Trello card: https://trello.com/c/7MGgrIwa/26-set-up-travel-advice-publisher

### DIFF
--- a/travel-advice-publisher/Dockerfile
+++ b/travel-advice-publisher/Dockerfile
@@ -1,0 +1,21 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim && apt-get clean
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && apt install -y ./google-chrome*.deb
+# for capybara-webkit
+RUN apt-get update -qq && apt-get install -y libqtwebkit4 libqt4-dev xvfb
+RUN gem install foreman
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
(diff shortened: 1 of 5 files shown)


travel-advice-publisher/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential vim && apt-get clean
```
benthorner: Minor thing: we've removed 'build-essential' as it turns out to be redundant.

travel-advice-publisher/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    depends_on:
+      - mongo
+      - redis
+      - publishing-api-backend
+      - asset-manager-backend
+      - content-store-live
+      - nginx-proxy
+    environment:
+      MONGODB_URI: "mongodb://mongo/travel-advice-publisher"
+      TEST_MONGODB_URI: "mongodb://mongo/travel-advice-publisher-test"
```
benthorner: Minor, but we shouldn't need this particular env var outside of the default stack, since we're (hopefully) not running the tests here.

travel-advice-publisher/Makefile
```diff
@@ -0,0 +1,3 @@
+travel-advice-publisher: ../travel-advice-publisher
```
barrucadu: `asset-manager content-store publishing-api` should be added to the dependencies

brucebolt:
Makefile updated and new dependencies added

PR #19, opened 2019-07-02
Clarify some bits of the top-level makefile

barrucadu:


benthorner:
Thanks @barrucadu. Just to help my understanding:

   - Is the .PHONY target mostly a best-practice thing? I'm not particularly familiar with Makefiles, and I'm guessing other people might wonder why it's there with the current comment.

   - Reading the comment for the '../%' rule, I'm wondering if the Makefile guard clause is necessary? Also what the '..' means - possibly this is just knowledge of make that I'm missing?

barrucadu:
> Is the .PHONY target mostly a best-practice thing? I'm not particularly familiar with Makefiles, and I'm guessing other people might wonder why it's there with the current comment.

Yes, it's just a best practice thing.  Given that it's not needed for the apps, it's not actually essential here.

> Reading the comment for the '../%' rule, I'm wondering if the Makefile guard clause is necessary?

I guess it's optional, but I think the error message is nicer with the guard.  At the moment, cloning a nonexistent repository is like this:

```
bash-3.2$ make ../foo-bar-baz
make: *** No rule to make target `../foo-bar-baz'.  Stop.
```

Without the guard we get

```
bash-3.2$ make ../foo-bar-baz
if [ ! -d ""/Users/michaelswalker/govuk"/foo-bar-baz" ]; then \
                echo "foo-bar-baz" && git clone "git@github.com:alphagov/foo-bar-baz.git" ""/Users/michaelswalker/govuk"/foo-bar-baz"; \
        fi
foo-bar-baz
Cloning into '/Users/michaelswalker/govuk/foo-bar-baz'...
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
make: *** [../foo-bar-baz] Error 128
bash-3.2$ echo $?
2

```

> Also what the '..' means - possibly this is just knowledge of make that I'm missing?

That's just a relative path, `%` is a wildcard, so `../%` matches anything in the parent directory.

benthorner:
Aha, I think I'm understanding now. So '..' is necessary because we only want to run the clone for repo x if '../x'  (i.e. ~/govuk/x) isn't present. This makes me wonder why we still have an 'if' in the rule?

Also, it looks like you can use '%' in the definition of the rule, which I think would avoid the 'subst' stuff and make the definition clearer...

https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html#Automatic-Variables

barrucadu:
Without the `if` we get this:

```
bash-3.2$ make ../content-store
echo "content-store" && git clone "git@github.com:alphagov/content-store.git" ""/Users/michaelswalker/govuk"/content-store"
content-store
fatal: destination path '/Users/michaelswalker/govuk/content-store' already exists and is not an empty directory.
make: *** [../content-store] Error 128
```

Based on this and the `.PHONY` thing, I suspect `make` doesn't treat directories as satisfying targets.

benthorner:
Ah, so does that mean the '../%' is somewhat arbitrary?

barrucadu:
I guess so, it could equally well be `%_clone`, but I thought given that the rule does produce a thing on the filesystem, that should be the target name.

PR #21, opened 2019-07-02
Add support for custom apps home directory

DilwoarH:
This allows users with a different root dir to override the standard path

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-GOVUK_ROOT_DIR="${HOME}/govuk"
+GOVUK_ROOT_DIR ?= "${HOME}/govuk"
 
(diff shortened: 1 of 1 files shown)


DilwoarH:
@barrucadu - updated

Makefile
```diff
@@ -1,4 +1,4 @@
-GOVUK_ROOT_DIR="${HOME}/govuk"
+GOVUK_ROOT_DIR ?= "${GOVUK_APPS_HOME}/govuk"
```
barrucadu: `GOVUK_APPS_HOME` is no longer being set, that should be `HOME`

DilwoarH: Good spot, updated now

PR #23, opened 2019-07-02
Add setup and setup-this subcommands

barrucadu:
'setup' takes a list of app names, so you can do this:

    govuk-docker setup content-publisher government-frontend

### DIFF
--- a/bin/govuk-docker
+++ b/bin/govuk-docker
@@ -48,11 +48,6 @@ case $COMMAND in
     set -x
     cd $GOVUK_DOCKER_DIR
     make $SERVICE;;
-  "build-this")
-    check_service
-    set -x
-    docker-compose $COMPOSE_FLAGS build $SERVICE-default
-    ;;
   "run-this")
     STACK=$1
     shift
--- a/bin/govuk-docker
+++ b/bin/govuk-docker
@@ -3,6 +3,8 @@
 COMMAND=$1
 shift
 
+GOVUK_DOCKER_DIR=${GOVUK_ROOT_DIR:-$HOME/govuk}/govuk-docker
+
 COMPOSE_FILES="$(dirname $0)/../services/*/docker-compose.yml"
 COMPOSE_FLAGS="-f $(dirname $0)/../docker-compose.yml"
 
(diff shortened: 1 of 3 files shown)


bin/govuk-docker
```diff
@@ -32,6 +34,20 @@ function check_stack {
...
+    set -x
+    cd $GOVUK_DOCKER_DIR
+    for GOVUK_DOCKER_SERVICE in "$@"; do
+      make $GOVUK_DOCKER_SERVICE
+    done;;
+  "setup-this")
+    check_service
+    set -x
+    cd $GOVUK_DOCKER_DIR
+    make $SERVICE;;
```
benthorner: Is that a typo here?

barrucadu: Where?

barrucadu:
I think there should be a command to set up an app by name, because otherwise you need to be in a directory with the right name, or be in the govuk-docker directory and run `make`.  And it's simple enough to make that command take a list of arguments.

tijmenb:
What would be the difference between `build` and `setup`?

barrucadu:
`build` just does the docker compose bit, `setup` clones the repositories, does the build, and runs setup tasks.  So `build` is a subset of `setup`.

tijmenb:
Worth documenting when to use which one! I'm getting a bit confused already.

benthorner:
Yeah. I think we can remove 'build-this'. We already have a command to setup an app by name - `make app-name` - so I'm keen that we only add one thing to the CLI to solve a given problem. The CLI still worries me, as it's only meant to be a lightweight wrapper - the more stuff we add - the more confusing the environment is likely to become, as it'll become harder to differentiate between docker/compose/make/CLI.

bin/govuk-docker
benthorner: We seem to have two semicolons - is that necessary?

barrucadu: Yes, it's how to end a `case` branch (the other branches do this too)

barrucadu:
> We already have a command to setup an app by name - `make app-name` - so I'm keen that we only add one thing to the CLI to solve a given problem

But that requires you to be in the govuk-docker directory, whereas this just requires `govuk-docker` to be in your path.

benthorner:
That's true. What I'm trying to get at, which should also be in the PR description, is: what specific problem(s) / need(s) are we trying to address/support? 

If we're clear about that, it would be easier to understand the change. Currently, this PR adds 3 new commands, if you count the looping behaviour:

   - `govuk-docker setup app-name`
   - `govuk-docker setup app-name...`
   - `govuk-docker setup-this`

It would be helpful if we could document the need devs have for each of these three variants, so that future devs will understand and preserve these commands.

barrucadu:
This is outdated as the `govuk-docker` script has been rewritten in Ruby

PR #27, opened 2019-07-02
Remove the downloaded google-chrome deb package after installation

cbaines:
As this seems neater than leaving it around.


cbaines:
> Could we do this for phantomjs as well?

I believe so.

PR #32, opened 2019-07-02
Remove `make build` from Makefile usage message

barrucadu:
Also clarify `make APP-NAME` pulls in dependencies too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,10 @@ GOVUK_ROOT_DIR ?= "${HOME}/govuk"
 APPS ?= $(shell ls services/*/Makefile | xargs -L 1 dirname)
 
 default:
-	@echo "Run 'make build' to bootstrap govuk-docker"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,8 +13,7 @@ GOVUK_ROOT_DIR ?= "${HOME}/govuk"
 APPS ?= $(shell ls services/*/Makefile | xargs -L 1 dirname)
 
 default:
-	@echo "Run 'make build' to bootstrap govuk-docker"
-	@echo "Or 'make APP-NAME' to set up an app"
+	@echo "Run 'make APP-NAME' to set up an app and its dependencies"
```
DilwoarH: Should we put an example ie. `make content-publisher` or is it obvious to first timers?

barrucadu: Sure, I can add an example

DilwoarH:
🥇

PR #37, opened 2019-07-02
Use specific rake rails

benthorner:
https://trello.com/c/yj0PUGVS/20-asset-manager-rake-error

This ensures we run 'rake' and 'rails' in the context of the local bundle for each service. Not doing this has caused intermittent errors for people running `make app-name` tasks.

It's also a piece of left-over debt from this decision: https://trello.com/c/yj0PUGVS/20-asset-manager-rake-error.

### DIFF
--- a/services/asset-manager/Makefile
+++ b/services/asset-manager/Makefile
@@ -1,4 +1,4 @@
 asset-manager: $(GOVUK_ROOT_DIR)/asset-manager
 	$(COMPOSE_RUN) asset-manager-default bundle
-	$(COMPOSE_RUN) asset-manager-default sh -c 'rake db:migrate 2>/dev/null || rake db:setup'
-	$(COMPOSE_RUN) asset-manager-default rails runner 'User.first || User.create'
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ GOVUK_ROOT_DIR ?= "${HOME}/govuk"
 #     $ touch clean
 #     $ make clean
 #     make: `clean' is up to date.
-.PHONY: clone pull clean test
+.PHONY: clone pull clean test setup-all
```
sihugh: should this be `setup`?

benthorner:
Good spot @sihugh - fixed :-).

benthorner:
@tijmenb that looks like it's more of a stylistic thing. After trying it for a while, I ended up going back to 'rake', which is the framework we're trying to use anyway and makes more sense to me - rails does too much.

tijmenb:
@benthorner this needs a few merge conflict fixes.

benthorner:
@tijmenb done - hopefully the last time now :-).

PR #40, opened 2019-07-02
Set up calendars with docker

1pretz1:
To populate content-store and draft-content-store with calendars, this
rake task needs to be run using the setup service -
https://github.com/alphagov/calendars/blob/master/lib/tasks/publishing_api.rake.

`govuk-docker run-this setup rake publishing_api:publish`

## Running the tests
Screen Shot 2019-07-02 at 18 09 05

## Draft stack
Screen Shot 2019-07-02 at 18 08 38

## Live stack
Screen Shot 2019-07-02 at 18 07 17

Trello:
https://trello.com/c/sm2rw3Nr/27-set-up-calendars

### DIFF
--- a/services/calendars/docker-compose.yml
+++ b/services/calendars/docker-compose.yml
@@ -1,0 +1,49 @@
+version: '3.7'
+
+x-calendars: &calendars
+  build:
+    context: .
+    dockerfile: calendars/Dockerfile
+  image: calendars
+  volumes:
+    - ..:/govuk:delegated
+    - bundle:/usr/local/bundle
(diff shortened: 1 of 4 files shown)


services/calendars/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
+version: '3.7'
+
+x-calendars: &calendars
+  build:
+    context: .
+    dockerfile: calendars/Dockerfile
```
tijmenb: I think this needs to be `services/calendars/Dockerfile` now.

PR #41, opened 2019-07-02
Add smart-answers

leenagupte:
Trello: https://trello.com/c/tQL1TvP8

It seems that smart-answer needs both Google Chrome and PhantomJS to run it's tests:
https://github.com/alphagov/smart-answers/blob/b63316766b31064e7ccfa78fbcffcfcb90c6f72d/doc/smart-answers-app-development/testing.md#external-dependencies

# Publishing smart-answers

Run:  `govuk-docker run rake publishing_api:publish`

# Rendering smart-answers

Run:  `govuk-docker startup`

http://smart-answers.dev.gov.uk/pay-leave-for-parents/y

Screen Shot 2019-07-04 at 12 57 16

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ The following apps are supported by govuk-docker to some extent.
    - ✅ router
    - ✅ router-api
    - ✅ signon
+   - ✅ smart-answers
    - ✅ specialist-publisher
    - ⚠ static
       * JavaScript 404 errors when previewing pages, possibly [related to analytics](https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/init.js.erb#L28)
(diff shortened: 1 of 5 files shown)


tijmenb:
> Traditionally smart-answers has always been served on "smartanswers.{environment}.gov.uk.
> That has been preserved here.

What's the reason for keeping this? Might be worth removing the inconsistency now. Otherwise, maybe serve it on both domains?

leenagupte:
> > Traditionally smart-answers has always been served on "smartanswers.{environment}.gov.uk.
> > That has been preserved here.
> 
> What's the reason for keeping this? Might be worth removing the inconsistency now. Otherwise, maybe serve it on both domains?

@tijmenb I kept it because it's what people are used to, but I'm happy to remove it.

leenagupte:
@barrucadu Updated. Can you re-review?

PR #42, opened 2019-07-03
Rewrite govuk-docker CLI in Ruby

thomasleese:
This PR rewrites the existing `govuk-docker` command into Ruby. This has a number of benefits including the ability to write tests, more readable code and more extensible.

There is one breaking change which is that any unrecognised commands are no longer passed through to `docker-compose`. If you want this behaviour you have to use the explicit `govuk-docker compose` command. This decision was made to avoid potential future clashes with commands and because user research said that people prefer longer more explicit commands.

Paired with @bevanloon, @edwardkerry, @erino, @AlanGabbianelli, @andysellick.


lib/commands/run_this.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  attr_reader :stack, :args
+
+  def container_name
+    "#{service}-#{stack}"
+  end
+
+  def extra_args
+    return [] if args.empty?
+    return args if args == ["env"]
```
benthorner: Do we want to check this is the first arg, rather than the only arg?

PR #43, opened 2019-07-03
Make stack a keyword argument

thomasleese:
By default the stack will be 'default' and for other stacks you can pass in the `--stack` argument.

This makes it clearer what the argument means rather than an anonymous positional argument as it is currently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,13 @@ publish a document end-to-end e.g.
 
 ```
 # Run whitehall rake plus any required dependencies (DBs)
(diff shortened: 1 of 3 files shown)


benthorner:
👍. However, I think we need to revise all the Makefiles to accommodate this. If you're doing that, can you try and build in https://github.com/alphagov/govuk-docker/pull/37#pullrequestreview-257013246?

thomasleese:
Yeah, the Makefile uses `compose run` so it isn't affected by this. I've added some tests for this now too.

PR #45, opened 2019-07-03
Add stack for dummy government-frontend

benthorner:
https://trello.com/c/akpi3nt3/2-%F0%9F%92%AA-optimise-day-to-day-development

This adds a 'dummy' stack to the government-frontend service to make it
use external content items - in this case, from the Heroku sample app.

Emulates some of the current behaviour in https://github.com/alphagov/government-frontend/blob/master/startup.sh

### DIFF
--- a/services/government-frontend/docker-compose.yml
+++ b/services/government-frontend/docker-compose.yml
@@ -44,3 +44,17 @@ services:
       VIRTUAL_HOST: draft-government-frontend.dev.gov.uk
       PLEK_HOSTNAME_PREFIX: "draft-"
       HOST: 0.0.0.0
+
+  government-frontend-app-live:
+    <<: *government-frontend
+    depends_on:
+      - nginx-proxy
+    environment:
+      GOVUK_WEBSITE_ROOT: https://www.gov.uk
+      PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
+      PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
+      VIRTUAL_HOST: government-frontend.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    command: bin/rails s -P /tmp/rails.pid


DilwoarH:
I think instead of "dummy", we should use the "live" / "prod" endpoints. When debugging issues on the frontend, we usually want to see how changes would actually affect live content items.

ref: https://github.com/alphagov/government-frontend/blob/master/startup.sh#L15

services/government-frontend/docker-compose.yml
```diff
@@ -44,3 +44,17 @@ services:
...
+
+  government-frontend-dummy:
+    <<: *government-frontend
+    depends_on:
+      - nginx-proxy
+    environment:
+      GOVUK_WEBSITE_ROOT: https://www.gov.uk
+      PLEK_SERVICE_CONTENT_STORE_URI: https://govuk-content-store-examples.herokuapp.com/api
+      PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
+      VIRTUAL_HOST: dummy-government-frontend.dev.gov.uk
```
DilwoarH: Should we keep this consistent with the stack name, ie. `government-frontend-dummy.dev.gov.uk`?

vanitabarrett:
At the moment, our `./startup.sh` script takes a `--live` flag which runs government-frontend pointing at live static, content-store etc. The naming of the stacks may be confusing to someone who is coming from that way of running things, as that isn't what the "live" stack is doing.

DilwoarH:
I've tested the actual code, it works as expected. So once we have agreed everything, it's 👍

services/government-frontend/docker-compose.yml
benthorner: So the naming here is consistent with e.g. 'draft-government-frontend'. We could also just called it 'government-frontend', as long as people aren't wanting to spinup this stack and the live stack at the same time. Would that be better?

benthorner:
So the consensus at the moment is for this to use data from the production environment; we can always add more stacks or potentially go back to a script if we really need the other 3 environments.

Question remains about how to name this stack. In making this change, I've gone with 'real' to distinguish it from the data available on the private Docker network. How about 'public'? Or 'www'? Or 'govuk'? Just a few suggestions - feel free to make your own.

services/government-frontend/docker-compose.yml
```diff
@@ -44,3 +44,17 @@ services:
...
+
+  government-frontend-real:
+    <<: *government-frontend
+    depends_on:
+      - nginx-proxy
+    environment:
+      GOVUK_WEBSITE_ROOT: https://www.gov.uk
+      PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
+      PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
+      VIRTUAL_HOST: real-government-frontend.dev.gov.uk
```
DilwoarH: maybe `www-government-frontend.dev.gov.uk` is better than real.

DilwoarH: Happy for any other suggestions. `prod-government-frontend.dev.gov.uk` or even `live-government-frontend.dev.gov.uk`.

benthorner: We could just go with 'government-frontend', which just leaves the stack name to decide on. Does that help?

tijmenb: 👍 for keeping the domain `government-frontend.dev.gov.uk` for all different stacks, you're not likely to run 2 different versions at the same time.

services/government-frontend/docker-compose.yml
```diff
@@ -44,3 +44,17 @@ services:
       VIRTUAL_HOST: draft-government-frontend.dev.gov.uk
       PLEK_HOSTNAME_PREFIX: "draft-"
       HOST: 0.0.0.0
+
+  government-frontend-real:
```
tijmenb: May I suggest `govuk-docker run --stack production`?

tijmenb: We could also rename all the `live` stacks and use that?

benthorner:
Updated after the big renaming - would someone like to re-review?

PR #46, opened 2019-07-03
Add govuk-docker doctor

bevanloon:
Add a `govuk-docker doctor` command to the CLI.

Add checks for dnsmasq to determine whether or not it is installed and running.

Print some helpful messages to report the state of dnsmasq and some remedial steps if needed.

### DIFF
--- a/lib/govuk_docker_cli.rb
+++ b/lib/govuk_docker_cli.rb
@@ -4,6 +4,7 @@ require_relative "./commands/build_this"
 require_relative "./commands/compose"
 require_relative "./commands/prune"
 require_relative "./commands/run_this"
+require_relative "./doctor/dnsmasq"
 
 class GovukDockerCLI < Thor
   package_name "govuk-docker"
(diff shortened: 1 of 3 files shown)


lib/govuk_docker_cli.rb
```diff
@@ -28,4 +29,10 @@ def prune
   def run_this(*args)
     Commands::RunThis.new(options[:stack], args).call
   end
+
+  desc "doctor", "Various tests to help diagnose issues when running govuk-docker"
+  def doctor
```
thomasleese: Would you be able to move this after `compose` so this stays in alphabetical order?

spec/doctor/dnsmasq_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it "should tell me to install dnsmasq if it is not yet installed" do
+      allow(subject).to receive(:system).with("which dnsmasq 1>/dev/null").and_return(false)
+      subject { described_class.new }
+      expect(subject.call).to include("Dnsmasq not found")
+    end
+  end
+
+  context "when dnsmasq is installed" do
+    it "should report that dnsmasq is installed if it is" do
+      subject { described_class.new }
```
thomasleese: You could move this line to the top of the `describe` since it's the same in each `it` block.

lib/govuk_docker_cli.rb
bevanloon: 👍 yep done

spec/doctor/dnsmasq_spec.rb
bevanloon: good idea - done that too

PR #48, opened 2019-07-03
Add an all-apps target

barrucadu:
This is useful for teams which involve all the apps, like Platform Health.


barrucadu:
Whoops, we even removed the `build` target entirely

PR #50, opened 2019-07-03
use `-` instead of `_` to keep consistent with previous implementation

DilwoarH:
Using _ in URL names is not valid.

tester: https://regexr.com/3au3g

### DIFF
--- a/services/govuk_publishing_components/docker-compose.yml
+++ b/services/govuk_publishing_components/docker-compose.yml
@@ -21,7 +21,7 @@ services:
     depends_on:
       - nginx-proxy
     environment:
-      VIRTUAL_HOST: govuk_publishing_components.dev.gov.uk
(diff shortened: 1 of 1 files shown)


benthorner:
Has this actually caused any problems for people? I understand the purity aspect of it. Do we think people will still be able to guess the domain from the repo, or will they have to go and look it up?

DilwoarH:
I was wondering if after everything has been started, we should print a message saying:

```
App live on http://govuk-publishing-components.dev.gov.uk
```

benthorner:
@thomasleese how about it ^^^? I'm still wary of making this change unless something is actually broken - at the moment we have [something in the README](https://github.com/alphagov/govuk-docker#introduction) that encourages people to think of repo names as URLs. If we want to move away from that, we should have something to support people.

PR #55, opened 2019-07-03
Add a govuk-docker startup command

thomasleese:
This makes it easy to run a service in the `app` stack. It's designed to make it very clear what its doing by the name `startup`. It doesn't allow you to change the command being run, for that you would need to use `govuk-docker run --stack app <command>`.

This command is designed to make it difficult to use this to run a stack which isn't `app` since you should be using the `run` command for that.

Examples:

- `govuk-docker startup live`
- `govuk-docker startup`

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,14 @@ publish a document end-to-end e.g.
 # Run whitehall rake plus any required dependencies (DBs)
 whitehall$ govuk-docker run rake
 
+# Run whitehall rake plus a minimal app stack
+whitehall$ govuk-docker run --stack app rake
+
 # Start content-tagger rails plus a minimal app stack
-content-tagger$ govuk-docker run --stack app
+content-tagger$ govuk-docker startup
 
 # Start content-publisher rails plus an end-to-end stack
-content-publisher$ govuk-docker run --stack app-e2e
+content-publisher$ govuk-docker startup e2e
 ```
 
 In the last two commands, the app will be available in your browser at *app-name.dev.gov.uk*.
--- a/lib/govuk_docker_cli.rb
+++ b/lib/govuk_docker_cli.rb
@@ -53,4 +53,10 @@ class GovukDockerCLI < Thor
   def run(*args)
     Commands::Run.new(options[:stack], args).call
   end
+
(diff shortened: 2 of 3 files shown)


thomasleese:
We're actually not going to standardise on `backend`, it's just that we got this done before renaming all the stacks. Perhaps it would be better to wait on that before merging this PR?

benthorner:
Yeah - I think the fact that we can't really update the README indicates that. So we'll be changing e.g. 'web' to 'backend' in some of the standalone repos?

thomasleese:
@benthorner I've updated it following the decision to change to `app`. Although we haven't renamed the stacks yet, this command itself should be ready to go.

lib/govuk_docker_cli.rb
```diff
@@ -53,4 +53,10 @@ def prune
   def run(*args)
     Commands::Run.new(options[:stack], args).call
   end
+
+  desc "startup [APP_STACK]", "Run the service in the current directory with the `app` stack. Variations can be provided, for example `live` or `draft`."
+  def startup(app_stack = nil)
+    stack = app_stack ? "app-#{app_stack}" : "app"
```
benthorner: Minor, but could we use 'variation' instead of 'app_stack', so we don't have too many stack concepts floating around?

thomasleese: Agreed, I've amended that now.

README.md
```diff
@@ -17,11 +17,14 @@ publish a document end-to-end e.g.
...
-content-tagger$ govuk-docker run --stack backend
+# Run whitehall rake plus a minimal app stack
+whitehall$ govuk-docker run --stack app rake
+
+# Start content-tagger rails plus a minimal app stack
+content-tagger$ govuk-docker startup
 
 # Start content-publisher rails plus an end-to-end stack
-content-publisher$ govuk-docker run --stack e2e
+content-publisher$ govuk-docker startup e2e
```
barrucadu: Will that work?  Is the e2e stack actually called "app-e2e"?

thomasleese: Although it's not currently the case, we will be renaming it to that. See https://github.com/alphagov/govuk-docker/pull/60.

barrucadu: :+1: I misread slack and thought we were going with "lite", "app" + variations, and "e2e"

PR #57, opened 2019-07-03
Rename `build-this` to `build` and make it run the setup tasks

AlanGabbianelli:
See specific commit messages for more context
Paired with @thomasalanlee


lib/commands/build.rb
```diff
@@ -0,0 +1,9 @@
+require_relative './base'
+require_relative './compose'
+
+class Commands::Build < Commands::Base
+  def call
+    check_service_exists
+    system.call("make", "-f", "#{config_directory}/Makefile", service)
```
benthorner: Can we just do `make service` here?

thomasleese: By passing the Makefile directly, you can run this from the app directory itself not just `govuk-docker`.

benthorner: Oh, I thought that's what @barrucadu fixed in https://github.com/alphagov/govuk-docker/pull/54?

benthorner: Ah, sorry - I see what you mean now. (It's not the service-specific one.)

PR #59, opened 2019-07-03
Add linting

tijmenb:
🚔  Woop wop it's the sound of Rubocop 👮 

https://trello.com/c/6QSFbxIv

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,2 @@
+inherit_gem:
+  govuk-lint: "configs/rubocop/all.yml"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,3 +3,6 @@ inherit_gem:
 
 AllCops:
   TargetRubyVersion: 2.6
+
+  Exclude:
+      - 'vendor/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,2 +1,5 @@
 inherit_gem:
   govuk-lint: "configs/rubocop/all.yml"
+
+AllCops:
+  TargetRubyVersion: 2.6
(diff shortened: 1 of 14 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,4 @@
+inherit_gem:
+  govuk-lint: "configs/rubocop/all.yml"
```
erino: Should all of the Rails specific cops be disabled?

tijmenb: I'm not sure if they run / interfere with anything?

erino: They're not failing anything at the moment, if it's a problem it can always be disabled later

.rubocop.yml
```diff
@@ -0,0 +1,5 @@
...
+
+AllCops:
+  TargetRubyVersion: 2.6
```
erino: I don't think we need this [given that Rubocop](https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#setting-the-target-ruby-version) uses [.ruby-version](https://github.com/alphagov/govuk-docker/blob/master/.ruby-version)

tijmenb: https://travis-ci.org/alphagov/govuk-docker/builds/553761983 🤷‍♂

erino: 🤷‍♂

erino: I've found the problem, the gems are vendored and Rubocop is checking all of them

vendoring: https://travis-ci.org/alphagov/govuk-docker/builds/553761983#L565
all: https://travis-ci.org/alphagov/govuk-docker/builds/553761983#L650

.rubocop.yml
```diff
@@ -0,0 +1,5 @@
+inherit_gem:
+  govuk-lint: "configs/rubocop/all.yml"
+
+AllCops:
```
erino: Can we excluded vendored ruby
```yaml
Exclude:
    - 'vendor/**/*'
```

PR #61, opened 2019-07-03
Add doctor docker

bevanloon:
Add a check for docker and docker-compose installation on the system.

Output either a message confirming that it is installed or some basic instructions on where to get docker.

This creates a class that is remarkably similar to https://github.com/alphagov/govuk-docker/blob/adb20b5208b101bb4bc722180a2ef9890852bfa2/lib/doctor/dnsmasq.rb and there's certainly room for some refactoring. We'll get to that in a later commit.

### DIFF
--- a/lib/doctor/docker.rb
+++ b/lib/doctor/docker.rb
@@ -1,0 +1,36 @@
+module Doctor
+  class Docker
+    def initialize
+      @message = []
+    end
+
+    def call
+      install_state?
+
+      message.join
+    end
+
+  private
+
+    attr_reader :message
+
+    DOCKER_INSTALLED = "✅ Docker is installed".freeze
+    INSTALL_DOCKER = <<~HEREDOC.freeze
+      ❌ Docker not found.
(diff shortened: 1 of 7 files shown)


lib/doctor/docker.rb
```diff
@@ -0,0 +1,35 @@
...
+    attr_reader :message
+
+    DOCKER_INSTALLED = "✅ Docker is installed".freeze
+    INSTALL_DOCKER = <<~HEREDOC.freeze
+      ❌ Docker not found.
+      You should install docker by grabbing the latest image from https://docs.docker.com/docker-for-mac/release-notes/.
+      For manual installation, visit https://docs.docker.com/install/.
+    HEREDOC
+
+    def get_install_state
```
erino: Can this method be called `install_state`? or `add_install_state_message`

bevanloon: 👍I've changed it to `install_state?`. Also changed the corresponding names in https://github.com/alphagov/govuk-docker/blob/18420192368c1a1d1bfc71d0e5639efd16c4b06e/lib/doctor/dnsmasq.rb

PR #64, opened 2019-07-03
Fix dependencies in calculators app

thomasleese:
This should fix the error we get currently.

```ERROR: Service 'calculators-app' depends on service 'router-live' which is undefined.```


tijmenb:
Should have been caught by CI!

PR #68, opened 2019-07-03
Install Dnsmasq

thomasleese:
This adds a `govuk-docker install` command which currently only installs and configures Dnsmasq. This is currently limited only to MacOS machines, but it could be extended in the future to support different operating systems.

### DIFF
--- a/lib/install/dnsmasq.rb
+++ b/lib/install/dnsmasq.rb
@@ -1,8 +1,10 @@
+require "thor"
 require_relative "../doctor/dnsmasq"
 
 module Install
   class Dnsmasq
     def call
+      check_macos
       install_dnsmasq
       configure_etc_resolver_devgovuk
       configure_usr_local_etc_dnsmasq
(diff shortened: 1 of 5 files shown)


lib/install/dnsmasq.rb
```diff
@@ -0,0 +1,38 @@
...
+      verify_dns
+      puts "✅ Dnsmasq installation and configuration complete!"
+    end
+
+  private
+
+    def install_dnsmasq
+      return if Doctor::Dnsmasq.new.installed?
+      puts "⏳ Installing dnsmasq"
+      system("brew install dnsmasq")
```
bevanloon: I wonder if we should have something like `system("brew install dnsmasq") if %x[uname -a].include?("Darwin")` just in case someone's not on a Mac?

thomasleese: Yeah, I'm thinking actually the location of the config files only apply when installed by brew so I think I'll put a check right at the top which asks the user to confirm if they really want to do this if they're not on a Mac.

lib/install/dnsmasq.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    def install_dnsmasq
+      return if Doctor::Dnsmasq.new.installed?
+
+      puts "⏳ Installing dnsmasq"
+      system("brew install dnsmasq")
+    end
+
+    def configure_etc_resolver_devgovuk
+      write_file(
```
bevanloon: Is this going to destroy the current contents of the file if they exist?
Do we want to warn people of that ?

thomasleese: That's a good idea, I've changed the whole script to require confirmation before continuing and it outputs to the user what it's going to do.

PR #74, opened 2019-07-03
Add verbose option to `run` and `compose` commands

edwardkerry:
Unless the user explicitly passes a `--verbose` flag the potentially very long list of service's docker compose files will be replaced in the terminal output by `-f [...]`

`whitehall $govuk-docker run --verbose`

```
docker-compose -f /Users/example_user/govuk/govuk-docker/lib/commands/../../docker-compose.yml -f /Users/example_user/govuk/govuk-docker/lib/commands/../../services/content-store/docker-compose.yml -f /Users/example_user/govuk/govuk-docker/lib/commands/../../services/signon/docker-compose.yml  run --rm --service-ports whitehall-default
```

`whitehall $govuk-docker run`
```    
docker-compose -f [...] run --rm --service-ports whitehall-default
```

### DIFF
--- a/lib/commands/run.rb
+++ b/lib/commands/run.rb
@@ -2,8 +2,8 @@ require_relative './base'
 require_relative './compose'
 
 class Commands::Run < Commands::Base
-  def initialize(stack, args, service = nil, config_directory = nil)
-    super(service, config_directory, nil, stack)
+  def initialize(stack, verbose, args, service = nil, config_directory = nil)
+    super(service, config_directory, nil, stack, verbose)
     @args = args
   end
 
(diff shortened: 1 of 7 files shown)


lib/commands/run.rb
```diff
@@ -2,16 +2,16 @@
 require_relative './compose'
 
 class Commands::Run < Commands::Base
-  def initialize(stack, args, service = nil, config_directory = nil)
-    super(service, config_directory, nil, stack)
+  def initialize(stack, verbose, args, service = nil, config_directory = nil)
```
benthorner: Minor, but this is probably a good time to start using hash args :-).

thomasleese: I agree. I think the whole base command initialiser could do with a refactor now anyway which I'm planning to do in a later PR, so we can merge this one in for now I think.

PR #75, opened 2019-07-03
Standard cleanup

benthorner:
This PR looks at standardising our approach to cleanup, so that we always do it after setting-up something. After writing this, I'm wondering if we should even have automated cleanup at all? It doesn't seem write to forcibly stop all services after making one of them, and the global case is so rarely used that someone could reasonably be expected to do the two individual commands themselves.


benthorner:
Superseded by https://github.com/alphagov/govuk-docker/pull/119.

PR #86, opened 2019-07-04
Add service-manual-frontend

tijmenb:
Tests work and the app can be run against production (`govuk-docker startup live`). Most of the Dockerfile is similar to government-frontend.

https://trello.com/c/DMaWjgs8

### DIFF
--- a/services/nginx-proxy/docker-compose.yml
+++ b/services/nginx-proxy/docker-compose.yml
@@ -22,16 +22,17 @@ services:
           - draft-router.dev.gov.uk
           - draft-static.dev.gov.uk
           - government-frontend.dev.gov.uk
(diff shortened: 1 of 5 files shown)


services/nginx-proxy/docker-compose.yml
```diff
@@ -22,16 +22,17 @@ services:
           - draft-router.dev.gov.uk
           - draft-static.dev.gov.uk
           - government-frontend.dev.gov.uk
-          - govuk-developer-docs.dev.gov.uk
           - govuk_publishing_components.dev.gov.uk
```
benthorner: Not exactly in this PR, but I believe this needs updating to use '-' instead of '_'.

PR #90, opened 2019-07-04
Add collections-publisher directory

hannako:
Pair effort with @sihugh 

Adds files to run collections publisher in a container. 

🎾 Tests passing 
🎾 Works in browser

Screen Shot 2019-07-04 at 13 00 07

https://trello.com/c/cEZDDhQK/25-set-up-collections-publisher

### DIFF
--- a/services/collections-publisher/Dockerfile
+++ b/services/collections-publisher/Dockerfile
@@ -5,6 +5,7 @@ RUN apt-get install -y libxss1 libappindicator1 libindicator7
 RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && apt install -y ./google-chrome*.deb
 
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt-get update -qq && apt-get install -y nodejs
 
 RUN useradd -m build
 USER build
--- a/services/collections-publisher/Dockerfile
+++ b/services/collections-publisher/Dockerfile
@@ -1,0 +1,12 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim
+RUN apt-get install -y libxss1 libappindicator1 libindicator7
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && apt install -y ./google-chrome*.deb
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+
+RUN useradd -m build
+USER build
+
+ENV EDITOR=vim
--- a/services/collections-publisher/docker-compose.yml
+++ b/services/collections-publisher/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - redis
       - content-store-app
       - link-checker-api-app
-      - nginx-proxy
+      - nginx-proxy-app
       - publishing-api-app
       - collections-publisher-worker
     environment:
--- a/services/collections-publisher/docker-compose.yml
+++ b/services/collections-publisher/docker-compose.yml
@@ -1,0 +1,55 @@
+version: '3.7'
+
+x-collections-publisher: &collections-publisher
+  build:
+    context: .
+    dockerfile: services/collections-publisher/Dockerfile
+  image: collections-publisher
+  volumes:
+    - ..:/govuk:delegated
+    - bundle:/usr/local/bundle
+    - home:/home/build
+  working_dir: /govuk/collections-publisher
+
+services:
+  collections-publisher-lite:
+    <<: *collections-publisher
+    privileged: true
+    depends_on:
+      - mysql
+      - redis
+    environment:
+      # the app uses this URL to generate a test database URL
+      DATABASE_URL: "mysql2://root:root@mysql/collections_publisher_development"
+      REDIS_URL: redis://redis
+      DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: "true"
+
+  collections-publisher-app: &collections-publisher-app
+    <<: *collections-publisher
+    depends_on:
+      - mysql
+      - redis
+      - content-store-app
+      - link-checker-api-app
+      - nginx-proxy
+      - publishing-api-app
+      - collections-publisher-worker
+    environment:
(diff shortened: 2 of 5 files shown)


services/collections-publisher/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim
+RUN apt-get install -y libxss1 libappindicator1 libindicator7
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && apt install -y ./google-chrome*.deb
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update -qq && apt-get install -y yarn nodejs
+
+RUN apt-get clean
```
sihugh: Apparently we don't need this because official docker images like `ruby` [automatically run it (at the bottom of the section in a call out)](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#apt-get)

services/collections-publisher/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim
+RUN apt-get install -y libxss1 libappindicator1 libindicator7
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && apt install -y ./google-chrome*.deb
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update -qq && apt-get install -y yarn nodejs
```
benthorner: It looks like we don't need yarn for this service - we don't use the command in the Makefile :-).

services/collections-publisher/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    depends_on:
+      - redis
+      - mysql
+      - publishing-api-app
+    environment:
+      DATABASE_URL: "mysql2://root:root@mysql/collections_publisher_development"
+      REDIS_URL: redis://redis
+    command: bundle exec sidekiq -C ./config/sidekiq.yml
+
+  collections-publisher-app-e2e:
```
benthorner: We don't have to have an E2E stack :-). If we do have one, it will usually have heavier dependencies and the testing would involve seeing we can publish a document and see it on the frontend.

services/collections-publisher/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+
+  collections-publisher-app: &collections-publisher-app
+    <<: *collections-publisher
+    depends_on:
+      - mysql
+      - redis
+      - content-store-app
+      - link-checker-api-app
+      - nginx-proxy
+      - publishing-api-app
```
sihugh: did we want the worker to appear here too (so adding `- collections-publisher-worker`)?

services/collections-publisher/Dockerfile
hannako: 👍

services/collections-publisher/docker-compose.yml
hannako: yes, thanks - it was there before... I think I must have done a wonky rebase and it got knocked off!

services/collections-publisher/docker-compose.yml
hannako: 👍

services/collections-publisher/Dockerfile
hannako: Thanks

hannako:
@benthorner I've updated, please could you re-review?

services/collections-publisher/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim
+RUN apt-get install -y libxss1 libappindicator1 libindicator7
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb && apt install -y ./google-chrome*.deb
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
```
benthorner: I think we’re missing the actual install for ‘nodejs’ now. You might want to try ‘govuk-docker build collections-publisher-lite’ to ensure you’re working with the image represented by this file, as I suspect the tests will now fail if nodejs is required.

hannako: Running ‘govuk-docker build collections-publisher-lite’ gave me an argument error. 

But I ran `govuk-docker build` which did
```
/Users/jessicajones/govuk/govuk-docker/bin/govuk-docker compose run collections-publisher-lite sh -c 'rake db:migrate 2>/dev/null || rake db:setup'
docker-compose -f [...] run collections-publisher-lite sh -c rake db:migrate 2>/dev/null || rake db:setup
```
...and the tests are still green and the app seems to be running fine in the browser. Guess nodejs isn't required?

benthorner: I’m afraid not. The image still needs to be rebuilt. The build command should probably be changed to actually try and build the image - it wasn’t necessary in the beginning because it gets built automatically the first time.

hannako: `govuk-docker compose down` and then `govuk-docker build`? Or do I need to run Leena's trifecta of destruction explode all the things commands?

benthorner: Oops - I gave you the wrong command earlier. I meant: `govuk-docker compose build collections-publisher-lite`.

This command should really be in all Makefiles to ensure that the image for each service is in-sync with its Dockerfile - would you be willing to do a PR on that?

hannako: ah ok that works, and now I'm getting the node errors. Will add it back 👍

PR #93, opened 2019-07-04
Switch to automatic ruby versioning

benthorner:
https://trello.com/c/ADvk7Bbx/9-%F0%9F%94%AE-make-it-stick

Previously we used hard-coded ruby base images, which makes it harder to
upgrade ruby versions in stages across the platform; some services were
also using a different ruby version to the one specified in their repo.
Using service-specific base images also makes it harder to consolidate
their Dockerfiles, since different apps need different versions of Ruby.

This replaces the ruby base image with 'buildpack-deps' (the one 'ruby'
is based on) and installs rbenv to automatically manage ruby versions
for each app as part of the make/setup process. Doing this comes at the
cost of an increased and potentially more volatile local setup time, but
also a more familiar 'rbenv install' experience.


Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Standard Dockerfile for GOV.UK services. Random service-specific stuff should *not*
+# go in this file. Instead, you should copy this file into the directory for the
```
tijmenb: Unsure about this advice. Can't we tell people to import the image?

Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Standard Dockerfile for GOV.UK services. Random service-specific stuff should *not*
```
tijmenb: Perhaps we can call this file something like `govuk-base.dockerfile` to differentiate it from the other `Dockerfiles`?

tijmenb:
This is a good solution to https://trello.com/c/1PBngwbn I guess! Any idea how much slower setup will be with this?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update -qq && apt-get install -y yarn nodejs
+
+RUN git clone https://github.com/sstephenson/rbenv.git /rbenv
+RUN git clone https://github.com/sstephenson/ruby-build.git /rbenv/plugins/ruby-build
+RUN /rbenv/plugins/ruby-build/install.sh
+ENV PATH /rbenv/bin:$PATH
+ENV PATH /rbenv/shims:${PATH}
+RUN mkdir /rbenv/versions /rbenv/shims
+
```
tijmenb: Could we pre-install some of the most used Rubies here, so most of the apps don't have to install the same thing?

benthorner:
@thomasleese indeed. I've actually made one of the changes - we now have 'Dockerfile.govuk-base' - in order to avoid CI trying to treat the repo as a Docker thing we need to deploy.

The card could be a good place to iterate from. https://trello.com/c/1PBngwbn/53-allow-us-to-easily-upgrade-rubies

Dockerfile
benthorner: I've gone with 'Dockerfile.govuk-base', which seems to be the convention for variants of Dockerfiles.

Dockerfile
benthorner: Agreed, this isn't ideal. I've documented the 'single bloated image/Dockerfile' on the card as something to think about more.

PR #94, opened 2019-07-04
Expose mongodb ports

peteglondon:
Exposing mongodb (27017,27018) ports so we can connect to it from
the host machine.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,9 @@ services:
     image: mongo:2.4
     volumes:
       - mongo:/data/db
+    ports:
+      - "27017:27017"
+      - "27018:27018"
 
   mysql:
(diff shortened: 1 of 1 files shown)


benthorner:
How about this instead...

```
govuk-docker compose start mongo
govuk-docker compose exec mongo mongo
```

peteglondon:
I want to be able to run tests on the host machine, against mongodb in docker.

benthorner:
One thing to bear in mind here is that this may start to cause problems with local setups, if we're using the normal ports, since these will conflict with anything running on the local machine e.g. if you have a local mongo, or a local postgres running.

tijmenb:
Let's take the risk of the local mongo conflict, as this might help people develop faster.

PR #95, opened 2019-07-04
Add publisher

peteglondon:
Tests are passing.

App is running and you can create a thing.

Screenshot 2019-07-04 at 14 06 50

Haven't added e2e stack as I don't know what to put in it.

https://trello.com/c/gn3pWjbk/28-set-up-publisher

### DIFF
--- a/services/calendars/Makefile
+++ b/services/calendars/Makefile
@@ -1,3 +1,4 @@
 calendars: $(GOVUK_ROOT_DIR)/$@
 	$(GOVUK_DOCKER) compose build $@-lite
 	$(GOVUK_DOCKER) compose run $@-lite bundle
+	$(GOVUK_DOCKER) compose run $@-setup bin/rake publishing_api:publish
--- a/services/publisher/Dockerfile
+++ b/services/publisher/Dockerfile
@@ -1,0 +1,19 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim
+  RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb 2>&1 && \
+  apt install -y ./google-chrome*.deb && \
+  rm ./google-chrome*.deb
+
+RUN wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2
+RUN tar -xf phantomjs-2.1.1-linux-x86_64.tar.bz2
+RUN cp phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/local/bin/phantomjs
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt-get update -qq && apt-get install -y nodejs
+
+RUN useradd -m build
+USER build
+
+ENV EDITOR=vim
+
--- a/services/publisher/docker-compose.yml
+++ b/services/publisher/docker-compose.yml
@@ -1,0 +1,24 @@
+version: '3.7'
+
+x-publisher: &publisher
+  build:
+    context: .
+    dockerfile: services/publisher/Dockerfile
+  image: publisher
+  volumes:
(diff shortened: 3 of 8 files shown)


services/publisher/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+      MONGODB_URI: "mongodb://mongo/publisher"
+      TEST_MONGODB_URI: "mongodb://mongo/publisher-test"
+      REDIS_URL: redis://redis
+      VIRTUAL_HOST: publisher.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    command: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT:-3000}
+
+  publisher-worker:
```
benthorner: Looks like nothing is using this - do we need it?

services/publisher/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+      - router-app
+    environment:
+      MONGODB_URI: "mongodb://mongo/publisher"
+      TEST_MONGODB_URI: "mongodb://mongo/publisher-test"
+      REDIS_URL: redis://redis
+      VIRTUAL_HOST: publisher.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    command: bundle exec unicorn -c ./config/unicorn.rb -p ${PORT:-3000}
```
benthorner: Would it be possible to use `rails s` here, like we do everywhere else?

services/publisher/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+      - redis
+      - mongo
+      - calendars-app
+      - publishing-api-app
+      - asset-manager-app
+      - content-store-app
+      - router-app
+    environment:
+      MONGODB_URI: "mongodb://mongo/publisher"
+      TEST_MONGODB_URI: "mongodb://mongo/publisher-test"
```
benthorner: We don't need test vars for app stacks.

services/publisher/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+  rm ./google-chrome*.deb
+
+RUN wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2
+RUN tar -xf phantomjs-2.1.1-linux-x86_64.tar.bz2
+RUN cp phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/local/bin/phantomjs
+
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt-get update -qq && apt-get install -y nodejs
+
+RUN apt-get clean
```
benthorner: We figured out this happens automatically, so we shouldn't need to do this :-).

kevindew:
@benthorner I've picked this up and made quite a lot of changes, all your comments should now be covered.

Example of a published page: 
Screen Shot 2019-07-05 at 12 08 50

A few things:

- I noticed static was erroring due to lack of redis so I've resolved that.
- Calendar needs to publish routes before being usable so I've added that to setup.
- It didn't seem like link-checker-api had been working properly given it's set-up. I've been able to successfully check links.
- The publisher repo seems to have a lot of excess bits that don't seem to be used any more. As far as I can tell there is actually no need for Asset Manager and Imminence dependencies as they use models that are no longer used, though this is rather hard to confirm.

services/calendars/Makefile
```diff
@@ -1,2 +1,3 @@
 calendars: $(GOVUK_ROOT_DIR)/calendars
 	$(COMPOSE_RUN) calendars-lite bundle
```
benthorner: I'm doing a PR to add an explicit 'build' step for the 'lite' service. Can you add one in here?

benthorner: https://github.com/alphagov/govuk-docker/pull/110

kevindew: Done

PR #100, opened 2019-07-04
[do-not-merge] Adds search-api configuration

oscarwyatt:
Adds configuration for search-api.

The docker side of things is relatively simple (ish) but the problem here is that this breaks (or reveals broken) tests in search-api. @koetsier and I have fixed these in a branch of search-api called (temporarily, and not by me) `ozkar_spike`. Memoization was causing a lot of these problems and Jos is looking into ways to remove this memoization without causing performance problems (as some of the stuff that is memoized is quite expensive)

One thing to note is the environment variable ` ES_JAVA_OPTS=-Xms512m -Xmx512m` - Docker has a default memory size of 2gb and elasticsearch also has a default memory size of 2gb. When attempting to bring up both, the container runs out of memory and one gets killed so this prevents that

### DIFF
--- a/services/search-api/Dockerfile
+++ b/services/search-api/Dockerfile
@@ -1,0 +1,8 @@
+FROM ruby:2.6.3
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y vim && apt-get clean
+
+RUN useradd -m build
+USER build
+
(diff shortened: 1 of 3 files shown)


kevindew:
Closing in favour of: https://github.com/alphagov/govuk-docker/pull/101

PR #102, opened 2019-07-05
 Print when the application becomes live

thomasleese:
This means you see a message like

```
Application is available at: http://government-frontend.dev.gov.uk
```

when the app starts if you use `govuk-docker startup`.

### DIFF
--- a/lib/commands/startup.rb
+++ b/lib/commands/startup.rb
@@ -1,3 +1,7 @@
+require 'colorize'
+require 'net/http'
+require 'timeout'
+
 require_relative './base'
 require_relative './run'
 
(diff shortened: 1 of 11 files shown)


spec/govuk_docker_cli_spec.rb
```diff
@@ -97,10 +97,8 @@
 
     context "without a variation argument" do
       it "runs in the app stack" do
-        expect(Commands::Run)
-          .to receive(:new).with(stack: "app", verbose: false)
-          .and_return(command_double)
-        expect(command_double).to receive(:call).with(no_args)
+        expect(Commands::Startup).to receive(:new).and_return(command_double)
```
erino: Where these changes meant to be in commit 878283ed53bf5f52b552e0500d3e94aefe932c97?

spec/commands/startup_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "spec_helper"
+require_relative "../../lib/commands/startup"
+
+describe Commands::Startup do
+  let(:config_directory) { "spec/fixtures" }
+
+  subject { described_class.new(config_directory: config_directory, service: "example-service") }
+
+  let(:run_double) { instance_double(Commands::Run) }
+  before { allow(run_double).to receive(:call) }
```
erino: What about adding `allow(subject).to receive(:puts)` here to keep the spec runner output clean?

spec/govuk_docker_cli_spec.rb
thomasleese: Yep... I've fixed that now.

spec/commands/startup_spec.rb
thomasleese: Good idea, I didn't even notice the printing! I've added this now.

lib/commands/startup.rb
```diff
@@ -0,0 +1,72 @@
...
+  def wait_until_can_visit?(url)
+    loop do
+      sleep(1)
+      break if can_visit?(url)
+    end
+  end
+
+  def can_visit?(url)
+    case Net::HTTP.get_response(URI(url))
+    when Net::HTTPSuccess then
```
erino: This should also include 3xx responses as Content Publisher redirects / to /documents

lib/commands/startup.rb
```diff
@@ -0,0 +1,73 @@
...
+  def possible_hostnames
+    path = File.join(config_directory, "services", "nginx-proxy", "docker-compose.yml")
+    docker_compose = YAML.load_file(path)
+    docker_compose["services"]["nginx-proxy-app"]["networks"]["default"]["aliases"]
+  end
+
+  def find_hostname
+    search_hostname = "#{service.tr('_-', '')}.dev.gov.uk"
+    possible_hostnames.each do |hostname|
+      return hostname if hostname.tr('_-', '') == search_hostname
```
bevanloon: What are we guarding against with the `.tr` here and on line 45? Is it just because we inevitable forget which services have underscores and which have dashes when we type?

thomasleese: At the moment it's only `govuk_publishing_components` that needs this, https://github.com/alphagov/govuk-docker/pull/50.

The service is `govuk_publishing_components`, the address is `govuk-publishing-components.dev.gov.uk`.

PR #104, opened 2019-07-05
 Add a setup step to clone/pull the repo

thomasleese:
This takes us towards the idea of being able to do `gem install govuk-docker` because the next step can be to do `govuk-docker install` and the repo will be cloned automatically.

### DIFF
--- a/lib/setup/repo.rb
+++ b/lib/setup/repo.rb
@@ -1,0 +1,58 @@
+require_relative "./base"
+require_relative "../paths"
+
+class Setup::Repo < Setup::Base
+  def call
+    unless should_clone_or_pull?
+      puts "Ignoring git repo because of local changes."
+      return
+    end
+
+    return unless check_continue
+
+    clone_or_pull
+
+    puts "✅ govuk-docker is up to date!"
+  end
+
+private
+
+  def should_clone_or_pull?
+    return true unless File.directory?(path)
+    return false if repo_has_local_changes?
+
+    true
+  end
+
+  def repo_has_local_changes?
+    !system("git -C #{path} diff-index --quiet HEAD --")
+  end
+
+  def check_continue
+    puts "This will clone/pull `#{path}`."
+    puts "Any local changes may be overwriten."
+    puts
+
+    shell.yes?("Are you sure you want to continue?")
+  end
+
+  def clone_or_pull
+    if File.directory?(path)
+      pull
+    else
+      clone
+    end
+  end
+
+  def pull
+    system("git -C #{path} pull")
+  end
+
+  def clone
+    system("git clone https://github.com/alphagov/govuk-docker.git #{path}")
+  end
+
+  def path
+    GovukDocker::Paths.govuk_docker_dir
+  end
+end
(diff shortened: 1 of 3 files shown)


benthorner:
I wonder if 'setup' might work better here? I'm struggling with seeing the two 'install' commands together and have to stop and think about what the difference is. Might help to clarify here what we expect the gem to do vs. the cloned repo.

thomasleese:
@benthorner Are you thinking we have two separate commands, `setup` and `install`? Or that we change the name of the command to `setup`?

The idea behind `install` is that you can run it at any point and it will install anything missing to successfully run govuk-docker. So in most cases, it won't reclone the repo since it'll already exist. We're planning to add another `install` step shortly which will install Docker itself for you.

benthorner:
@thomasleese yes to the first question. I think it depends on the big picture. Do we have a complete list of commands we're aiming for? It's actually the 'gem install' that confuses me, as I'm struggling to see the benefit of doing that over 'git clone ...'. Playing a devil's advocate somewhat: "why do I need to install it after I've just installed it" is the question that comes to mind ;-).

thomasleese:
I guess one of the benefits is that you don't need to change your `PATH` to get it working, and it'll work under many different shells (right now the instructions only list how to do it for bash). Most respondents to the survey requested this method of installation so perhaps it's worth asking why.

I don't like the idea of two different installation-like commands because I think it'll get confusing, you'll have to ask yourself "which install command do I run to get the latest Dockerfiles? which one do I run if dnsmasq isn't working?".

At the moment the complete list is: Dnsmasq, Docker and the repo. Enough for the entire installation to be `gem install govuk-docker && govuk-docker install`.

I'm not opposed to renaming it to `govuk-docker setup` though, and I think that does make more sense if you look at it together with `gem install govuk-docker`.

benthorner:
Ah, the use of PATH would indeed be an advantage. It looks like we're asking similar questions - perhaps it's worth listing all the command we have / aim to have, so we can understand what they look like together and where potential confusion might occur, just like the scenario you mentioned.

For example, I think the next command after 'install'/'setup' would be something like 'build', possibly in a repo or with an arg/flag to specify the service; but would this too look confusing after 'install'/'setup'?

lib/setup/repo.rb
```diff
@@ -0,0 +1,56 @@
+require_relative "./base"
+require_relative "../doctor/dnsmasq"
+
+class Setup::Repo < Setup::Base
+  def initialize(*args)
+    super(*args)
+    @path = File.join(ENV.fetch("HOME"), "govuk", "govuk-docker")
```
barrucadu: We've got environment variable inconsistency between `govuk-docker` and the Makefile: https://github.com/alphagov/govuk-docker/blob/f56b7cc51f0ee9223b8ae2d6f158ece2774005de/Makefile#L1-L3

Can we have the same environment variables, with the same defaults?  The advantage of not hard-coding `File.join(ENV.fetch("HOME"), "govuk", "govuk-docker")` is that you can do things in a different directory.

thomasleese: Yep, good call... I've added that in now.

PR #105, opened 2019-07-05
Add `govuk-docker be` alias

AlanGabbianelli:
This is a shorthand version of `govuk-docker run bundle exec`.


lib/govuk_docker_cli.rb
```diff
@@ -89,6 +89,11 @@ def run(*args)
     Commands::Run.new(options).call(args)
   end
 
+  desc "run bundle exec [ARGS]", "Alias for `run bundle exec`"
```
thomasleese:
```suggestion
  desc "be [ARGS]", "Alias for `run bundle exec`"
```

PR #106, opened 2019-07-05
Respect exit codes

benthorner:

This fixes an issue we've been having where the CLI was always exiting with '0', even when the invoked command was malformed or had failed.

### DIFF
--- a/lib/commands/base.rb
+++ b/lib/commands/base.rb
@@ -11,6 +11,27 @@ module Commands
       @verbose = options[:verbose] || default_verbose
     end
 
+    def system_command(*args)
+      system(*args) || raise("Non-zero exit code")
+    end
+
+    def service_exists?
+      search_string = "services/#{service}/docker-compose.yml"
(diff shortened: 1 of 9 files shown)


lib/commands/base.rb
```diff
@@ -4,23 +4,26 @@
 
 module Commands
   class Base
+    attr_reader :config_directory, :service, :stack, :verbose
```
thomasleese: Do these need to be public?

benthorner: Ah, no - it got a bit mixed up in the rebase. Private again.

benthorner:
@bevanloon do you think we can incorporate this technique into the doctors?

bevanloon:
@benthorner I think at the moment, it would fail the individual check if the system calls returned an error anyway, so people will still realise something is broken (if they read the output). But yeah - we should probably add this in as an enhancement because more info is better when debugging.

PR #108, opened 2019-07-05
Add docker and docker-compose to Setup script

edwardkerry:
This will automatically attempt to install docker and docker-compose via brew when running `govuk-docker setup` unless they are already available.

### DIFF
--- a/lib/doctor/doctor.rb
+++ b/lib/doctor/doctor.rb
@@ -18,7 +18,7 @@ module Doctor
       running: "✅ Docker is running",
       not_running: <<~HEREDOC
         ❌ Docker is not running.
-        You should start it with `sudo brew services start dnsmasq`.
+        You should start it with `sudo brew services start docker`.
       HEREDOC
     }
   end
--- a/lib/doctor/doctor.rb
+++ b/lib/doctor/doctor.rb
@@ -18,7 +18,7 @@ module Doctor
       running: "✅ Docker is running",
       not_running: <<~HEREDOC
         ❌ Docker is not running.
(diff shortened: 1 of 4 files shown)


lib/doctor/docker.rb
```diff
@@ -0,0 +1,50 @@
+module Doctor
```
thomasleese: I think you could use https://github.com/alphagov/govuk-docker/blob/master/lib/doctor/checkup.rb instead of needing this class. Something like `Doctor::Checkup.new(service_name: 'docker', checkups: [], messages: []).installed?` (perhaps worth making `checkups` and `messages` default to empty lists in the constructor.

lib/govuk_docker_cli.rb
```diff
@@ -69,7 +70,13 @@ def doctor
   end
 
   desc "install", "Configures and installs the various dependencies necessary to run govuk-docker successfully"
+  long_desc <<~LONGDESC
```
thomasleese: I like this!

lib/doctor/doctor.rb
```diff
@@ -18,7 +18,7 @@ def self.docker_messages
       running: "✅ Docker is running",
       not_running: <<~HEREDOC
         ❌ Docker is not running.
-        You should start it with `sudo brew services start dnsmasq`.
+        You should start it with `sudo brew services start docker`.
```
thomasleese: I don't think it's possible to use `brew services` here, I think we'll just have to tell people to make sure the app is open.

edwardkerry: 👍

PR #111, opened 2019-07-05
Document environment variables

barrucadu:
Hopefully by documenting these here we can avoid behavioural drift
between govuk-docker and the Makefile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,14 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
 app.dev.gov.uk.		0	IN	A	127.0.0.1
 ```
 
+## Environment variables
+
+Both govuk-docker and the Makefile respect the following environment variables:
+
+- `$GOVUK_ROOT_DIR` - directory where app repositories live, defaults to `$HOME/govuk`
+- `$GOVUK_DOCKER_DIR` - directory where the govuk-docker repository lives, defaults to `$GOVUK_ROOT_DIR/govuk-docker`
+- `$GOVUK_DOCKER` - path of the govuk-docker script, defaults to `$GOVUK_DOCKER_DIR/bin/govuk-docker`
+
 ## Compatibility
 
 The following apps are supported by govuk-docker to some extent.


README.md
```diff
@@ -116,6 +116,14 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
 app.dev.gov.uk.		0	IN	A	127.0.0.1
 ```
 
+## Environment variables
+
+Both govuk-docker and the Makefile respect the following environment variables:
+
+- `$GOVUK_ROOT_DIR` - directory where app repositories live, defaults to `$HOME/govuk`
+- `$GOVUK_DOCKER_DIR` - directory where the govuk-docker repository lives, defaults to `$GOVUK_ROOT_DIR/govuk-docker`
```
benthorner: I'm not sure this one is actually respected, since we do a '..' in all the compose files to mount the govuk directory. Perhaps we should change that?

README.md
```diff
@@ -116,6 +116,14 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
+- `$GOVUK_DOCKER` - path of the govuk-docker script, defaults to `$GOVUK_DOCKER_DIR/bin/govuk-docker`
```
benthorner: Just wondering if this is ever changed, or mainly a utility to avoid typing?

barrucadu: It's a utility to avoid typing out the definition everywhere

README.md
barrucadu: That's a good point.  I'd be in favour of changing that - can you lookup environment variables with a default value in compose files?

barrucadu: https://docs.docker.com/compose/environment-variables/ doesn't have anything on it.  So I think if we want to use `$GOVUK_ROOT_DIR` in the compose files we'd have to have `govuk-docker` set it before calling `docker-compose`.  Does that sound reasonable, or is it too magic?

benthorner: That's annoying. Ideally a user should be able to copy the verbose command and achieve the same result, so we maintain some transparency - it's been useful when I've debugged stuff. Could it a prefix to `docker-compose` i.e. `GOVUK_ROOT_DIR=... docker-compose`?

barrucadu: I'll try things out in another branch

barrucadu: Ah, I was looking at the wrong bit of the manual.  This is the relevant bit: https://docs.docker.com/compose/compose-file/#variable-substitution

> `${VARIABLE:-default}` evaluates to `default` if `VARIABLE` is unset or empty in the environment.

barrucadu: docker-compose doesn't seem to like when the default value contains an environment variable.  These don't work:

```
${GOVUK_ROOT_DIR-$HOME/govuk}:/govuk:delegated
${GOVUK_ROOT_DIR-"$HOME/govuk"}:/govuk:delegated
${GOVUK_ROOT_DIR-"$HOME"/govuk}:/govuk:delegated
${GOVUK_ROOT_DIR-${HOME}/govuk}:/govuk:delegated
${GOVUK_ROOT_DIR-"${HOME}/govuk"}:/govuk:delegated
${GOVUK_ROOT_DIR-"${HOME}"/govuk}:/govuk:delegated
```

Just to confirm that the default syntax works, this *does*:

```
${GOVUK_ROOT_DIR-HOME/govuk}:/govuk:delegated
```

Argh

barrucadu:
@benthorner Happy for this to be merged now that the docker-compose.yml issue has been fixed?

PR #113, opened 2019-07-05
Add a paths helper class

thomasleese:
More and more commands are starting to need to know where govuk-docker is so I thought it would be useful to pull that out into a class. By doing this we can be consistent with the use of environment variables and make sure they match the ones used by the Makefile.

See #111 for more details on the environment variables.

### DIFF
--- a/spec/paths_spec.rb
+++ b/spec/paths_spec.rb
@@ -1,0 +1,45 @@
+require "spec_helper"
+require_relative "../lib/paths"
+
+describe GovukDocker::Paths do
+  describe "govuk_root_dir" do
+    subject { described_class.govuk_root_dir }
+
+    context "when GOVUK_ROOT_DIR is not set" do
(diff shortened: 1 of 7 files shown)


spec/paths_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "spec_helper"
+require_relative "../lib/paths"
+
+describe GovukDocker::Paths do
+  describe "govuk_root_dir" do
+    subject { described_class.govuk_root_dir }
+
+    it "uses HOME environment variable" do
```
AlanGabbianelli: Should we put this in `context "when GOVUK_ROOT_DIR is not set" do`?

thomasleese: Yes! I was actually just writing that as you added the comment :smile: I've pushed up an amend with that now.

PR #114, opened 2019-07-05
Add manuals-frontend

sihugh:
Everything works, however we need content in the content store for the app stack to work nicely.  Because of this, I've included a live-app stack.

Screenshot 2019-07-05 13 52 55

### DIFF
--- a/services/manuals-frontend/docker-compose.yml
+++ b/services/manuals-frontend/docker-compose.yml
@@ -1,0 +1,59 @@
+version: '3.7'
+
+x-manuals-frontend: &manuals-frontend
+  build:
+    context: .
+    dockerfile: services/manuals-frontend/Dockerfile
+  image: manuals-frontend
+  volumes:
+    - $GOVUK_ROOT_DIR:/govuk:delegated
+    - bundle:/usr/local/bundle
+  working_dir: /govuk/manuals-frontend
+
+services:
+  manuals-frontend-lite:
+    <<: *manuals-frontend
+    privileged: true
+
+  manuals-frontend-app: &manuals-frontend-app
+    <<: *manuals-frontend
+    depends_on:
+      - router-app
+      - content-store-app
+      - static-app
+    environment:
+      GOVUK_ASSET_ROOT: manuals-frontend.dev.gov.uk
+      VIRTUAL_HOST: manuals-frontend.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    command: bin/rails s -P /tmp/rails.pid
+
+  manuals-frontend-app-draft:
+    <<: *manuals-frontend
+    depends_on:
+      - router-app-draft
+      - content-store-app-draft
+      - static-app-draft
+    environment:
+      GOVUK_ASSET_ROOT: draft-manuals-frontend.dev.gov.uk
+      VIRTUAL_HOST: draft-manuals-frontend.dev.gov.uk
+      PLEK_HOSTNAME_PREFIX: "draft-"
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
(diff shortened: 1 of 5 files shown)


services/manuals-frontend/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+  volumes:
+    - ..:/govuk:delegated
+    - bundle:/usr/local/bundle
+  working_dir: /govuk/manuals-frontend
+
+services:
+  manuals-frontend-lite:
+    <<: *manuals-frontend
+    privileged: true
+    volumes:
```
leenagupte: I don't think you need this, because the volumes are already included in `manuals-frontend` which the `lite` image inherits from.

services/manuals-frontend/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    environment:
+      GOVUK_ASSET_ROOT: manuals-frontend.dev.gov.uk
+      VIRTUAL_HOST: manuals-frontend.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    command: bin/rails s -P /tmp/rails.pid
+
+  manuals-frontend-app-draft:
+    <<: *manuals-frontend-app
```
leenagupte: Is this supposed to inherit from `manuals-frontend-app`  than `manuals-frontend` like the other images?

services/manuals-frontend/docker-compose.yml
sihugh: You've got to be quick on these, I've rebased 3 times in the last 5 minutes 😂

services/manuals-frontend/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    depends_on:
+      - router-app-draft
+      - content-store-app-draft
+      - static-app-draft
+      - static-app
```
leenagupte: Why do you need both `static-app` and `static-draft`?

services/manuals-frontend/docker-compose.yml
leenagupte: I saw that! Had to keep clicking refresh!

services/manuals-frontend/docker-compose.yml
sihugh: It's a bit of a toss up because of the inheritance, but I think I'll change it to what you suggest as it's possibly more expected.

services/manuals-frontend/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
+version: '3.7'
+
+x-manuals-frontend: &manuals-frontend
+  build:
+    context: .
+    dockerfile: services/manuals-frontend/Dockerfile
+  image: manuals-frontend
+  volumes:
+    - ..:/govuk:delegated
```
barrucadu: This is now `$GOVUK_ROOT_DIR:/govuk:delegated`

services/manuals-frontend/docker-compose.yml
sihugh: 🤷‍♂  yaml blindness?

barrucadu: It all blurs together after a while

services/manuals-frontend/docker-compose.yml
sihugh: Thanks @barrucadu, just as I thought it was safe to go into the repo...

barrucadu: Never underestimate my desire to suddenly change everything

services/manuals-frontend/docker-compose.yml
sihugh: Should be fixed up now @leenagupte 👍

PR #120, opened 2019-07-05
Make into a Gem

thomasleese:
This PR turns the `govuk-docker` CLI (and associated library) into a Gem. The main reason to do this is so that it's possible to install it using `gem install.`

This should mean that the installation steps will become only:

```
$ gem install govuk-docker
$ govuk-docker setup
```

And to run an app, all a user will need to do is:

```
$ cd ~/govuk/content-publisher
$ govuk-docker build
$ govuk-docker startup
```

This is the first time I've created a Gem so it's very possible I've missed something...


govuk-docker.gemspec
```diff
@@ -0,0 +1,30 @@
...
+  s.version      = GovukDocker::VERSION
+  s.platform     = Gem::Platform::RUBY
+  s.authors      = ["GOV.UK Dev"]
+  s.email        = ["govuk-dev@digital.cabinet-office.gov.uk"]
+  s.summary      = "Docker-based GOV.UK environment"
+  s.homepage     = "http://github.com/alphagov/govuk-docker"
+  s.description  = "A local environment for GOV.UK powered by Docker"
+
+  s.required_ruby_version = ">= 2.6.0"
+  s.files        = Dir.glob("exe/**/*") + Dir.glob("lib/**/*") + %w(README.md)
```
kevindew: I think you'd need services here too and the Makefile?

bevanloon:
Looking good.....I think the namespacing hasn't quite gone right though because I get this when I try to build and run locally:

```
~/govuk/govuk-docker  gemspec ✗                                                                                                                                           24m ◒
> gem build govuk-docker.gemspec
WARNING:  licenses is empty, but is recommended.  Use a license identifier from
http://spdx.org/licenses or 'Nonstandard' for a nonstandard license.
WARNING:  open-ended dependency on colorize (>= 0) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on thor (>= 0) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on climate_control (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on govuk-lint (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on pry (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on rspec (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: govuk-docker
  Version: 0.0.1
  File: govuk-docker-0.0.1.gem

~/govuk/govuk-docker  gemspec ✗                                                                                                                                           25m ◒
> gem install ./govuk-docker-0.0.1.gem
Successfully installed govuk-docker-0.0.1
Parsing documentation for govuk-docker-0.0.1
Installing ri documentation for govuk-docker-0.0.1
Done installing documentation for govuk-docker after 0 seconds
1 gem installed

~/govuk/govuk-docker  gemspec ✗                                                                                                                                           25m ◒
> govuk-docker
Traceback (most recent call last):
	11: from /Users/bevanloon/.rbenv/versions/2.6.3/bin/govuk-docker:23:in `<main>'
	10: from /Users/bevanloon/.rbenv/versions/2.6.3/bin/govuk-docker:23:in `load'
	 9: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/exe/govuk-docker:3:in `<top (required)>'
	 8: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	 7: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	 6: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/cli.rb:3:in `<top (required)>'
	 5: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/cli.rb:3:in `require_relative'
	 4: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/commands/build.rb:1:in `<top (required)>'
	 3: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/commands/build.rb:1:in `require_relative'
	 2: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/commands/base.rb:2:in `<top (required)>'
	 1: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/commands/base.rb:2:in `require_relative'
/Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/govuk-docker-0.0.1/lib/govuk_docker/paths.rb:1:in `<top (required)>': uninitialized constant GovukDocker (NameError)
```

thomasleese:
@bevanloon Thanks, I think I found the problem, and I've amended. Would you be able to try now?

govuk-docker.gemspec
```diff
@@ -0,0 +1,30 @@
+# -*- encoding: utf-8 -*-
+
+lib = File.expand_path("lib", __dir__)
+$:.unshift lib unless $:.include?(lib)
+
+require "govuk_docker/version"
+
+Gem::Specification.new do |s|
+  s.name         = "govuk-docker"
+  s.version      = GovukDocker::VERSION
```
erino: What about adding a licence here https://guides.rubygems.org/specification-reference/#license=

govuk-docker.gemspec
```diff
@@ -0,0 +1,30 @@
...
+  s.summary      = "Docker-based GOV.UK environment"
+  s.homepage     = "http://github.com/alphagov/govuk-docker"
+  s.description  = "A local environment for GOV.UK powered by Docker"
+
+  s.required_ruby_version = ">= 2.6.0"
+  s.files        = Dir.glob("exe/**/*") + Dir.glob("lib/**/*") + %w(README.md)
+  s.bindir       = "exe"
+  s.executables  = s.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  s.require_path = "lib"
+  s.add_dependency "colorize"
```
erino: `gem build` recommends using a bounded requirement here. Such as `~> 0.8`

bevanloon:
@thomasleese yep - getting there....
:) 
```
> gem install govuk-docker-0.0.1.gem
Successfully installed govuk-docker-0.0.1
Parsing documentation for govuk-docker-0.0.1
Installing ri documentation for govuk-docker-0.0.1
Done installing documentation for govuk-docker after 0 seconds
1 gem installed

~/govuk/govuk-docker  gemspec ✗                                                                                                                                           42m ◒
> govuk-docker
govuk-docker commands:
  govuk-docker be [ARGS]            # Alias for `run bundle exec`
  govuk-docker build                # Build the containers for the service, equivalent to running `make` in the `govuk-docker` directory
  govuk-docker compose ARGS         # Run `docker-compose` with ARGS in the `govuk-docker` environment
  govuk-docker doctor               # Various tests to help diagnose issues when running `govuk-docker`
  govuk-docker help [COMMAND]       # Describe available commands or one specific command
  govuk-docker prune                # Remove all docker containers, volumes and images
  govuk-docker run [ARGS]           # Run the container for a service, with option arguments
  govuk-docker setup                # Configures and installs the various dependencies necessary to run `govuk-docker` successfully
  govuk-docker startup [VARIATION]  # Run the container for a service in the `app` stack, with optional variations, such as `live` or `draft`

Options:
  [--service=SERVICE]          # The service, defaults to the name of the current working directory (for example `government-frontend`)
  [--stack=STACK]              # The stack of the service (for example `lite`)
                               # Default: lite
  [--verbose], [--no-verbose]  # If verbose, the docker-compose arguments will be displayed in full


~/govuk/govuk-docker  gemspec ✗                                                                                                                                           42m ◒
> govuk-docker doctor
Checking govuk-docker
Traceback (most recent call last):
	7: from /Users/bevanloon/govuk/govuk-docker/bin/govuk-docker:29:in `<main>'
	6: from /Users/bevanloon/govuk/govuk-docker/bin/govuk-docker:29:in `load'
	5: from /Users/bevanloon/govuk/govuk-docker/exe/govuk-docker:5:in `<top (required)>'
	4: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
	3: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
	2: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
	1: from /Users/bevanloon/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
/Users/bevanloon/govuk/govuk-docker/lib/govuk_docker/cli.rb:63:in `doctor': uninitialized constant GovukDocker::CLI::Doctor (NameError)
Did you mean?  GovukDocker::Doctor
```

govuk-docker.gemspec
thomasleese: Yeah, I thought about it, but I didn't want to put it in since I thought that might be better as an explicit PR which adds the licence. But perhaps we can just go ahead and give it the MIT licence?

govuk-docker.gemspec
thomasleese: I chose to leave them out because it currently matches the `Gemfile` so we get the same behaviour before and after. I was thinking this could be fixed in a subsequent PR, but perhaps I can just go ahead and fix it now in this PR?

thomasleese:
Thanks all for the review.

I've merged in the namespacing changes which I pulled out into a separate PR: https://github.com/alphagov/govuk-docker/pull/124. 

> I'm a bit concerned about the jump to a gem at this stage in this projects development. This would mean that for every change we'd have to be pushing out new versions of the gem. Given the current stability I'd also imagine that every person using this for a reasonable length of time will have to make changes I'd imagine most people would have to continue using it the non gem way (and run into problems of $PATH ordering).

Perhaps a better name for this Gem is `govuk-docker-cli` or perhaps we need to improve the documentation, but the actual services and `Makefile` wouldn't be included as part of this. When a user runs `govuk-docker setup` for the first time it will clone/pull their local `govuk-docker` repo and use the files in that. This prevents us having to release a new version of the Gem for every Docker change, only for changes in the CLI. I'm hopeful that the CLI itself is fairly stable now such that we won't need to have too many releases, but perhaps it's worth waiting a bit longer.

In terms of local development, you could run `bundle exec ./exe/govuk-docker` to run the development version of the CLI, rather than the `gem install`ed one.

> We'd also have to add some information that this is not to be installed via bundler.

Agreed, we can improve the documentation for this.

> Do we gain anything else other than the automatic path change?

This is the main reason we wanted this, it greatly simplifies how to install the CLI. Using `gem install` is also much more familiar to devs and the current instructions only work for `bash` whereas we'd need to include installation instructions for other shells.

govuk-docker.gemspec
erino: A subsequent PR works for me.

govuk-docker.gemspec
erino: There's already a [licence](b8caed810a8414b40239047c0802b3e52c2ec3af) file in the repo, so don't think we need a explicit PR to add a licence here.

benthorner:
I had a little think about this over the weekend, and came up with a couple of reasons we should hold back on this:

   - The local development process becomes confusing. Since part of the functionality of govuk-docker is directory-specific, people would need to uninstall the gem or alter their path in order to test local changes.
   - The upgrade process becomes confusing, since we're splitting the way people source the files they need. I think it would become confusing as to whether people need to update the gem or clone the repo, or both.

rbenv/ruby-build is the only software I can think of that takes a similar approach, and the result is pretty confusing (`rbenv install 1.2.3` -> "cannot find ruby version").

@thomasleese, perhaps we can go with your suggestion to wait on this a bit. Setting up the PATH is just a one-liner, after all. We could provide a few more examples for different shells/OSs?

tijmenb:
Nice, this looks good, but I'm worried about it accidentally causing more work than it's worth.

If I understand it correctly, the trade-off in gemifying is that instead of the path fiddling:

1. Developers have to manually update the gem. They also have to pull the git repo at the same time, because the gem depends on the structure of the repo.
1. Developers have to manually re-install the Ruby gem for every Ruby version bump
1. Developers have to manually install the Ruby gem for each Ruby they're in (the Ruby in my home directory is not up to date, so I had to do `rbenv global 2.6.3` to get the gem working). 
1. Related, we might run into issues with Ruby compatibility. At the moment we only run in the Ruby specified by the `.ruby-version` in the repo, but with a gem it could be anything.
1. It might be unclear when to release a new gem to developers without a lot of experience with the project. This might lead to not accidentally not releasing the code, which could be very confusing because developers will be looking at the repo to see what the features are, and it differing from the gem.
1. Local development becomes more confusing. I'm worried that it might not even work: because the scripts call `govuk-docker`, you run it from the repo I'm unclear what executable will be called.

thomasleese:
I'm going to close this for now as I don't think we're confident enough for it to go into master, perhaps we can revisit it when govuk-docker becomes more stable and is more commonly used across GOV.UK.

The main reason we decided to go this route is because our user research suggested that most people would have preferred to be able to install this using `gem install`. However, the points above are all valid, it does introduce some confusion, particularly in terms of the distinction between the CLI and the Dockerfiles.

In terms of local development, my understanding was that it shouldn't make much difference. `$ govuk-docker` would point to the "installed" version, while running `$ bundle exec ./exe/govuk-docker` would run your own local version with any changes you've made.

PR #121, opened 2019-07-05
Add phantomjs to whitehall

sihugh:
It is required to run tests


benthorner:
@sihugh would you mind holding off on this? It's actually covered by https://github.com/alphagov/govuk-docker/pull/93.

PR #125, opened 2019-07-05
Fix makefile dependencies and bundler versioning

benthorner:

Previously we installed bundler without a version restriction, which
caused problems since Bundler 2.0 is not compatible with most of our
repos. We may also need to cope with installing specific versions of
bundler for certain services. This splits the installation step for
bundler and ensures the major version is consistent with each service.

Previously we used '$@' in the target list for each service, which is
not valid (it can only be used within the recipe). This reverts the use
of '$@' back to the service name in each service Makefile.

### DIFF
--- a/services/asset-manager/Makefile
+++ b/services/asset-manager/Makefile
@@ -1,6 +1,7 @@
-asset-manager: $(GOVUK_ROOT_DIR)/$@
+asset-manager: $(GOVUK_ROOT_DIR)/asset-manager
 	$(GOVUK_DOCKER) compose build $@-lite
-	$(GOVUK_DOCKER) compose run $@-lite rbenv install -s && gem install --conservative bundler
+	$(GOVUK_DOCKER) compose run $@-lite rbenv install -s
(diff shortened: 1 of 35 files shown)


benthorner:
We definitely want to consolidate our Makefiles!

services/asset-manager/Makefile
```diff
@@ -1,6 +1,7 @@
-asset-manager: $(GOVUK_ROOT_DIR)/$@
+asset-manager: $(GOVUK_ROOT_DIR)/asset-manager
 	$(GOVUK_DOCKER) compose build $@-lite
-	$(GOVUK_DOCKER) compose run $@-lite rbenv install -s && gem install --conservative bundler
+	$(GOVUK_DOCKER) compose run $@-lite rbenv install -s
+	$(GOVUK_DOCKER) compose run $@-lite gem install --conservative bundler -v "~>1.0"
```
sihugh: as per the civil service code, I'm not sure we're allowed to show preference for a conservative bundler.

benthorner: How about `--liberal bundler`?

sihugh: 😑

PR #129, opened 2019-07-05
Setup email alert frontend

kevindew:
Trello: https://trello.com/c/Zp9f7DT7/59-set-up-email-alert-frontend

This sets up Email Alert Frontend to run. I hit a nasty pain point with CSRF tokens where I needed to update the nginx config, more details in the commits.

### DIFF
--- a/services/email-alert-frontend/Makefile
+++ b/services/email-alert-frontend/Makefile
@@ -1,3 +1,7 @@
-email-alert-frontend: $(GOVUK_ROOT_DIR)/email-alert-frontend
+email-alert-frontend: $(GOVUK_ROOT_DIR)/email-alert-frontend email-alert-api
 	$(GOVUK_DOCKER) compose build $@-lite
 	$(GOVUK_DOCKER) compose run $@-lite bundle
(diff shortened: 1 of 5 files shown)


services/email-alert-frontend/Makefile
```diff
@@ -0,0 +1,7 @@
+email-alert-frontend: $(GOVUK_ROOT_DIR)/email-alert-frontend email-alert-api
+	$(GOVUK_DOCKER) compose build $@-lite
+	$(GOVUK_DOCKER) compose run $@-lite bundle
+	$(GOVUK_DOCKER) compose run $@-setup bin/rake publishing_api:publish_email_signup_page
```
tijmenb: We should find some way of doing this in the app.

services/email-alert-frontend/Makefile
```diff
@@ -0,0 +1,7 @@
+email-alert-frontend: $(GOVUK_ROOT_DIR)/email-alert-frontend email-alert-api
+	$(GOVUK_DOCKER) compose build $@-lite
```
benthorner: Little bit late here, but I think we're missing the new rbenv/gem steps :-|.

PR #130, opened 2019-07-08
Fix clone dependency for router

barrucadu:


benthorner:
Dah - thanks for spotting @barrucadu !

PR #131, opened 2019-07-08
Ignore `docker volume rm $(...)` failing

barrucadu:
This will fail if there are no dangling volumes.  That's not an error
condition, it's a reasonable thing to happen.

### DIFF
--- a/lib/govuk_docker/commands/prune.rb
+++ b/lib/govuk_docker/commands/prune.rb
@@ -10,7 +10,7 @@ private
   def commands
     [
       "docker container prune -f",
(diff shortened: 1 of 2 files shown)


thomasleese:
There's also https://github.com/alphagov/govuk-docker/pull/128 open which I think solves the same problem. Do you think it matters if we ignore the errors for any of these commands?

barrucadu:
I don't think it matters

benthorner:
I've merged my one in now - it has the minor advantage that we're ignoring errors in a very explicit way ;-).

PR #135, opened 2019-07-08
Add search-admin

tijmenb:
Confirmed that the tests pass, and the app runs in the browser.

Screen Shot 2019-07-08 at 14 33 16

https://trello.com/c/KZvfH32h

### DIFF
--- a/services/search-admin/docker-compose.yml
+++ b/services/search-admin/docker-compose.yml
@@ -1,0 +1,40 @@
+version: '3.7'
+
+x-search-admin: &search-admin-base
+  build:
+    context: .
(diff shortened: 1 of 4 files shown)


services/search-admin/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+  working_dir: /govuk/search-admin
+
+services:
+  search-admin-lite:
+    <<: *search-admin-base
+    privileged: true
+    depends_on:
+      - mysql
+    environment:
+      # the app uses this URL to generate a test database URL
```
benthorner: Minor, but I think this comment is for another service ;-).

PR #138, opened 2019-07-09
Don't show a stacktrace for command errors

tijmenb:
When a system command fails, it's not useful to see the stack trace since it's always the same. This removes the stack trace by using the built-in Thor error class, which only prints the error message.

## Before

Screen Shot 2019-07-09 at 14 45 32

## After

Screen Shot 2019-07-09 at 14 45 18


erino:
What about suggesting that you run `govuk-docker doctor` to see what the problem might be?

tijmenb:
@erino good idea! PRs are welcome ;)

PR #139, opened 2019-07-09
Update installation and usage instructions

tijmenb:
👉  [Rendered version](https://github.com/alphagov/govuk-docker/blob/usage-and-installation/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,58 +52,19 @@ Now in the `~/govuk` directory, run the following commands.
 git clone git@github.com:alphagov/govuk-docker.git
 cd govuk-docker
 bundle install
-```
-
-You can now clone and setup the apps you need with `make APP-NAME`,
-for example:
-
-```
-make content-publisher government-frontend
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,58 +52,19 @@ Now in the `~/govuk` directory, run the following commands.
...
-make content-publisher government-frontend
-```
-
-The govuk-docker command can configure Dnsmasq for you:
-
-```
 govuk-docker setup
 ```
 
-If this doesn't work for whatever reason, follow the instructions below to
```
thomasleese: I think this we should still keep these instructions but in a place further down, perhaps in the troubleshooting section? There may be times when the `setup` command doesn't work exactly, and having to look through the code to figure out what it does might be annoying.

ChrisBAshton:
Have now pulled this into https://github.com/alphagov/govuk-docker/pull/141 as requested 👍

PR #141, opened 2019-07-09
Improve README

ChrisBAshton:
1. Move manual Dnsmasq instructions into standalone section
In a best case scenario, devs shouldn't need to follow these instructions,
so let's move them out of the main Installation section and into the How
To section.

2. Clarify command parameters (especially `govuk-docker startup`)

3. Simplify setup instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,16 +16,42 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
 
 ## Usage
 
+To run a GOV.UK application, `govuk-docker build` it first (you only need to do this once per application). We'll use collections-publisher as an example:
+
+```sh
+cd ~/govuk/collections-publisher
+govuk-docker build
+```
+
+Then you run it with `govuk-docker startup`:
+
+```sh
+# Start collections-publisher including dependencies. Visit it at collections-publisher.dev.gov.uk
+govuk-docker startup
+```
+
+govuk-docker knows which application we're running based on the name of the current directory, matching it with the [corresponding docker-compose.yml in govuk-docker/services](https://github.com/alphagov/govuk-docker/blob/master/services/content-tagger/docker-compose.yml)). Alternatively, you can build and run applications from any directory by specifying a `service` CLI parameter:
+
+```sh
+govuk-docker build --service collections-publisher
+govuk-docker startup --service collections-publisher
+```
+
+`govuk-docker startup` runs the application on the `app` [stack](#stacks) by default, but you can override the stack by passing an unnamed parameter which will be taken as the stack name, with an `app-` prefix. Example:
+
+```sh
+# Start content-publisher plus the "app-e2e" (end-to-end) stack
+govuk-docker startup --service content-publisher e2e
+```
+
+Another useful govuk-docker command is `run`:
+
 ```sh
 # Run a Rake task on Whitehall
-whitehall$ govuk-docker run bundle exec rake -T
-
-# Start content-tagger including dependencies. Visit it at content-tagger.dev.gov.uk
-content-tagger$ govuk-docker startup
-
-# Start content-publisher plus an "end-to-end" stack
-content-publisher$ govuk-docker startup e2e
+cd ~/govuk/whitehall && govuk-docker run bundle exec rake -T
 ```
+
+For a full list of govuk-docker commands, run `govuk-docker help`.
 
 ## Installation
 
@@ -206,6 +232,8 @@ make pull
 ```
 
 ### How to: set up Dnsmasq manually
+
+If the [installation instructions](#setup) above didn't work for you, you may need to do some things manually as outlined below.
 
 If you have been using the vagrant based dev vm, take a backup
 of  `/etc/resolver/dev.gov.uk`.
--- a/README.md
+++ b/README.md
@@ -219,19 +219,23 @@ Then create or update `/etc/resolver/dev.gov.uk`. If you've been using the vagra
 ```
 nameserver 127.0.0.1
 ```
+
 To check if the new config has been applied, you can run `scutil --dns` to check that `dev.gov.uk` appears in the list.
 
-Then append the following to the bottom of `/usr/local/etc/dnsmasq.conf`
+Then append the following to the bottom of `/usr/local/etc/dnsmasq.conf`:
+
 ```
 conf-dir=/usr/local/etc/dnsmasq.d,*.conf
 ```
 
-Then create or append to `/usr/local/etc/dnsmasq.d/development.conf`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,17 +16,37 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
 
 ## Usage
 
+To run a GOV.UK application, `make` it first (you only need to do this once per application):
```
tijmenb: This isn't necessary. If you do `govuk-docker build` in the app it will build the app + dependencies.

You can also do this from any directory:

```
govuk-docker build --service collections-publisher
```

README.md
```diff
@@ -16,17 +16,37 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
...
+# This assumes you're in the ~/govuk/govuk-docker directory.
+$ make content-tagger
+```
+
+Then run it with `startup`:
 
+```sh
 # Start content-tagger including dependencies. Visit it at content-tagger.dev.gov.uk
-content-tagger$ govuk-docker startup
+$ govuk-docker startup --service=collections-publisher
```
tijmenb:
```suggestion
$ govuk-docker startup --service collections-publisher
```

README.md
```diff
@@ -16,17 +16,37 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
...
+To run a GOV.UK application, `make` it first (you only need to do this once per application):
+
 ```sh
-# Run a Rake task on Whitehall
-whitehall$ govuk-docker run bundle exec rake -T
+# This assumes you're in the ~/govuk/govuk-docker directory.
+$ make content-tagger
+```
+
+Then run it with `startup`:
```
tijmenb:
```suggestion
Then run it with `govuk-docker startup`:
```

README.md
```diff
@@ -16,17 +16,37 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
...
+```sh
+# Run a Rake task on Whitehall
+$ cd ~/govuk/whitehall && govuk-docker run bundle exec rake -T
+```
+
+For a full list of govuk-docker commands, run `govuk-docker help`.
+
 ## Installation
 
 ### Prerequisites
```
tijmenb: I think we can drop the prerequisites, as they're installed by `govuk-docker setup`?

README.md
```diff
@@ -182,6 +165,21 @@ the app. To provide consistency we have a convention for these names:
   - **app-e2e**: to run the app with all the other apps necessary to provide
     full end to end user journeys.
 
+`govuk-docker startup` defaults to the `app` stack, but can be overridden in two ways:
```
tijmenb: I'd move this into the usage section if needed. The stack section in here is only to document the stack name conventions.

README.md
```diff
@@ -43,7 +63,7 @@ First make sure the following are installed on your system:
 Start with the following in your bash config.
 
 ```
-export PATH=$PATH:~/govuk/govuk-docker/bin
+export PATH=$PATH:/usr/local/sbin:~/govuk/govuk-docker/bin
```
thomasleese: I've opened a PR #143 so users shouldn't have to do this.

README.md
```diff
@@ -182,6 +165,21 @@ the app. To provide consistency we have a convention for these names:
...
+2. Or you can specify a named `--stack` parameter.
+
+Example:
+
+```sh
+# Start content-publisher plus an "end-to-end" stack
+$ govuk-docker startup --service=content-publisher e2e
+
+# This command does the same thing as above
+$ govuk-docker startup --service=content-publisher --stack=app-e2e
```
thomasleese: I think we want to discourage people using `startup` with a stack which isn't an `app` prefixed stack. Since `startup` is really just a thin wrapper around `run`, people should be using `run` in most situations.

I think ideally the command wouldn't even work if you pass `--stack`, it's more of an accident that it does due to the way we parse arguments across all commands in the same way.

README.md
```diff
@@ -242,6 +240,48 @@ Sometimes it's useful to get all changes for all repos e.g. to support finding t
 make pull
 ```
 
+### How to: set up Dnsmasq manually
```
thomasleese: I wonder if we should add a line to the top of this section with something like "This is only necessary if `govuk-docker setup` doesn't work" in case someone spots this without knowing about `govuk-docker setup`.

But I like that we've still got this here because there are situations where the `setup` command won't work.

ChrisBAshton: Yep, I had to follow the "Then create or append to `/usr/local/etc/dnsmasq.d/development.conf`:" step personally.

ChrisBAshton: Now added a line 👍

README.md
ChrisBAshton: Cool, thanks 👍

README.md
ChrisBAshton: Removed the line

README.md
ChrisBAshton: I've given it an update 👍

README.md
```diff
@@ -16,28 +16,55 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
...
 
 ### Prerequisites
 
-First make sure the following are installed on your system:
+govuk-docker has the following dependencies:
 
-- [dnsmasq](http://www.thekelleys.org.uk/dnsmasq/doc.html) to make *$app-name.dev.gov.uk* work. You can install this using `brew install dnsmasq`.
-- [docker](https://hub.docker.com/) and [docker-compose](https://docs.docker.com/compose/install/)
-- [git](https://git-scm.com) if you're setting everything up from scratch
+- [brew](https://brew.sh/)
```
thomasleese: There are some developers on GOV.UK who don't use a Mac, in which case installing Brew won't be possible for them, but `govuk-docker` itself will still work. For them, they'll have to manually install dnsmasq themselves. It might be good to add something about that here.

README.md
```diff
@@ -16,28 +16,55 @@ To guide development we [have documented the user needs](docs/NEEDS.md) and [ass
 
 ## Usage
 
+To run a GOV.UK application, `govuk-docker build` it first (you only need to do this once per application):
+
 ```sh
-# Run a Rake task on Whitehall
-whitehall$ govuk-docker run bundle exec rake -T
+# This assumes you're in the ~/govuk/govuk-docker directory.
```
thomasleese: The following command can actually be run from any directory - and I think in general I think we would expect most people to run it (and the other commands) from the directory of the app itself to avoid having to pass `--service` around. I think it's still a good idea to publicise the `--service` option, but my making it the first thing people see, they may assume you have to do it this way and not read on.

PR #151, opened 2019-07-19
Use publishing-api-app-e2e in collections-publisher

leenagupte:
We need publishing-api-app-e2e so that the worker is started and we can test
pushing updates to publishing-api

### DIFF
--- a/services/collections-publisher/docker-compose.yml
+++ b/services/collections-publisher/docker-compose.yml
@@ -32,7 +32,6 @@ services:
       - link-checker-api-app
       - nginx-proxy-app
       - publishing-api-app
-      - collections-publisher-worker
     environment:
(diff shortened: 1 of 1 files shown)


benthorner:
I wonder if we should consider having a different stack here? `publishing-api-app-e2e` is **very** expensive - is it possible to use (most of) collections-publisher with a lighter dependency?

[Content Publisher](https://github.com/alphagov/govuk-docker/blob/master/services/content-publisher/docker-compose.yml#L43) makes a distinction for this very reason: most of the time we can meet our development needs with a smaller resource overhead.

leenagupte:
@benthorner I have made `collections-publisher-app-e2e` inherit from `collections-publisher-app`, so I assume that means I don't have to re-add mysql and redis as dependencies?

PR #156, opened 2019-07-26
Add healthchecks for backing services

barrucadu:
By adding a healthcheck we ensure that containers which rely on these
services don't come up until the service reports a healthy state.
e.g. search-api won't be available to do things before elasticsearch
is running.  This is useful because at the moment there is a race
condition in commands like this:

    GOVUK_DOCKER_SERVICE=search-api \
    govuk-docker run curl -XDELETE http://elasticsearch5:9200/_all

As the containers are started at the same time, without waiting, there
is a good chance the `curl` will happen before Elasticsearch is able
to respond to requests.  A healthcheck fixes this.

These checks were copied from the e2e tests.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,9 +11,16 @@ volumes:
   elasticsearch5:
   elasticsearch6:
 
+x-default-healthcheck: &default-healthcheck
+  interval: 10s
+  timeout: 20s
+
 services:
   postgres:
     image: postgres:9.6
+    healthcheck:
+      << : *default-healthcheck
+      test: "psql --username 'postgres' -c 'SELECT 1'"
     volumes:
       - postgres:/var/lib/postgresql/data
 
@@ -22,6 +29,9 @@ services:
 
   mongo:
     image: mongo:2.4
+    healthcheck:
+      << : *default-healthcheck
+      test: "echo 'db.stats().ok' | mongo localhost:27017/test --quiet"
     volumes:
       - mongo:/data/db
     ports:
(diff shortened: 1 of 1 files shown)


barrucadu:
Looking at the publishing-e2e-tests, it looks like it doesn't have any healthchecks for apps we can copy: https://github.com/alphagov/publishing-e2e-tests/blob/master/docker-compose.yml

But we could certainly add something like hitting a known path and checking it gets a 200 response

benthorner:
How about `/healthcheck`? 🆗

kevindew:
@barrucadu I may be misunderstanding, but it looks like you're using these healthchecks as a means to wait for services to be healthy, is that correct?

If so, docker-compose doesn't do any waiting for services to be healthy so these checks introduced would only provide extra information when you use the `docker ps` command. Docker-compose did used to provide wait functionality in [v2 of docker-compose](https://docs.docker.com/compose/compose-file/compose-file-v2/#depends_on) but it was removed in v3, much to the annoyance of every developer that works with a local set up.

In publishing-e2e-tests there is [code that monitors service health](https://github.com/alphagov/publishing-e2e-tests/blob/e3bd58385a4e468e6b21a3c1eafc373c47d9c7dd/lib/docker_service.rb#L5) to perform this waiting. Though I'd wonder if we'd want to take one of the simpler and more common approaches of using scripts like [wait-for-it](https://github.com/vishnubob/wait-for-it) or [dockerize](https://github.com/jwilder/dockerize) than copying the bespoke approach used in end-to-end tests. The e2e tests approach was built only after recently discovering the healthcheck wait functionality had been dropped.

barrucadu:
Ah, I thought it was still there.  In that case it's just coincidence this has been working since I made the change 😢

PR #157, opened 2019-07-29
Add a "replication" volume to ES containers

barrucadu:
This is so we can download an ES snapshot from integration, copy it
into the container, and then restore the snapshot from the local
filesystem.  Restoring directly from s3 is undesirable because these
snapshots can be several gigabytes, and if the connection drops during
the restore it has to be started again.

The ES configuration files are needed because the path.repo option
can't be set by environment variables.


config/elasticsearch5.yml
```diff
@@ -0,0 +1,5 @@
+cluster.name: "docker-es5-cluster"
```
benthorner: Could these live under 'services/elasticsearchX', like we do for other services?

barrucadu:
[Here is the script I've got working](https://github.com/barrucadu/dotfiles/blob/master/misc/bin/govuk-scripts/docker-sync.sh), which is based on [the script for the dev VM](https://github.com/alphagov/govuk-puppet/blob/master/development-vm/replication/sync-aws-elasticsearch.sh
).  The key bits are [lines 52 to 75]( https://github.com/barrucadu/dotfiles/blob/master/misc/bin/govuk-scripts/docker-sync.sh#L52-L75):

```shell
# `docker cp` reliably hangs after copying 6.4GB, so instead of
# using that spin up a container with the host directory
# bind-mounted and run `cp` inside the container.
local replication_dir="/replication/elasticsearch-${ver}"
govuk docker compose run --rm -v $archive_path:/import "elasticsearch${ver}" bash -xc "
  [[ -e '${replication_dir}' ]] && rm -r '${replication_dir}'
  cp -a /import '${replication_dir}'
"

govuk docker run search-api lite bash -xc "
  es='http://elasticsearch${ver}:9200'
  curl -XDELETE \"\${es}/_all\"
  curl \"\${es}/_snapshot/snapshots\" -X PUT -H 'Content-Type: application/json' -d '{
    \"type\": \"fs\",
    \"settings\": {
      \"compress\": true,
      \"readonly\": true,
      \"location\": \"${replication_dir}\"
    }
  }'
  snapshot_name=\$(curl \"\${es}\"/_snapshot/snapshots/_all | ruby -e 'require \"json\"; STDOUT << (JSON.parse(STDIN.read)[\"snapshots\"].map { |a| a[\"snapshot\"] }.sort.last)')
  curl -XPOST \"\${es}/_snapshot/snapshots/\${snapshot_name}/_restore?wait_for_completion=true\"
"
```

The replication script is run inside the container, there's a temporary volume mounted to `/import` which is used to copy the replicated data from the host to the `replication` volume.  I thought about using the same volume for each service which needs replication, but there's no particular need for that to be the case - I'm open to having separate volumes with more specific names

benthorner:
Since we remove everything in `/replication` during the script, it seems strange to keep it around as a permanent volume. Does it kill performance to do the restore directly from the bind-mounted `/import` volume? Or perhaps there's some other use of `/replication` I'm missing.

barrucadu:
I'll have a go at modifying my script to not need a persistent volume

barrucadu:
I've got a mostly-working script which doesn't need a new volume, or configuration files in this repository (though it does generate a config file when it runs).  I'll close this PR.

PR #160, opened 2019-07-29
Create directory before writing file as sudo

bevanloon:
If a directory does not exist before writing a file, setup will error
out.

Add function to create the directory if it does not yet exist.

Since one of the directories we need to create is /etc/resolver we need
to allow for creating the directory as `sudo`. To allow for this, add
`as_sudo` as a parameter to existing `write_file`. This defaults to
false.

We also need to create the file (/etc/resolver/dev.gov.uk) as sudo, so 
adapt the code to take that into account.

### DIFF
--- a/lib/govuk_docker/setup/dnsmasq.rb
+++ b/lib/govuk_docker/setup/dnsmasq.rb
@@ -30,6 +30,22 @@ private
     shell.yes?("Are you sure you want to continue?")
   end
 
+  def check_continue_as_sudo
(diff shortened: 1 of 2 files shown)


issyl0:
I just encountered this error and this branch fixes it so that `govuk-docker setup` continues. 👍from me.

bevanloon:
Thanks @issyl0 ....I need to make some time to sort out the tests for this.

PR #161, opened 2019-08-02
Rename bin/bundle to resolve hijacking bundler

bevanloon:
When govuk-docker preceeds your chosen ruby version manager in your
path, it will hijack the `bundle` command, causing unexpected behaviour.

Rename the docker specific binstub to avoid this.


erino:
It works on my machine...

kevindew:
Ah ha I flagged this as an issue a while ago: https://trello.com/c/2bxWjW1c/95-we-should-have-binstubs-if-we-expect-the-bin-directory-to-be-in-a-users-path

I don't think we need the bundle binstub to be honest since bundle is installed by default by Ruby for the version this repo is using.

PR #163, opened 2019-08-05
Add MySQL command to increase max allowed packet size

emmabeynon:
Importing replicated data into Whitehall's database has proved problematic as
the default max_allowed_packet value for MySQL isn't large enough. Previously
we tried adding max_allowed_packet as an environment variable but this didn't
work. Instead we are passing it as a command argument, noting that the
entrypoint file for MySQL looks for this argument and prepends it with
"mysqld".

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,9 +45,9 @@ services:
       test: "mysql --user=root --password=root -e 'SELECT 1'"
     volumes:
       - mysql:/var/lib/mysql
+    command: --max_allowed_packet=1073741824
     environment:
       MYSQL_ROOT_PASSWORD: root
-      max_allowed_packet: 1073741824
 
   redis:
     image: redis


benthorner:
This looks good. I'd like us to try and avoid changing the command if we can, though: if someone needs to override the mysql command, they'll have to remember to include this flag. As with the Rails commands, the idea is to try and avoid any surprising flags and use other techniques.

It looks like there's another way to get around this, by [specifying a custom config file](https://hub.docker.com/_/mysql). This is similar to what we used to (have to) do with the `.chromedriver-version` file: [example here](https://github.com/alphagov/govuk-docker/pull/20/files).

The advantage of this approach is that it allows us to change other options in future, without ending up with a humungous command.

benthorner:
A more current example would be how we add more config for the nginx-proxy service, which more clearly demos the precedent for this sort of thing: https://github.com/alphagov/govuk-docker/blob/master/services/nginx-proxy/docker-compose.yml#L10

kevindew:
This looks good to me. I would think this is fine to go in as is since it's essentially a bug fix, is consistent with the intentions of the MySQL image entrypoint and the low likelihood of someone running mysqld a different way.

I do get your concerns @benthorner, Emma and I started working on this from a config file approach but hit a few things that pushed it beyond a no-brainer. So most consistent route would be to create a mysql service directory, thereby having that separate from the other persistent datastores? That'd break the healthcheck shared data but that doesn't seem a big concern to repeat. I guess the nicest way for the config would to have a `services/mysql/mysql.cnf` that was set as a volume to `/etc/mysql/conf.d/govuk-docker.cnf`

benthorner:
@kevindew it sounds like you've tried a few alternatives before arriving at this. On the surface, it seems unlikely someone would override it, but then what about data import?

Having a `services/mysql` directory with any necessary config files makes sense to me. I don't understand how that would impact on the data volume, since they're mounted in separate places. On the whole, it doesn't seem to matter that much, but I'm interested to know the specific reason(s) why we can't use the same approach as for other services, like nginx-proxy.

kevindew:
> @kevindew it sounds like you've tried a few alternatives before arriving at this. On the surface, it seems unlikely someone would override it, but then what about data import?
> 
> Having a `services/mysql` directory with any necessary config files makes sense to me. I don't understand how that would impact on the data volume, since they're mounted in separate places. On the whole, it doesn't seem to matter that much, but I'm interested to know the specific reason(s) why we can't use the same approach as for other services, like nginx-proxy.

data import isn't a concern as that runs a mysql client rather than the server. 

We can mount data volumes in a different docker-compose file they merge like services do.

@emmabeynon if you're interested in setting up the config file approach I can give you a hand.

emmabeynon:
@kevindew @benthorner I know this problem is blocking quite a few other people so can I suggest merging this for now to fix it quickly, and opening up a follow-up PR to implement Ben's approach?  I'm new to Docker so it'll probably take me quite a bit of time to look into setting up a config file.

PR #165, opened 2019-08-06
Remove redundant healthchecks

benthorner:
This removes the healthchecks for global services, since we established
they have little to no benefit when used with docker-compose.

Original PR: https://github.com/alphagov/govuk-docker/pull/156


benthorner:
@barrucadu this is my current understanding. Has the situation changed? If they are indeed redundant, then I think we should remove them until we can put a consistent approach in place that can work across the apps.

PR #166, opened 2019-08-07
Specify port in /etc/resolver/dev.gov.uk

benthorner:
Previously we only encouraged people to have a /etc/resolver/dev.gov.uk
    file with a single 'namespace' line. Just specifying 'namespace' seems
    to cause problems for people moving from a 'vagrant dns' setup, where an
    additional 'port 5300' line was also present. Simply removing this line
    doesn't seem to be enough to make the OS notice the file has changed, so
    this explicitly specifies 'port 53', even though 53 is the default DNS
    port, to avoid people having to restart their machine to get the changes.

### DIFF
--- a/lib/govuk_docker/doctor/checkup.rb
+++ b/lib/govuk_docker/doctor/checkup.rb
@@ -87,7 +87,7 @@ module GovukDocker::Doctor
     end
 
     def dnsmasq_resolver?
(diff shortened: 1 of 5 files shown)


benthorner:
@kevindew thanks for the comment - I've change commit and PR messages to make this clearer.

PR #168, opened 2019-08-08
Allows comments and newlines in dnsmasq resolver

### LINKED ISSUE #153: `govuk-docker doctor` dnsmasq check fails on comments in /etc/resolver/...
If you have something like this in `/etc/resolver/dev.gov.uk`:

```
# some comment
nameserver 127.0.0.1
```

...then `govuk-docker doctor` fails with:

```
❌ Something else (possibly Vagrant-dns) is configured to resolve DNS requests for dev.gov.uk.
          The /etc/resolver/dev.gov.uk file needs to be updated to only contain
          nameserver 127.0.0.1
```

It should ignore comments, as we can save new joiners a few moments of confusion by not erroring on comments. Comments were in `/etc/resolver/dev.gov.uk` by default on mine and @AgaDufrat's machines, so I suspect a lot of devs come across this.

ChrisBAshton:
Fixes #153


benthorner:
Good to get this sorted - it's been a pain for a few people 👍. I wonder if we can do something a bit simpler if we did a regex comparison, instead of '=='? If we're allowing comments, that means we're happy for people to have a 'customised' version of the file, so perhaps we just want to be more lenient with this check? Also, [my PR](https://github.com/alphagov/govuk-docker/pull/166) will break your PR 😆 .

ChrisBAshton:
In mine & Aga's case it was literally just a comment that broke `govuk-docker doctor`, we had no other DNS entries. Allowing additional entries may not be something we want to encourage - I don't know enough about possible use cases. This PR aims for feature parity in that sense, just ignoring comments and newlines.

Looks like your PR could build onto mine (introducing something like `allowed_values = ["nameserver 127.0.0.1", "port 53"]`) and checking each of the `dns_lines` is an allowable value. We could probably use regex for that but I don't know how permissive we'd want it to be 😀

benthorner:
Hmmm...tricky. Thinking about it a bit more, the current behaviour actually seems pretty good: the comment it's complaining about doesn't make sense - the file *isn't* generated by vagrant-dns - so while it's annoying, it also seems reasonable to not give the file a clean bill of health ;-).

What about tweaking the [failure message](https://github.com/alphagov/govuk-docker/blob/master/lib/govuk_docker/doctor/doctor.rb#L69) so it's a bit more general e.g. "your DNS configuration looks a bit iffy"?

ChrisBAshton:
Something like this...?

> ⚠️ Your DNS configuration contains some unexpected content.
> govuk-docker only expects /etc/resolver/dev.gov.uk to contain one line: `nameserver 127.0.0.1`
> Update /etc/resolver/dev.gov.uk to remove any extra lines, or proceed at your own risk.

I agree this is better than what's on `master` currently 👍 though it probably underplays the error if the dns configuration is _actually_ broken (missing `nameserver 127.0.0.1`). Would be nice to have a noisier error state when that happens:

> ❌ Your DNS configuration is missing `nameserver 127.0.0.1`.
> Without this, govuk-docker will not work!
> Update /etc/resolver/dev.gov.uk to contain `nameserver 127.0.0.1`.

This could probably be accomplished by adding a `dnsmasq_resolver_strict` checker for the former 🤔

benthorner:
I like the sound of that, although it gets a bit prescriptive towards the end - the number of lines isn't particularly important, noting that my PR will add another 😆.

There may be something of an elephant in the room here. It looks like the doctor stuff was written with a generic aim, but very specific things (about dnsmasq) have crept in. Longterm, we probably want to separate the groups of checks to give us more room to expand. With this space, we could add more accurate/useful checks, like an `nslookup` to see if the resolution *actually* works. I wonder if that might allay your concerns about the strictness of this check?

ChrisBAshton:
I'd be inclined to change the dnsmasq_resolver test _now_ to to see if the resolution actually works - if that's possible! Then it doesn't really matter what else is in the config 😄 but I've no idea how involved that is.

I may be able to spend an hour on it today. Any pointers appreciated!

ChrisBAshton:
Have discussed with Ben IRL and we've decided to close this PR on the basis that govuk-docker should write the file for you and you should never need to customise its contents (thereby giving us complete control over the presence/absence of comments and newlines).

The root cause of the problem is that `govuk-docker setup` doesn't overwrite the DNS config file if it already exists, so `govuk-docker doctor` then flags up issues. We should make the `setup` command overwrite the file in the first instance, that way we can guarantee the file is as we expect it to be.

PR #170, opened 2019-08-13
Add instructions for replicating data locally

brucebolt:
These are very rough and ready, but should help us solve the problem of replicating data properly until the old replication scripts can be repurposed for use with govuk-docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,76 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
 app.dev.gov.uk.		0	IN	A	127.0.0.1
 ```
 
+### How to: replicate data locally
+
+There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples reqire pv, which can be installed on a Mac using Brew (`brew install pv`).
+
+#### MySQL
+
+1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/mysql/?region=eu-west-1&tab=overview)
+
+2. Drop and recreate any existing database, e.g. for Whitehall:
+
+```
+govuk-docker compose up -d mysql
+govuk-docker compose run mysql mysql -h mysql -u root --password=root -e "DROP DATABASE IF EXISTS whitehall_development"
+govuk-docker compose run mysql mysql -h mysql -u root --password=root -e "CREATE DATABASE whitehall_development"
+```
+
+3. Import the file into the local MySQL database, e.g. for Whitehall:
+
+```
+pv whitehall_production.dump.gz | gunzip | govuk-docker compose run mysql mysql -h mysql -u root --password=root whitehall_development
+```
+
+#### PostgreSQL
+
+1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/postgres/?region=eu-west-1&tab=overview)
+
+2. Drop and recreate any existing database, e.g. for Publishing API:
+
+```
+govuk-docker compose up -d postgres
+govuk-docker compose run postgres /usr/bin/psql -h postgres -U postgres -qAt DROP DATABASE IF EXISTS "publishing-api"
+govuk-docker compose run postgres /usr/bin/createdb -h postgres -U postgres publishing-api
+```
+
+3. Import the file into the local Postgres database, e.g. for Publishing API:
+
+```
+pv publishing_api_production.dump.gz  | gunzip | govuk-docker compose run postgres /usr/bin/psql -h postgres -U postgres -qAt -d publishing-api
+```
+
+#### MongoDB
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -256,6 +256,58 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
+
+#### PostgreSQL
+
+1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/postgres/?region=eu-west-1&tab=overview)
+
+2. Import the file into the local Postgres database, e.g. for Publishing API:
+
+
+```
+pv publishing_api_production.dump.gz  | gunzip | docker exec -i `docker ps | grep postgres | awk '{print $1}'` /usr/bin/psql -U postgres -qAt -d publishing-api
```
erino: I think the database should be dropped and then created, otherwise not all of the data is imported. Something like this should work

```bash
docker exec -i `docker ps | grep postgres | awk '{print $1}'` /usr/bin/psql -U postgres -qAt DROP DATABASE IF EXISTS "publishing-api"
docker exec -i `docker ps | grep postgres | awk '{print $1}'` /usr/bin/createdb -U postgres publishing-api
```

README.md
```diff
@@ -256,6 +256,58 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
+pv whitehall_production.dump.gz | gunzip | docker exec -i `docker ps | grep mysql | awk '{print $1}'` mysql -u root --password=root whitehall_development
+```
+
+#### PostgreSQL
+
+1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/postgres/?region=eu-west-1&tab=overview)
+
+2. Import the file into the local Postgres database, e.g. for Publishing API:
+
+
```
ChrisBAshton: extra newline here

kevindew:
This looks super useful docs.

Couple of things that'd be useful to add: You need to install pipeviewer and it doesn't state that you need to have the db containers already running.

I also think you can also make the commands to run the db queries simpler by using compose.

For example: 
```
docker exec -i `docker ps | grep postgres | awk '{print $1}'` /usr/bin/psql -U postgres -qAt -d publishing-api
```
could be 

```
govuk-docker compose run postgres psql -U postgres -qAt -h postgres -d publishing-api
```

and you could cover the starting of the db and the command with a single line: 
```
govuk-docker compose up -d postgres && govuk-docker compose run postgres psql -U postgres -qAt -h postgres -d publishing-api
```

README.md
```diff
@@ -256,6 +256,69 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
+```
+govuk-docker compose up -d postgres
+govuk-docker compose run postgres /usr/bin/psql -U postgres -qAt DROP DATABASE IF EXISTS "publishing-api"
+govuk-docker compose run postgres /usr/bin/createdb -U postgres publishing-api
+```
+
+3. Import the file into the local Postgres database, e.g. for Publishing API:
+
+```
+pv publishing_api_production.dump.gz  | gunzip | govuk-docker compose run postgres /usr/bin/psql -U postgres -qAt -d publishing-api
```
kevindew: You'll want to specify the host as postgres with `-h postgres`

README.md
```diff
@@ -256,6 +256,69 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
+#### PostgreSQL
+
+1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/postgres/?region=eu-west-1&tab=overview)
+
+2. Drop and recreate any existing database, e.g. for Publishing API:
+
+```
+govuk-docker compose up -d postgres
+govuk-docker compose run postgres /usr/bin/psql -U postgres -qAt DROP DATABASE IF EXISTS "publishing-api"
+govuk-docker compose run postgres /usr/bin/createdb -U postgres publishing-api
```
kevindew: Might be simpler to do this with rails - though not sure we need it if we're not doing this for any other dbs?

README.md
```diff
@@ -256,6 +256,69 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
+
+#### MySQL
+
+1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/mysql/?region=eu-west-1&tab=overview)
+
+2. Import the file into the local MySQL database, e.g. for Whitehall:
+
+```
+govuk-docker compose up -d mysql
+pv whitehall_production.dump.gz | gunzip | govuk-docker compose run mysql mysql -u root --password=root whitehall_development
```
kevindew: I imagine you'd need a host of mysql specified here too.

README.md
brucebolt: I've added a similar step for MySQL.  Mongo also has the `--drop` flag, which replicates the behaviour of dropping then recreating the database.

Maybe when this is refactored to be slightly more automated, we can look into a way of doing this with Rails?

PR #171, opened 2019-08-13
Overwrite dnsmasq conf instead of abort

### LINKED ISSUE #153: `govuk-docker doctor` dnsmasq check fails on comments in /etc/resolver/...

ChrisBAshton:
Previously, if dnsmasq conf already exists then the setup command
would return without writing to it, which meant that the DNS may
not be configured properly.

After some discussion (https://github.com/alphagov/govuk-docker/pull/168\#issuecomment-519891271)
we've concluded that there is no foreseeable scenario whereby a
developer will need to ever configure this file manually, so we
should be able to overwrite it safely.

'govuk-docker doctor' now fails if /etc/resolver/dev.gov.uk is in
any way different to dnsmasq.conf in govuk-docker.

Arises from #168.

Fixes #153.


lib/govuk_docker/doctor/doctor.rb
```diff
@@ -67,10 +67,8 @@ def self.dnsmasq_messages
       HEREDOC
       dnsmasq_resolver: "✅ Dnsmasq is resolving DNS requests",
       not_dnsmasq_resolver: <<~HEREDOC,
-        ❌ Something else (possibly Vagrant-dns) is configured to resolve DNS requests for dev.gov.uk.
-          The /etc/resolver/dev.gov.uk file needs to be updated to only contain
-          nameserver 127.0.0.1
-          port 53
+        ❌ Your DNS resolver file (/etc/resolver/dev.gov.uk) is out of date with govuk-docker.
+        It should be overwritten with the contents of ~/govuk/govuk-docker/config/dnsmasq.conf
```
thomasleese: We should use `govuk_docker_dir` here to generate this path.

lib/govuk_docker/doctor/checkup.rb
```diff
@@ -87,7 +87,7 @@ def running_user_message
     end
 
     def dnsmasq_resolver?
-      File.read('/etc/resolver/dev.gov.uk').strip == "nameserver 127.0.0.1\nport 53"
+      File.read('/etc/resolver/dev.gov.uk') == File.read(File.join(GovukDocker::Paths::govuk_docker_dir, 'config', 'dnsmasq.conf'))
```
thomasleese: It might be worth extracting `File.join(GovukDocker::Paths::govuk_docker_dir, 'config', 'dnsmasq.conf'` into a variable or method here as this line is quite long.

lib/govuk_docker/setup/dnsmasq.rb
```diff
@@ -44,7 +44,8 @@ def install_dnsmasq
   def configure_etc_resolver_devgovuk
     write_file(
       "/etc/resolver/dev.gov.uk",
-      "nameserver 127.0.0.1\nport 53"
+      File.read(File.join(GovukDocker::Paths::govuk_docker_dir, 'config', 'dnsmasq.conf')),
```
thomasleese: Same thing about extracting this out into a variable or method.

lib/govuk_docker/doctor/checkup.rb
ChrisBAshton: Have created a `relative_to_govuk_docker` method in `GovukDocker::Paths`.

ChrisBAshton:
CC'ing @benthorner as we discussed this work before you went away. Is this what you had in mind?

lib/govuk_docker/setup/dnsmasq.rb
```diff
@@ -44,7 +44,8 @@ def install_dnsmasq
   def configure_etc_resolver_devgovuk
     write_file(
       "/etc/resolver/dev.gov.uk",
-      "nameserver 127.0.0.1\nport 53"
+      File.read(GovukDocker::Paths::dnsmasq_conf),
```
benthorner: Much better! I didn't realise you could use `::` to refer to a class method! I think it's more conventional to use `.`, though.

spec/paths_spec.rb
```diff
@@ -42,4 +42,25 @@
       end
     end
   end
+
+  describe "dnsmasq_conf" do
```
benthorner: Minor, but we could simplify this test by stubbing the `govuk_docker_dir` it delegates to.

lib/govuk_docker/setup/dnsmasq.rb
ChrisBAshton: You can only do it if you specify `self` before the method's definition.

Thought the other methods were being called with `::` but have double-checked and they're not - will update to `.` 👍

Issue #173, opened 2019-08-21
Unable to specify stacks that don't have `app-` prefix


ChrisBAshton:
E.g. govuk-content-schemas:

```bash
$govuk-docker startup
Unknown stack: app.

Available stacks:

lite


$govuk-docker startup lite

Unknown stack: app-lite.

Available stacks:

lite
```

Have tried passing `--stack=lite` which used to work but has now been removed from govuk-docker.

thomasleese:
This is intentional, albeit slightly confusing. If the service doesn't have an `app-` prefixed stack, then it can't be run using the `startup` command. Instead you would use the normal `run` command.

This is to make a distinction between apps where "starting up" is meaningful (such as a web application) and apps where this makes less sense (such as the content schemas).

The error message could definitely be improved though...

ChrisBAshton:
Cool, thanks for explaining 👍 perhaps we can check if an app has no `app-` prefixed stacks, and throw a more helpful error message if the dev tries running `startup`.

PR #175, opened 2019-08-21
Audit services list in README

ChrisBAshton:
- Reorders alphabetically
- Adds missing services and their statuses
- Adds note about the special nginx-proxy service

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,14 @@ The following apps are supported by govuk-docker to some extent.
    - ✅ govuk_app_config
    - ✅ govuk_publishing_components
    - ✅ govuk-cdn-config
+   - ❓ govuk-content-schemas
+      * Service exists in govuk-docker but is untested
    - ✅ govuk-developer-docs
    - ✅ govuk-lint
    - ✅ info-frontend
+   - ⚠ link-checker-api
+      * Works in isolation but not in other services' `e2e` stacks, so must be run in a separate process.
+        See https://github.com/alphagov/govuk-docker/issues/174 for details.
    - ✅ manuals-frontend
    - ✅ miller-columns-element
    - ✅ plek
--- a/README.md
+++ b/README.md
@@ -93,29 +93,29 @@ The following apps are supported by govuk-docker to some extent.
    - ⚠ cache-clearing-service
       * Tests pass
       * Queues are not set-up, so cache-clearing-service can't be run locally
-   - ✅ calendars
    - ⚠ calculators
       * Web UI doesn't work without the content item being present in the content-store.
+   - ✅ calendars
+   - ⚠  collections
+      * You will need to [populate the Content Store database](#mongodb) or run the live stack in order for it to work locally.
+      * To view topic pages locally you still need to use the live stack as they rely on Elasticsearch data which we are yet to be able to import.
    - ✅ collections-publisher
    - ⚠ content-data-admin
       * **TODO: Missing support for a webserver stack**
    - ✅ content-publisher
    - ✅ content-store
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,44 +93,48 @@ The following apps are supported by govuk-docker to some extent.
...
    - ✅ email-alert-frontend
    - ✅ finder-frontend
    - ❌ frontend
    - ✅ government-frontend
    - ✅ govspeak
+   - ✅ govuk_app_config
+   - ✅ govuk_publishing_components
    - ✅ govuk-cdn-config
+   - ⚠ govuk-content-schemas
+      * Pending https://github.com/alphagov/govuk-docker/issues/173
```
benthorner: I'm not sure this one is really an issue with this particular service.

README.md
```diff
@@ -141,6 +145,10 @@ The following apps are supported by govuk-docker to some extent.
       * Rake task to [create a test taxon](https://github.com/alphagov/whitehall/blob/master/lib/tasks/taxonomy.rake#L11) for publishing is not idempotent
       * Placeholder images don't work as missing proxy for [/government/assets](https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/news_article_presenter.rb#L133)
 
+NB, there is an nginx-proxy service under `services/`, which is the only service
```
benthorner: I don't think we need this, as it's an implementation detail - we could also have other services for postgres, etc., if we needed to override config or something. Can we do without this line?

ChrisBAshton: I can remove this line 👍 better yet, would it make sense to move the `nginx-proxy` service to a separate directory to keep application config and the 'inner workings of govuk-docker' a little more separate from each other?

README.md
```diff
@@ -93,44 +93,48 @@ The following apps are supported by govuk-docker to some extent.
...
    - ✅ govuk-cdn-config
+   - ⚠ govuk-content-schemas
+      * Pending https://github.com/alphagov/govuk-docker/issues/173
    - ✅ govuk-developer-docs
    - ✅ govuk-lint
-   - ✅ govuk_app_config
-   - ✅ govuk_publishing_components
    - ✅ info-frontend
+   - ⚠ link-checker-api
+      * Pending fix https://github.com/alphagov/link-checker-api/pull/284
```
ChrisBAshton: Now merged. Update ^

README.md
benthorner: Hmm...I guess there's a clash of concepts here. The 'services/' directory refers to docker-compose services (vs. a govuk-docker 'service' you can run/startup). I was going to say that's mostly an implementation detail, but then I see you can do this:

```
finder-frontend$guidance-links gdr --service nginx-proxy
Unknown stack: lite.

Available stacks:

app
```

While that's not great, in practice I doubt people will be browsing the services directory for things to run. I'd suggest waiting to see if others find this confusing before we introduce another place for config. How does that sound?

README.md
ChrisBAshton: Fair point 👍 can't quite remember why I linked this issue with this particular service.

ChrisBAshton: Not quite sure what to put in its absence. Reluctant to put a `✅` without _knowing_ it works, and because of issue #173 I'm unable to verify that.

Thinking of putting this, unless you can put my mind at rest?

```
   - ❓ govuk-content-schemas
      * Service exists in govuk-docker but is untested
```

README.md
ChrisBAshton: Have removed 👍

README.md
benthorner: Works for me :thumbsup:

PR #177, opened 2019-08-30
Document how to work with local gems

ChrisBAshton:
Couldn't find this documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,14 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
 app.dev.gov.uk.		0	IN	A	127.0.0.1
 ```
 
+### How to: work with local gems
+
+Provide a local gem path relative to the location of the Gemfile you're editing:
+
+```ruby
+gem 'govuk_publishing_components', path: '../govuk_publishing_components' 
+```
+
 ### How to: replicate data locally
 
 There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples reqire pv, which can be installed on a Mac using Brew (`brew install pv`).


README.md
```diff
@@ -262,6 +262,14 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
 app.dev.gov.uk.		0	IN	A	127.0.0.1
 ```
 
+### How to: work with local gems
+
+Most govuk-docker services have been configured to mount `~/govuk` at `/govuk`, so you can reference local applications like so:
+
+```ruby
+gem 'govuk_publishing_components', path: '/govuk/govuk_publishing_components'
```
andysellick: The alternative I use is `path: '../govuk_publishing_components'`

leenagupte: My `govuk` folder is in my home directory, so I use either:

`gem 'govuk_publishing_components', path: '~/govuk/govuk_publishing_components'`
Or
gem 'govuk_publishing_components', path: '../govuk_publishing_components'

ChrisBAshton: Can't use `~/govuk/govuk_publishing_components`, as `/govuk` is not the home directory in Docker. End up getting this error when trying to `govuk-docker build`:

```
The path `/home/build/govuk_publishing_components` does not exist.
```

`../govuk_publishing_components` works though, and is probably better because that works for both Docker and VM environments. 👍

ChrisBAshton: Thanks for approving @benthorner - I've just pushed up a tweak in response to this thread, could you re-read and re-approve please?

README.md
```diff
@@ -262,6 +262,16 @@ To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.
...
 
+### How to: work with local gems
+
+Provide a local gem path relative to the location of the Gemfile you're editing:
+
+```ruby
+gem 'govuk_publishing_components', path: '../govuk_publishing_components' 
+```
+
+You can reference any applications you have under `~/govuk`, which is mounted as a volume to your container at `/govuk`.
```
benthorner: I'm not understanding this - do we need to say it?

ChrisBAshton: Probably not.

PR #178, opened 2019-09-09
Run docker commands directly, without docker-compose

benthorner:
Currently we have three CLIs to understand as part of using govuk-docker: `docker`, `docker-compose` and `govuk-docker` itself. This was OK in the beginning, when `govuk-docker` was a mere 70 line bash script, but since then we've moved to a large Ruby codebase that's more open to development, and several fundamental issues with docker-compose have come to light:

   - The `run` command [isn't robust when dependencies aren't ready](https://docs.docker.com/compose/startup-order/), which means some commands - e.g. `rake` tasks - sporadically fail the first time they are run
   - The `run` (or `startup`) command doesn't monitor the logs or status of dependencies, which means silent failures or API errors are not visible in the console*

*the startup command does have a `abort-on-container-exit` option, but I've not yet seen it work; it also provides combined logging output, but only for the top-level container(?!)

I'd like to raise this as an issue, but also suggest a potential way forward: dropping `docker-compose`. The aim of this PR is to illustrate the relatively small amount of code it takes to drop `docker-compose` and have a foundation upon which to resolve the above issues. There are some immediate pros and cons to making this change:

   - Pro: we are free to fix the issues we're experiencing
   - Pro: we have one less CLI to understand
   - Con: we're reinventing part of a wheel
   - Con: we don't benefit from updates to `docker-compose`

 I'd like to gather thoughts on this approach and alternatives, ideally culminating in a decision about what to do next. The challenge is to find an approach that reduces the learning curve, but also provides a means to fix the issues.


kevindew:
Thanks for putting this together Ben, I'm impressed with how much docker wrangling you managed to achieve here 👍. It'd be useful to get some context on how this change would help resolve the problems identified for context.

Some other cons I can think of are:

- we'd have lots of docker-compose.yml (or different named files) that are just in a free form syntax of our own that resembles docker-compose but is not necessarily the same
- have to maintain and document a port of the docker-compose bits we do use
- having to replicate any new docker-compose features we want to use
- lose commands I use like: logs, down, ps, run without dependencies, and options like service-ports

Personally having access to the compose command on govuk-docker has been one of my favourite features as I've been able to re-use all of my existing knowledge of compose with govuk-docker without having to add things to the govuk-docker CLI.

barrucadu:
I've had a look for all instances of `govuk-docker compose` in my shell scripts:

- `govuk-docker compose stop` - self-explanatory
- `govuk-docker compose ps --services | grep "$app"` - list stacks for an app
- `govuk-docker compose run -d --rm -v $archive_path:/replication -v $cfg_path:/usr/share/elasticsearch/config/elasticsearch.yml -p 9200:9200 "elasticsearch${ver}"` - elasticsearch data sync

So from my perspective we'd need all those to work (or something equivalent).

---

A few cons which come to mind are:

- **On-boarding new people:** if someone is familiar with docker from a previous role, they're almost certainly familiar with `docker-compose`.  It's needlessly confusing if we reimplement a `docker-compose` feature but with a different name or arguments.

- **Complexity:** I suspect there are a lot of different ways in which people are using `govuk-docker compose`, and adding support for all of them (and any future uses) increases the amount of code here, and so the potential for bugs.

---

On the other hand, I quite like Kevin's approach with dockerize.  The most significant pain point in `govuk-docker`, to me, is that containers don't wait for their dependencies.  I think solving that would be a huge benefit and, if we could do it without reimplementing a significant chunk of `docker-compose`, even better.

benthorner:
@kevindew thanks for the suggestions. I'm conscious this PR is oriented around the architectural change, rather than solving the compose-specific issues, with the understanding that "anything is possible in code". On it's own, it's a way of reducing the complexity, so could be a change in its own right for that single reason. I do believe the complexity is a problem, so I'm interested in suggestions about how we could reduce it, as well as fix the compose issues.

I agree with the downsides of dropping a popular framework, and ideally we should avoid this. And I also make use of `stop` and ` logs` with compose, but not `ps` (just plan `docker ps`), so I'm interested to know about the use case for that one. Specifically for the `logs` command, is the use case to replicate the multi-app logs we used to get with `foreman`? If so, it seems relevant to the second compose issue I mentioned. I'm wondering if we might remove the need for this if compose could include logs for dependencies when running an app.

I've had a look at `wait-for-it` and `wait-for` in the past, but not `dockerize`, which is annoying because this seems to be the only recommended solution that: supports multiple wait conditions; also works with `entrypoint`. I'm keen not to diddle with the commands as part of doing this, and I was pleased to find I could achieve the following.

```
docker-compose -f [...] run --rm --service-ports content-publisher-app
...
Starting govuk-docker_asset-manager-app_1  ... done
Starting govuk-docker_router-api-app_1     ... done
Starting govuk-docker_content-store-app_1  ... done
2019/09/11 16:14:29 Waiting for: http://publishing-api.dev.gov.uk/healthcheck
2019/09/11 16:14:29 Received 502 from http://publishing-api.dev.gov.uk/healthcheck. Sleeping 1s
2019/09/11 16:14:36 Received 200 from http://publishing-api.dev.gov.uk/healthcheck
=> Booting Puma
```

While I have a concern that we're adding yet more stuff, to solve problems with other stuff, the thing I like about this is that we only need to use it where flakiness is actually a problem, rather than some generic healthcheck mechanism, which would be overkill for the overheads it would introduce.

## Self Summary

   - Can we think of other approaches to reduce the total complexity of `govuk-docker`?
   - `logs`, `ps` and `stop` are popular, but could we do better?
   - `dockerize` looks good - should we use it to fix https://trello.com/c/YbvBZ4iq/68-build-steps-can-fail-when-services-arent-ready?

kevindew:
> @kevindew thanks for the suggestions. I'm conscious this PR is oriented around the architectural change, rather than solving the compose-specific issues, with the understanding that "anything is possible in code". On it's own, it's a way of reducing the complexity, so could be a change in its own right for that single reason. I do believe the complexity is a problem, so I'm interested in suggestions about how we could reduce it, as well as fix the compose issues.

No worries - I think it's hard to consider this outside of the issues it is intended to solve, but I am indeed very impressed how you managed to replicate so much of docker compose so easily 👍 

Re complexity: Is that a reference to the amount of ruby code we have for the CLI? One area where I see an opportunity for complexity reduction is the makefiles, which - were they done with Ruby - would be one less universe to understand. Though I don't if it that's realistic or actually offer benefits. And I may be looking at the complexity from a different perspective to you.

> I agree with the downsides of dropping a popular framework, and ideally we should avoid this. And I also make use of `stop` and ` logs` with compose, but not `ps` (just plan `docker ps`), so I'm interested to know about the use case for that one. Specifically for the `logs` command, is the use case to replicate the multi-app logs we used to get with `foreman`? If so, it seems relevant to the second compose issue I mentioned. I'm wondering if we might remove the need for this if compose could include logs for dependencies when running an app.

I mostly use compose ps is mostly preference born out of habbit. I like that it by default shows which container exited with the code and it just find it a bit more readable. It's also handy that it's scoped just to the govuk-docker containers.

Mostly I use the logs command to check a single or couple of containers. I'll go for the whole stack on occasion but given there's so much output I don't tend to find it very useful (which was a similar problem I had with the vm bowl command). When I've been keen on monitoring a couple of apps output I've tended to run them both, for example `govuk-docker compose up email-alert-frontend-app finder-frontend-app` for recent work on comms between these apps.

> I've had a look at `wait-for-it` and `wait-for` in the past, but not `dockerize`, which is annoying because this seems to be the only recommended solution that: supports multiple wait conditions; also works with `entrypoint`. I'm keen not to diddle with the commands as part of doing this, and I was pleased to find I could achieve the following.

Ah ha - that's just luck. I saw the three options on the docker docs and picked dockerize because of it being wrote by jwilder. I hadn't realised those tools had that limitation, though looks like there are forks that add it in. I also found a less popular option that allowed the waits to be configured by an env var which looked a concise way of expressing it https://github.com/ufoscout/docker-compose-wait though doesn't look suitable for an entrypoint.

> While I have a concern that we're adding yet more stuff, to solve problems with other stuff, the thing I like about this is that we only need to use it where flakiness is actually a problem, rather than some generic healthcheck mechanism, which would be overkill for the overheads it would introduce.

Sure, may be worth tinkering to see how many changes are needed to get a `govuk-docker setup` to run successfully.

benthorner:
@barrucadu thanks for giving your take on this. I agree that the dependency waiting issue seems to be the one people have voiced most often, and we could look at adding `dockerize` to fix those dependencies that are currently creaking.

I agree the definite con of reproducing parts of compose is somewhat unbounded, and we'd have to have strong motivation to go that way. And we'd lose the documentation and prior knowledge benefits that come with it, although prior knowledge seems scarce!

benthorner:
I'm going to close this PR (/bundle of issues). Thanks for everyone who commented with their thoughts - it's been food for my thoughts, and I've also learnt a thing or two out of it!

After some reflection, I'm inclined to agree that dropping docker-compose would not have a net benefit. I'd like to summarise the issues people have raised in the comments:

   - Dropping the compose CLI would mean our config files no longer follow a well-documented format, making them harder to understand for people unfamiliar with the repo
   - Replacing the compose CLI is an unbounded task, and arguably gives us excessive freedom to deviate or [bikeshed](https://en.wiktionary.org/wiki/bikeshedding) in what we choose to support
   - One of the 2 functional issues in the original description can be resolved using [an existing approach](https://github.com/jwilder/dockerize), which I wasn't aware of when writing this PR

That said, the three issues - waiting for dependencies, dependencies exiting, and overall complexity - still exist and we should still think about how best to resolve them. 

I think this PR has served its purpose: to determine if one approach would be practical. I'll continue to think about other ways to get the complexity down, as having 3 layers of CLI is a clear barrier for people learning the tool. We also now have a piece of work to look at functional issues, which should hopefully come to address the two listed here (if they really are a problem!).

https://trello.com/c/Cij7BWiD/102-quarterly-survey-prioritisation-and-fixing-of-issues

PR #180, opened 2019-09-13
Document how to remove Docker containers

ChrisBAshton:
Myself and a few members of the step by step team have had to resort to this a few times
when working with collections-publisher, which would sometimes get in a painful state trying
to interact with publishing-api or link-checker-api.  It did not appear to be fixable through the
normal methods of updating and rebuilding individual services, running in an `e2e` stack or
restarting Docker itself.

The most reliable way to fix some of the more obscure problems has been to clear all
containers and rebuild them. The commands are not very intuitive or easy to remember, so they
are worth documenting in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,17 @@ Sometimes it's useful to get all changes for all repos e.g. to support finding t
 make pull
 ```
 
+### How to: clear your Docker containers
+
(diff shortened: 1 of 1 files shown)


ChrisBAshton:
I've pushed up a revised PR 👍 I'm quite happy to leave it as two commands for now, in the spirit of each command doing one thing well, and the fact we seldom have to run the commands. If we find we have to run these more regularly then it may be worth combining them in future.

PR #182, opened 2019-09-17
Add details on increasing Docker resource settings (to help with Whitehall import)

chao-xian:
Local data is required for rendering pages that Whitehall Frontend
serves directly. But as the app is so intense, Docker needs to be
beefed up to cope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,12 +267,12 @@ app.dev.gov.uk.		0	IN	A	127.0.0.1
 Provide a local gem path relative to the location of the Gemfile you're editing:
 
 ```ruby
-gem 'govuk_publishing_components', path: '../govuk_publishing_components' 
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
 ```
 
 ### How to: replicate data locally
 
-There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples reqire pv, which can be installed on a Mac using Brew (`brew install pv`).
+There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples require pv, which can be installed on a Mac using Brew (`brew install pv`).
 
 #### MySQL
 
@@ -291,6 +291,16 @@ govuk-docker compose run mysql mysql -h mysql -u root --password=root -e "CREATE
 ```
 pv whitehall_production.dump.gz | gunzip | govuk-docker compose run mysql mysql -h mysql -u root --password=root whitehall_development
 ```
+
+*For Whitehall*
+
+In order to ensure that Whitehall runs smoothly, your Docker settings may need to be changed in order to handle Whitehall. By default, Docker defaults to using 2 CPU cores and 2GB of RAM, and this is simply not enough to render Whitehall in a Docker container.
+
+You should set it to use at least 4 CPU cores and 8GB of RAM - 6 CPUs and 8 GB of RAM seems to work ok. Open the Docker dropdown via the whale icon in the OSX menu bar and selecting the `Advanced` option.
+
+You should also increase the Docker disk image size to the maximum.
+
+We have found no tangible improvement in increasing swap space size compared to CPU, RAM and disk allocation.
 
 #### PostgreSQL
 
--- a/README.md
+++ b/README.md
@@ -58,6 +58,19 @@ govuk-docker has the following dependencies:
 
 All other dependencies will be installed for you automatically.
 
+#### Docker settings
+These need to be changed in order to handle intensive apps like Whitehall. Docker defaults to using 2 CPU cores and 2GB of RAM. Open the Docker dropdown via the Docker whale icon in the macOS menu bar, and select the `Advanced` option.
+
+Change to ideally:
+
+* 6 CPUs
+* 12 GB RAM
+* Docker disk image size default is 64GB under the Disk tab
+  * depending on how much free space you have
+  * the minimum should be > 50GB
+
+We have found no tangible improvement in increasing swap space size compared to CPU, RAM and disk allocation.
+
 ### Setup
 
 Start with the following in your bash config.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -292,6 +292,16 @@ govuk-docker compose run mysql mysql -h mysql -u root --password=root -e "CREATE
 pv whitehall_production.dump.gz | gunzip | govuk-docker compose run mysql mysql -h mysql -u root --password=root whitehall_development
 ```
 
+*For Whitehall*
+
+In order to ensure that Whitehall runs smoothly, your Docker settings may need to be changed in order to handle Whitehall. By default, Docker defaults to using 2 CPU cores and 2GB of RAM, and this is simply not enough to render Whitehall in a Docker container.
+
+You should set it to use at least 4 CPU cores and 8GB of RAM - 6 CPUs and 8 GB of RAM seems to work ok. Open the Docker dropdown via the whale icon in the OSX menu bar and selecting the `Advanced` option.
```
kevindew: I'd suggest that these settings apply beyond just Whitehall and are pretty much a pre-requisite to running govuk-docker. 

Maybe rather than giving 2 numbers (min and ideal) we just suggest an ideal and it's at a user's risk if they go lower. I'd actually suggest 6 cores and 12GB of ram rather than 8GB.

README.md
```diff
@@ -292,6 +292,16 @@ govuk-docker compose run mysql mysql -h mysql -u root --password=root -e "CREATE
...
+
+You should also increase the Docker disk image size to the maximum.
```
kevindew: Do you know what level people are using to hit the max? I've not changed mine from the default of 64 GB and have been fine to import Whitehall (my MySQL usage is 22GB.

As the maximum might well be relative it might be best to specify a number - 100GB is probably plenty for most usages, whereas if you plan to clone the whole of GOV.UK you may need more still.

README.md
```diff
@@ -267,12 +267,12 @@ app.dev.gov.uk.		0	IN	A	127.0.0.1
...
 
 ```ruby
-gem 'govuk_publishing_components', path: '../govuk_publishing_components' 
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
 ```
 
 ### How to: replicate data locally
 
-There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples reqire pv, which can be installed on a Mac using Brew (`brew install pv`).
+There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples require pv, which can be installed on a Mac using Brew (`brew install pv`).
```
issyl0: Rather than adding [more individual `brew install` commands](https://github.com/alphagov/govuk-docker/search?q=brew+install&unscoped_q=brew+install), how about switching to use a [Brewfile](https://github.com/Homebrew/homebrew-bundle#usage)? Then, dependencies can be installed in one go with `brew bundle` at the start of the `govuk-docker setup` script.

chao-xian: Is somebody a contributor to Homebrew? @issyl0 🤣

README.md
issyl0:
```suggestion
You should set it to use at least 4 CPU cores and 8GB of RAM - 6 CPUs and 8 GB of RAM seems to work ok. Open the Docker dropdown via the whale icon in the macOS menu bar and selecting the `Advanced` option.
```

README.md
issyl0: I was trying very, very hard to not put that hat on while still being useful. :joy:

issyl0: Also sorry, the diff at a glance made it look like you added the entire line when you didn't. I'll tackle all this stuff separately I guess. :-)

chao-xian: @issyl0 I like your suggestion, but can I do it in a separate PR with the change to the setup script?

chao-xian:
@benthorner yup, I agree it's getting long but not at this point. I'll raise a separate PR for that.

README.md
chao-xian: Done in https://github.com/alphagov/govuk-docker/pull/182/commits/c9a83d33effc7f0402cdeb636eb9f2a512f402b9

README.md
chao-xian: I really don't have 100 GB and it seemed I had less than the default. Amended for that case in https://github.com/alphagov/govuk-docker/pull/182/commits/c9a83d33effc7f0402cdeb636eb9f2a512f402b9

chao-xian: I really don't have 100 GB and it seemed I had less than the default. Amended for that case in https://github.com/alphagov/govuk-docker/pull/182/commits/c9a83d33effc7f0402cdeb636eb9f2a512f402b9

README.md
```diff
@@ -58,6 +58,19 @@ govuk-docker has the following dependencies:
...
+#### Docker settings
+These need to be changed in order to handle intensive apps like Whitehall. Docker defaults to using 2 CPU cores and 2GB of RAM. Open the Docker dropdown via the Docker whale icon in the macOS menu bar, and select the `Advanced` option.
+
+Change to ideally:
+
+* 6 CPUs
+* 12 GB RAM
+* Docker disk image size default is 64GB under the Disk tab
+  * depending on how much free space you have
+  * the minimum should be > 50GB
```
kevindew: I don't think the disk setting should be in this section since that is managed in a different option than Advanced. 

Probably worth having a section for `Advanced` and one for `Disk`. I'm also not too sure what you're saying in the Disk section. 

I'd suggest tweaking this around to be something like:

```
#### Docker Settings

Running GOV.UK applications can be resource intensive and will easily exceed the default configuration of Docker for Mac. To change settings open the Docker dropdown via the Docker whale icon in the macOS menu bar and select the preferences option. 

In `Advanced` settings you should update CPU and RAM resources. These should be at least:

* 6 CPUs
* 12 GB RAM

In `Disk` you should ensure there is a high amount of disk space to allow replicating GOV.UK data. 64GB should be sufficient for most usages but you may need > 100GB to clone all GOV.UK integration data.
```

chao-xian: @kevindew Thanks Kevin! Updated in https://github.com/alphagov/govuk-docker/pull/182/commits/7d9e51df8b105471e501cae31e49a7521580377f 

And thanks for the rewrite. Not quite as as easy as hitting accept on a "Suggested commit" though 😉😂

PR #183, opened 2019-09-18
Use a Brewfile instead of loads of individual `brew install` commands

issyl0:
- Rather than having a load of `brew install` lines and tests in separate files, this adds a new class - `HomebrewDependencies` - that runs `brew bundle` (using the [Brewfile](https://github.com/Homebrew/homebrew-bundle#Usage)) as the second step of `govuk-docker setup`.
- This means that Homebrew will install or update all the pre-requisite software packages at once, and means that we can delete a lot of code.

TODO:
- [x] Check that everything still works for people who actually run `govuk-docker` day-to-day. 🙏
- [x] Maybe squash the commits?
- [x] If it works, or when it's fixed, make this PR not draft.


benthorner:
Thanks for raising this: I like this is making use of a tool that's intended for this sort of thing 🏅. One thing I think we should consider here is removing the yes/no choice: my hope is that when people run a command, they accept it might change something 😬. Since this whole Ruby CLI is a bespoke thing, we should be careful to only those features in it that we know to be necessary. Do you think it would cause a problem if we just went ahead with the install?

issyl0:
Nope, it shouldn't cause problems. It won't install things that are already installed. :-) I've removed the asking for permission from the `HomebrewDependencies` step.

lib/govuk_docker/setup/homebrew_dependencies.rb
```diff
@@ -0,0 +1,8 @@
+require_relative "./base"
+require_relative "../doctor/checkup"
+
+class GovukDocker::Setup::HomebrewDependencies < GovukDocker::Setup::Base
+  def call
+    system("brew bundle --file=Brewfile")
```
issyl0: This doesn't need to specify `--file=Brewfile` as Brewfile is the default file path.

This was leftover from when I tried to be too clever and use `GovukDocker::Paths.govuk_docker_dir` in case people ran `govuk-docker setup` from a location other than the root of the repo. But I think I was over-engineering - the README instructions say to run `govuk-docker setup` from the root - so I backed those changes out.

Should I remove this or is it still useful somehow?

benthorner: This seems like a difficult call, since it the rest of the command does seem to work in other directories. I reckon this level of graceful behaviour isn't necessary, but I don't think we can just lose the functionality in one place and not in others.

issyl0: I don't understand what you want me to do, sorry. 😟Shall we keep it as it is?

benthorner: Sorry for the confusion: I was trying to say that we should still make sure this works when people run it in directories outside of govuk-docker, since that appears to be supported by the other parts of the setup command. This seems unnecessary and we can probably lose it, but it would be confusing to do so in this PR. Does that make sense?

issyl0: Having re-read this three times, yep, this makes sense.

We should make it work from anywhere in this PR. I'll (try to) do that (but may need some help with the tests if my intial experiments were anything to go by). And in a separate PR we can potenetially remove the logic that does it?

issyl0: I pushed a commit to do the path thing, and squashed the other tiny changes: now running `govuk-docker setup` from anywhere works fine. 🎉 I'll make this into a Real PR now.

issyl0:
I appeased the linter (double quotes vs single quotes!) and I'm now making this a Real PR. I'd appreciate your reviews. In particular, I'm not sure about the value of how I've tested the "can run from anywhere" test for the location of the `Brewfile`.

issyl0:
Thanks for the reviews!

PR #184, opened 2019-09-18
Add short-url-manager to services

kevindew:
~Depends on: https://github.com/alphagov/short-url-manager/pull/390~

This service hadn't made the list last time. Short URL Manager uses an
unconventional approach of creating a user in an initializer so doesn't
require the regular seed or User.create approach.

### DIFF
--- a/services/short-url-manager/docker-compose.yml
+++ b/services/short-url-manager/docker-compose.yml
@@ -1,0 +1,39 @@
+version: '3.7'
+
+x-short-url-manager: &short-url-manager
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: short-url-manager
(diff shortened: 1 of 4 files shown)


services/short-url-manager/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: '3.7'
+
+x-short-url-manager: &short-url-manager
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: signon
```
benthorner: I think the image needs changing here.

kevindew: D'oh! thanks for catching. My C+P source revealed to the world 😊

PR #186, opened 2019-09-23
Sync elasticsearch data [WIP]

koetsier:
Add a 'sync' command to govuk-docker to populate services with data

This will download the current Elasticsearch snapshot from S3 and
import it into the local docker setup

Currently only Elasticsearch is supported.

Example:
govuk-docker sync --elasticsearch_service=elasticsearch5 --keep_download

Full help: govuk-docker help sync

Trello link: https://trello.com/c/tVQV1SMi/1025-enable-govuk-docker-to-import-elasticsearch-data-from-s3-bucket

TODO: 
- docs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,12 @@
 source "https://rubygems.org"
 
+gem "aws-sdk-s3", "~> 1"
 gem "climate_control"
 gem "colorize"
+gem "elasticsearch", "~> 6"
 gem "govuk-lint"
+gem "highline"
 gem "pry"
 gem "rspec"
+gem "ruby-progressbar"
 gem "thor"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,18 +2,48 @@ GEM
   remote: https://rubygems.org/
   specs:
     ast (2.4.0)
+    aws-eventstream (1.0.3)
+    aws-partitions (1.214.0)
+    aws-sdk-core (3.68.0)
+      aws-eventstream (~> 1.0, >= 1.0.2)
+      aws-partitions (~> 1.0)
+      aws-sigv4 (~> 1.1)
+      jmespath (~> 1.0)
+    aws-sdk-kms (1.24.0)
+      aws-sdk-core (~> 3, >= 3.61.1)
+      aws-sigv4 (~> 1.1)
+    aws-sdk-s3 (1.48.0)
+      aws-sdk-core (~> 3, >= 3.61.1)
+      aws-sdk-kms (~> 1)
+      aws-sigv4 (~> 1.1)
+    aws-sigv4 (1.1.0)
+      aws-eventstream (~> 1.0, >= 1.0.2)
     climate_control (0.2.0)
     coderay (1.1.2)
     colorize (0.8.1)
     diff-lcs (1.3)
+    elasticsearch (6.8.0)
+      elasticsearch-api (= 6.8.0)
+      elasticsearch-transport (= 6.8.0)
+    elasticsearch-api (6.8.0)
+      multi_json
+    elasticsearch-transport (6.8.0)
+      faraday
+      multi_json
+    faraday (0.15.4)
+      multipart-post (>= 1.2, < 3)
     ffi (1.11.1)
     govuk-lint (4.0.0)
       rubocop (~> 0.72)
       rubocop-rails (~> 2)
       rubocop-rspec (~> 1.28)
       scss_lint
+    highline (2.0.2)
     jaro_winkler (1.5.3)
+    jmespath (1.4.0)
     method_source (0.9.2)
+    multi_json (1.13.1)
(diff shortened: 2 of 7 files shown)


benthorner:
@koetsier thanks for raising this and getting some progress here - I know several people have been looking at replication and solving this rather thorny problem.

We're trying to bring some coordination to the work on govuk-docker, so that it grows to meet the needs of the wider GOV.UK developer team. My concern here is that this is quite a large body of code to be adding, compared to similar code in the README. I appreciate it's a WIP and it might change a lot anyway! It would be good to understand the problems we're solving here.

Reading through the code, I'm also wondering how much of this is motivated by the needs of a particular team, and how much can be reused by all the developers on GOV.UK. Data replication is definitely something we want to get right, and I think this fits into the epics I've discussed with some of the senior devs. In particular, @barrucadu is [looking at data replication](https://trello.com/c/kIj15vK0/105-figure-out-what-people-need-from-data-replication-and-do-it) as a whole. 

I know some of the above will be surprising and possibly frustrating - we're going to talk more about this in Tech Fortnightly. Ultimately we all want to be maintaining a dev environment where every piece has significant value, and we're going to be thinking harder about the changes we're making, to see how well they fit that goal. When it comes to reviewing this PR, it would help to:

   - Write up a card with @barrucadu to cover this work on the [dev board](https://trello.com/b/kmruvfMm/govuk-development-environment-%F0%9F%90%B3)
   - Update the PR description to discuss the problems we're solving here

koetsier:
Thanks for looking at the code and your comments.

Syncing data seems to me to be something people will want at some point. It is of course motivated somewhat by needs from our team (search and nav) - we would like to have an fuss-free way that can sync data so we can experiment with different Elasticsearch configurations.

It is more or less a port of the sync-aws-elasticsearch script in govuk-puppet. Reusability-wise, the S3 stuff is fairly generic - the Elasticsearch client is a bit more specific at the moment but I'm not sure what else we would want to do with Elasticsearch in the govuk_docker repo.

Anyway, this code here works. So anyone who wants to have Elasticsearch data in their local Docker installation can use this branch. If there's questions about the usability of this code then I'll refrain from working on this further and I'll park this here.

koetsier:
Couldn't leave it alone and removed the config-reading stuff and some more bits that were not needed so it is a bit shorter not.

If LOC is still a problem, then the code can be pruned further. Mostly it is the Ruby AWS api that requires more boilerplate - and Ruby is a bit wordier than the equivalent Shell script (maybe also more readable i.e. see l

kevindew:
@koetsier since you're working further on this there's a couple of things here that I think should be looked at before considering merging.

First the bin, we advise users to put the [bin directory of this project into their $PATH](https://github.com/alphagov/govuk-docker#setup) this would have the effect that the `sync-aws-elasticsearch` bin would be available system wide which wouldn't be ideal given it's specific for this env (there is a trello card to remove everything else that is in there: https://trello.com/c/2bxWjW1c/95-we-shouldnt-have-binstubs-if-we-expect-the-bin-directory-to-be-in-a-users-path). Thus I'd expect data steps to be commands within govuk-docker.

Second, the classes here seem to break out of the convention of everything being within the namespace of govuk_docker that was introduced.

koetsier:
Thanks for your comments @kevindew 

- I updated the code so that now it is part of govuk-docker; i.e. the command to use is now ```govuk-docker sync <options>```

- The relevant files are now also part of the GovukDocker namespace.

benthorner:
@bilbof thanks for looking at this - the PR was somewhat languishing here! While it's helpful to have something that works, I still have concerns, which I'll try to enumerate more clearly:

   - We have prior art for doing replication without the CLI ([see here](https://github.com/alphagov/govuk-docker#how-to-replicate-data-locally). Using the CLI for ES-specific replication would create a confusing inconsistency.
   - We also have prior art for doing ES-replication [in bash](https://github.com/alphagov/govuk-docker/pull/157#issuecomment-515932847). It may not be everyone's favourite language, but I believe it's the best tool for the job here.
   - This code isn't tested and would significantly reduce our coverage if we merged it in. I think we should have some tests unless there's an urgent need to merge now.
   - Choosing Ruby for the purpose of writing tests is somewhat dubious here, as the tests we can write have low value, since it's impractical to make them E2E.

I think we should be careful with arguments about language and complexity. We have two new abstractions/adapters to understand here, for example. I should say the code itself has a nice structure and I find it clear to read - I'm keen on [the procedural layout](https://github.com/alphagov/govuk-docker/pull/186/files#diff-7e59876b981257d6c5374f32e0fef9d3R29), for example. However, I still think it's worth addressing the above concerns.

What may help is for us to have a look at this together, and see if we can get @barrucadu's script working and maybe simplify it further. How does that sound?

barrucadu:
Closing this as we've had shell scripts for a while

Issue #174, opened 2019-08-21
Link-Checker-API not working inside Collections Publisher 


ChrisBAshton:
![image](https://user-images.githubusercontent.com/5111927/63435652-355f9e80-c41f-11e9-8983-cce6abd214ec.png)

Inside `collections-publisher`:

```
govuk-docker startup e2e
```

Then navigate to http://collections-publisher.dev.gov.uk/step-by-step-pages and create or view a step by step and then click "Check for broken links"; we get a 503 error.

~This happens even if you run `govuk-docker startup` inside link-checker-api in another tab.~ As of https://github.com/alphagov/link-checker-api/pull/284, you can now have link-checker-api running in another tab and this is available to collections-publisher, fixing the issue. But we shouldn't have to do this - the `e2e` stack should account for the link-checker-api dependency and resolve it automatically.

benthorner:
@ChrisBAshton I've tried this myself and it seems to work OK now, without any extra steps. I did have to pull the latest versions of the two repos in question.

![Screen Shot 2019-11-11 at 15 18 40](https://user-images.githubusercontent.com/9029009/68598408-a64ea980-0496-11ea-82b0-4186032e8a98.png)

ChrisBAshton:
Thanks Ben. Yes I think this issue crops up now and again and required an image rebuild of the repos in question. Think I updated the govuk-docker readme to that effect, but forgot to close this issue. Closing now :)

PR #187, opened 2019-09-23
Warn users about unsupported functionality in collections-publisher

ChrisBAshton:
Some parts of the collections-publisher app do not work seamlessly
(if at all). As the govuk-docker README is the centralised source of
truth on this, I think it's worth highlighting here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,12 @@ cd ~/govuk/content-publisher
 govuk-docker startup e2e
 ```
 
+### Interoperability of stacks
+
+Even if an e2e stack is started, functionality won't necessarily work as expected "end to end" as govuk-docker doesn't mimick the routing we have on GOV.UK. For example, publishing apps that link to draft-origin.dev.gov.uk frontends will see a server error, as draft-origin isn't a service in govuk-docker.
+
+In these cases, you can swap out the URL for the relevant frontend (such as draft-collections.dev.gov.uk). It's also worth noting that this frontend will need to be started separately, as publishing apps don't define frontend apps in their Docker service dependencies; a publishing app doesn't need its corresponding frontend in order to be able to publish.
+
 ## How to's
 
 ### How to: troubleshoot your installation


README.md
```diff
@@ -107,7 +107,9 @@ The following apps are supported by govuk-docker to some extent.
    - ⚠  collections
       * You will need to [populate the Content Store database](#mongodb) or run the live stack in order for it to work locally.
       * To view topic pages locally you still need to use the live stack as they rely on Elasticsearch data which we are yet to be able to import.
-   - ✅ collections-publisher
+   - ⚠ collections-publisher
+      * 'Checking for broken links' in step by step requires link-checker-api to be running in another process (see link-checker-api below).
```
benthorner: I think there's a potential for confusion here, since link-checker-api will be running as part of docker-compose dependency management. Can we clarify this to indicate the problem is with link-checker-api not working (vs. docker-compose not starting the process)?

README.md
```diff
@@ -107,7 +107,9 @@ The following apps are supported by govuk-docker to some extent.
...
+      * Previewing step by steps does not work as there is no draft stack.
```
benthorner: I wonder if we can identify the piece of work 'todo' here? When I read this, I was wondering if it referred to a missing dependency, or missing data, or something not working. Hopefully there's something simple we can do to get it working.

ChrisBAshton: I'm not sure how to fix it - would appreciate your experience. Clicking 'Preview' in a step by step in collections publisher takes me to `http://draft-origin.dev.gov.uk/my-step-by-step?token=[redacted]`, which shows a `503 Service Temporarily Unavailable` page.

I understand that step by steps are rendered by [collections](https://github.com/alphagov/collections), so tried running that in a separate terminal (`govuk-docker startup`) in case it has the same inter-connectivity issue as link-checker-api. When I refresh the preview URL I can see in the collections terminal window that a request has been made to `/my-step-by-step`, but I eventually get a  `Warning: Unable to communicate with application within 30 seconds.` and a `504 Gateway Time-out` page.

ChrisBAshton: Thanks for pairing IRL. Here's my attempt at summarising:

* `draft-origin.dev.gov.uk` isn't a thing in govuk-docker. Routing is separate to applications, so whilst it's not particularly dev friendly, it shouldn't be 'fixed' here.
* However, draft-origin.dev.gov.uk should be able to be swapped out for collections.dev.gov.uk, which is the rendering app.
* collections.dev.gov.uk also gave a 503, as it is not listed as a dependency of collections-publisher. Again, this isn't very dev friendly, but as collections and collections-publisher are decoupled from each other, it makes little sense to list collections as a dependency, and if we started to do things like that we'd essentially have to start up all frontend apps for all services.
* In a separate process, running `govuk-docker startup draft` on collections starts up collections.dev.gov.uk as expected. But the styling is a bit off. It transpires this is because the draft collections frontend relies on parts of static and parts of draft-static. This is a legitimate issue - have raised #189 to fix.

I'll go back and remove line 112 from this PR, and have a think about whether this information should be captured in a 'How to' or something.

benthorner: @ChrisBAshton that's all good, except all uses of 'collections.dev.gov.uk' should actually be 'draft-collections.dev.gov.uk' :-).

README.md
ChrisBAshton: Have rewritten - please check.

README.md
ChrisBAshton: Have removed this line and drafted a new "Interoperability of stacks" subsection in the "Stacks" section - please have a read.

README.md
```diff
@@ -184,6 +185,12 @@ cd ~/govuk/content-publisher
 govuk-docker startup e2e
 ```
 
+### Interoperability of stacks
+
+Even if an e2e stack is started, functionality won't necessarily work as expected "end to end" as govuk-docker doesn't mimick the routing we have on GOV.UK. For example, publishing apps that link to draft-origin.dev.gov.uk frontends will see a server error, as draft-origin isn't a service in govuk-docker.
+
+In these cases, you can swap out the URL for the relevant frontend (such as draft-collections.dev.gov.uk). It's also worth noting that this frontend will need to be started in a separate process, as publishing apps don't define frontend apps in their Docker service dependencies; a publishing app doesn't need its corresponding frontend in order to be able to publish.
```
benthorner: This looks good. just a note that 'process' might be a bit confusing, as we're really talking about a 'docker-compose service + dependencies'. Perhaps just say 'started separately'?

README.md
```diff
@@ -110,7 +110,8 @@ The following apps are supported by govuk-docker to some extent.
    - ⚠  collections
       * You will need to [populate the Content Store database](#mongodb) or run the live stack in order for it to work locally.
       * To view topic pages locally you still need to use the live stack as they rely on Elasticsearch data which we are yet to be able to import.
-   - ✅ collections-publisher
+   - ⚠ collections-publisher
+      * 'Checking for broken links' in step by step requires link-checker-api to be started manually in another process (despite collections-publisher starting it - see link-checker-api below).
```
benthorner: Same here as the comment below: 'started manually in anther process' -> 'started separately'.

ChrisBAshton:
> After noticing that [the associated issue](https://github.com/alphagov/govuk-docker/issues/174) has a merged PR to fix it(?), I tried it out myself, and everything seems to work fine without doing anything special...

It's definitely quite flaky 🤔 I just tried this:

```
govuk-docker startup e2e

# check links - fail

govuk-docker startup

# check links - works! 🎉

govuk-docker startup e2e # (again)

# check links - works... 😅
```

The PR intended to fix it, and appears to have largely fixed it, but I am still regularly running into interconnectivity issues that I haven't yet been able to reduce to a simple reproducible example.

benthorner:
@ChrisBAshton as discussed, it looks like this is a bug with the E2E stack specifically. I've just put up a fix for that: https://github.com/alphagov/govuk-docker/pull/201. So it looks like this PR is somewhat superseded - sorry it's taken so long! - and could be replaced with another two: one to remove the issue with link-checker-api; and one for the new E2E doc proposed here.

How does that sound?

ChrisBAshton:
Agreed - have raised #208. Thanks

PR #189, opened 2019-09-24
Add 'static-app' as a dependency to collections frontend draft stack

ChrisBAshton:
Collections frontend draft stack relies on both static-app-draft and
static-app. Without static-app, some styles are missing and the page
looks broken.

CC'ing @benthorner who helped me find the fix.

### DIFF
--- a/services/collections/docker-compose.yml
+++ b/services/collections/docker-compose.yml
@@ -37,6 +37,7 @@ services:
       - router-app-draft
       - content-store-app-draft
       - static-app-draft
+      - static-app
     environment:
       GOVUK_ASSET_ROOT: draft-collections.dev.gov.uk
       VIRTUAL_HOST: draft-collections.dev.gov.uk


kevindew:
Hmm it doesn't seem correct that a draft stack is talking to a live static instance. Do you know how this ends up being the case?

benthorner:
@kevindew I was surprised about this myself, when I was working on government-frontend. I believe it's because Static has two responsibilities: to serve the draft templates of pages, but also other asset-y stuff that's the same between draft and live. Adding the live version of Static is purely to fix the pages looking 'ugly', due to failed requests - it could be there's another env var that needs to be set.

benthorner:
@kevindew you're correct, and @ChrisBAshton I totally misinformed you earlier (sorry!). Here's a PR to fix the underlying problem: https://github.com/alphagov/govuk-docker/pull/190

PR #190, opened 2019-09-24
Fix draft static referring to live static assets

benthorner:
It makes me sad this wasn't spotted in all the other frontends we added...

Previously we depended on both the draft (draft-static.dev.gov.uk) and
app (static.dev.gov.uk) stacks of static when starting government-frontend,
since some of the network requests looked like they were hard-coded to use
the app stack. Further investigation shows this assumption was incorrect,
and that the draft stack of the static app can be corrected to correctly
refer to itself for asset URLs in the templates it renders.

### DIFF
--- a/services/government-frontend/docker-compose.yml
+++ b/services/government-frontend/docker-compose.yml
@@ -35,7 +35,6 @@ services:
       - router-app-draft
       - content-store-app-draft
       - static-app-draft
-      - static-app
     environment:
       GOVUK_ASSET_ROOT: draft-government-frontend.dev.gov.uk
       VIRTUAL_HOST: draft-government-frontend.dev.gov.uk
--- a/services/static/docker-compose.yml
+++ b/services/static/docker-compose.yml
@@ -33,5 +33,6 @@ services:
       - redis
     environment:
       REDIS_URL: redis://redis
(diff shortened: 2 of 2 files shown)


kevindew:
Nice one working this out Ben 👍 

With a bit of digging I notice this is all around the weird dark arts of asset routing that GOV.UK does (which led to the [grim nginx config](https://github.com/alphagov/publishing-e2e-tests/blob/master/docker/nginx.tmpl#L268-L373) in e2e tests). I guess this hasn't made it (yet) into govuk-docker as we don't have the concept of an asset-origin.

benthorner:
@kevindew indeed. I was talking to @ChrisBAshton yesterday about this, explaining that our aggregate 'draft-origin', 'www-origin' and 'asset' domains are pain point when developing locally, and the current environment-level workarounds aren't ideal.

When I was developing govuk-docker myself, I encountered this with the 'origin' domain most often, but I preferred having to know which actual frontend is rendering the page, starting that app specifically, and updating the 'origin' domain to refer to it. It's unclear to me how we could make this better without increasing the amount of containers. That said, it's also unclear to me how often people are actually using the 'app' and 'app-draft' stacks for the frontends.

PR #193, opened 2019-09-30
Add setup troubleshooting

ESKYoung:
Add How to guides for installation/setup issues, mainly for new
machines. Amend Dnsmasq manual setup instructions to reflect
error messages shown if file is out-of-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ The aim of govuk-docker is to make it easy to develop any GOV.UK app. It achieve
 
 ## Usage
 
+Clone your desired [service](#compatibility) into your `~/govuk` folder, e.g. collections-publisher.
+
 Do this to run the tests for a service:
 
 ```sh
@@ -52,6 +54,7 @@ govuk-docker has the following dependencies:
 - [brew](https://brew.sh/). (If you don't use a Mac, you'll need to dig into the `govuk-docker setup` command and manually install the things referenced).
 - [git](https://git-scm.com)
 - Ruby (whatever version is specified in [.ruby-version](https://github.com/alphagov/govuk-docker/blob/master/.ruby-version))
+  - Follow these [instructions to update Ruby using brew](#how-to-update-ruby)
 - A directory `~/govuk` in your home directory
 
 All other dependencies will be installed for you automatically.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,7 +241,84 @@ govuk-docker prune
...
+### How to: resolve setup issues
+
+The following sections resolve common installation issues.
+
+#### How to: update Ruby
+
+Update Ruby using [brew](https://brew.sh/), then add it to the PATH in your bash config.
+
+```
+brew install ruby
```
issyl0: The preferred way is to use `rbenv` to manage Ruby versions. With `rbenv`, an `rbenv install` in the root of the repo will install the version of Ruby specified in [`.ruby-version`](https://github.com/alphagov/govuk-docker/blob/master/.ruby-version), which is probably different (ie older) than the latest provided by Homebrew.

README.md
```diff
@@ -238,7 +241,84 @@ govuk-docker prune
...
+
+Install the correct version of bundler; you may have to overwrite existing bundle and bundler executable conflicts.
+
+```
+gem install bundler:X.Y.Z
+```
+
+#### How to: resolve issues caused by an existing docker install 
+
+During `govuk-docker setup`, if get the follow errors when pouring `docker-compose`:
```
issyl0:
```suggestion
During `govuk-docker setup`, if you get the following errors when pouring `docker-compose`:
```

README.md
```diff
@@ -238,7 +241,84 @@ govuk-docker prune
...
+
+```
+Error: It seems there is already an App at '/Applications/Docker.app'.
+```
+
+Then uninstall your existing docker, and restart the `govuk-docker setup` process to install a new version of docker using brew.
+
+#### How to: resolve `No such file or directory` errors for `dev.gov.uk`
+
+If you get the following error during `govuk-docker setup`:
```
issyl0: There's a PR to fix this error fully at #160, but these docs are good for now. 👍

ESKYoung:
Thanks for the useful feedback @issyl0 - I've made the changes you've suggested!

README.md
```diff
@@ -245,16 +245,18 @@ You should then be able to `govuk-docker build` your service and have confidence
 
 The following sections resolve common installation issues.
 
-#### How to: update Ruby
+#### How to: install/update Ruby
 
-Update Ruby using [brew](https://brew.sh/), then add it to the PATH in your bash config.
+Follow the instructions to install [rbenv](https://github.com/rbenv/rbenv#installation) using [brew](https://brew.sh/).
```
issyl0: We can replace this sentence with `brew install rbenv` in the code block below, I think.

I'm tempted to also suggest adding it to the [Brewfile](https://github.com/alphagov/govuk-docker/blob/master/Brewfile), but that only runs on `govuk-docker setup` and if you're in this section then I assume that `govuk-docker` isn't running at all.

ESKYoung: There's a few additional steps after `brew install rbenv` to complete installation, including setting up `rbenv` in the shell with an added line for `.bash_profile`, and then running the rbenv-doctor script to validate install.

Can add these all to the code block if appropriate?

issyl0: Now you mention all those steps, I'm going to say let's stick with your original suggestion of following the `rbenv` README. We don't want to duplicate every instruction.

ESKYoung: Thanks - have fixed-up branch, and will merge in PR.

PR #197, opened 2019-10-01
Remove hostname printer

benthorner:

Previously we added a helper thread to indicate where a started app
would be available. Initially the output from this thread confused me,
as it looked like it was coming from the app itself, as opposed to
govuk-docker. Despite rescuing an increasing number of exceptions,
it also still seems to error for me, or print output prematurely.
This removes the helper thread, on the basis that most domains should be
intuitive if the user has previously visited one for another app.

### DIFF
--- a/lib/govuk_docker/commands/startup.rb
+++ b/lib/govuk_docker/commands/startup.rb
@@ -1,7 +1,3 @@
-require "colorize"
-require "net/http"
-require "timeout"
-
 require_relative "./base"
 require_relative "./run"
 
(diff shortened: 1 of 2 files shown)


benthorner:
@thomasleese I know you worked on this originally in 23daf02e. What are you thoughts on keeping it?

benthorner:
@thomasleese that makes sense. Given the issues surrounding this, I'd like to see if we can make do with the [usage guidance](https://github.com/alphagov/govuk-docker#usage) in the README and see how that works as a basis for doing something like this in future.

PR #198, opened 2019-10-01
Align better with docker-compose for startup cmd

benthorner:

Previously we delegated to 'docker-compose run' when attempting to
'startup' an app, which is at odds with the conventional use of the
'docker-compose up' command for this purpose. Our over-use of 'run'
lead to some confusion:

   - 'run' recommends out the 'lite' stack when invoked via 'startup'
with a typoed variant
   - 'run' has a '--service-ports' flag, which only applies when
invoked via 'startup'

This switches the 'startup' command to do a 'docker-compose up',
which removes the need for the (now implicit) '--service-ports' flag,
and the point of confusion around recommending a 'lite' stack for
startup. Instead, the documentation now has more explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,8 @@ For a full list of govuk-docker commands, run `govuk-docker help`.
 ## Stacks
 
 Each service provides a number of different 'stacks' which you can use to run
-the app. To provide consistency we have a convention for these names:
+the app. You can see the stacks for a service in its [config file](services/content-publisher/docker-compose.yml).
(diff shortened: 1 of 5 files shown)


benthorner:
This should somewhat address [this issue](https://github.com/alphagov/govuk-docker/issues/173). Invoking `govuk-docker startup` for a pure service will still throw an error, but it will more clearly relate to the service config and avoids erroneously recommending the 'lite' stack.

PR #199, opened 2019-10-02
Remove Elasticsearch 5 from Docker builds

Tetrino:
As we're no longer relying on ES5 for our backend, running it in
Docker is an added weight we do not need. This removes the
requirement to run ES5 when running anything relying on
Search API in Docker.

Trello card: https://trello.com/c/zDmQgHQp/1039-remove-es5-from-govuk-docker-publishing-e2e-tests-and-ci

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,6 @@ volumes:
   mysql:
   mongo:
   go:
-  elasticsearch5:
   elasticsearch6:
 
 services:
@@ -43,16 +42,6 @@ services:
     image: rabbitmq
     volumes:
       - rabbitmq:/var/lib/rabbitmq
-
-  elasticsearch5:
-    image: elasticsearch:5.6.14
-    environment:
-      - http.host=0.0.0.0
-      - transport.host=127.0.0.1
-      - xpack.security.enabled=false
-      - ES_JAVA_OPTS=-Xms512m -Xmx512m
-    volumes:
-      - elasticsearch5:/usr/share/elasticsearch/data
 
   elasticsearch6:
     image: elasticsearch:6.7.0
--- a/services/search-api/docker-compose.yml
+++ b/services/search-api/docker-compose.yml
@@ -24,7 +24,6 @@ services:
     depends_on:
       - redis
       - rabbitmq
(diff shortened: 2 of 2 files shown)


Tetrino:
@benthorner included you on this in case you have any insight on places I may have missed, though I believe I have them all.

kevindew:
Should we rename elasticsearch6 to be just elasticsearch for consistency with other data backends (postgres, mysql, etc)?

Also looks like we still have ELASTICSEARCH_B_URI environment variables, and the readme still references both elasticsearch instances.

benthorner:
@Tetrino it looks like there is still some dual-ES presence in the environment variables, as @kevindew pointed out - not sure what the reference to 'readme' means, though. Do we still need to have both environment variables for search-api?

Tetrino:
> @Tetrino it looks like there is still some dual-ES presence in the environment variables, as @kevindew pointed out - not sure what the reference to 'readme' means, though. Do we still need to have both environment variables for search-api?

There are forthcoming PRs for more removals soon. The cleanup job isn't as slim as you'd hope! It's my understanding that there are some places where legacy has resulted in elasticsearch-5 being more of a machine name than an actual references to the item (typically in the Jenkinsfile) so there may be some leftover flags here and there anyway.

We'll be doing a final review and catching anything we've missed when we're done before we finalise the cards related to this, no worries.

Tetrino:
> Should we rename elasticsearch6 to be just elasticsearch for consistency with other data backends (postgres, mysql, etc)?
> 
> Also looks like we still have ELASTICSEARCH_B_URI environment variables, and the readme still references both elasticsearch instances.

Sorry to clarify the above, because of managing how puppet is handling the removal of ES5 instances we're having to roll things out as a stepped launch. So eventually all the variables etc will be renamed as appropriate, but we need puppet to do its thing before we can commit any of those changes.

barrucadu:
I don't think we should rename `elasticsearch6` to `elasticsearch`, as when we eventually upgrade to 7 we'll probably end up with ES6 and ES7 running in parallel for a bit.  Making the major version number part of the service name is a graceful way to handle that.

PR #201, opened 2019-10-03
Add missing deps for collections-publisher-e2e

benthorner:
Previously we overwrote the collections-publisher-app dependencies with
just 'collections-publisher-worker', missing out a couple of necessary
services that are included with the 'app' stack. This adds the missing
dependencies that aren't already provided by 'publishing-api-e2e'.

### DIFF
--- a/services/collections-publisher/docker-compose.yml
+++ b/services/collections-publisher/docker-compose.yml
@@ -55,4 +55,9 @@ services:
   collections-publisher-app-e2e:
     <<: *collections-publisher-app
     depends_on:
+      - redis
+      - mysql
+      - content-store-app
+      - link-checker-api-app
+      - nginx-proxy-app
       - collections-publisher-worker


services/collections-publisher/docker-compose.yml
```diff
@@ -55,4 +55,6 @@ services:
   collections-publisher-app-e2e:
     <<: *collections-publisher-app
     depends_on:
+      - link-checker-api-app
+      - nginx-proxy-app
```
ChrisBAshton: Just so I'm clear:

* YAML inheritance overwrites (rather than merges) properties, so we need to repeat `depends_on` values between aliases
* But we don't need to include `mysql` and `redis` here as they're already depended on by `collections-publisher-worker`, which we're pulling in below.
* And we don't need to include `publishing-api-app`, as the `collections-publisher-worker` we're depending on depends on `publishing-api-app-e2e` - and this brings in `content-store-app`, which was the other dependency we'd have had to have carried over from `collections-publisher-app`.

Is that right? And if so I'm wondering if that would be more clearly expressed by explicitly listing all of the dependencies, even if that means some duplication. That way each service draft configuration is its own 'thing' and doesn't require scrolling up and down to figure out what its 'actual' dependencies are.

benthorner: @ChrisBAshton all good questions/things-to-check. You're right in all cases, and I agree with your suggestion: my default preference is to be minimal, but I do see it's informative and less volatile to list the direct dependencies of a service on the service itself, rather than rely on transitive dependencies. I've updated the commit to reflect this 👍.

ChrisBAshton: Thanks - looks good to merge :)

PR #202, opened 2019-10-03
Add how-to address issue with new Ruby release

benthorner:


README.md
```diff
@@ -453,6 +453,21 @@ While most environment variables should be set in the config for a service, some
...
+ruby-build: definition not found: x.y.z
+```
+
+Most of our services share a common image, which we needs rebuilding to be aware of the new Ruby version. So, to fix the error, run the following commands, where `<service>` is e.g. `collections-publisher`.
+
+```
+govuk-docker compose build --no-cache <service>-lite
+govuk-docker build
+```
+
```
huwd: So will this process update the `.ruby-version` in a given repo?

If so is that worth stating? I always find it useful to check in on that file if I'm having issues, seeing it change might be a useful acceptance criteria for knowing this worked?

README.md
```diff
@@ -453,6 +453,21 @@ While most environment variables should be set in the config for a service, some
...
 
+### How to: install a new release of Ruby
+
+Many of our services use a `.ruby-version` file in conjunction with `rbenv`. When a new version of Ruby is released and we start upgrading our services, you may start seeing the following error when you run commands.
+
+```
+ruby-build: definition not found: x.y.z
+```
+
+Most of our services share a common image, which we needs rebuilding to be aware of the new Ruby version. So, to fix the error, run the following commands, where `<service>` is e.g. `collections-publisher`.
```
huwd: Is it work stating "docker image"? I know this is a docker repo, but i always find image a very slippary word!

JonathanHallam: I think this might be clearer along the lines of 
`So, to fix the error, run the following commands, replacing '<service>' with the name of the service, e.g. 'collections-publisher'`
Fully recognise that this is a bit pedantic though so am approving :)

issyl0: I agree with Jon. And remove the "So,"?

README.md
benthorner: Could you help me understand how you reached that conclusion? Perhaps it's worth pointing out the title of this section: 'install a new release of Ruby' - I don't think that implies updating anything.

README.md
benthorner: Sure. Updated 👍.

benthorner: Sure. I notice we've used 'Docker container' elsewhere too, so it seems like we have a precedent of being specific. Updated 👍.

benthorner:
@huwd thanks for the comments :-). In answer to your question about when someone will know this is fixed, it's as simple as "you won't get the error anymore".

README.md
```diff
@@ -453,6 +453,20 @@ While most environment variables should be set in the config for a service, some
...
 
+### How to: install a new release of Ruby
+
+Many of our services use a `.ruby-version` file in conjunction with `rbenv`. When a new version of Ruby is released and we start upgrading our services, you may start seeing the following error when you run commands.
+
+```
+ruby-build: definition not found: x.y.z
+```
+
+Most of our services share a common Docker image, which we needs rebuilding to be aware of the new Ruby version. To fix the error, run the following commands, replacing '<service>' with the name of the service, e.g. 'collections-publisher'.
```
maxgds: Remove "we" from "which we needs rebuilding"

benthorner: Good spot - updated 👍.

benthorner:
Blimey - my PR is getting a lot of love 😃.

benthorner:
After discussing this offline with @huwd, I think we're good to merge...

PR #205, opened 2019-10-03
Remove vague troubleshooting how-to

benthorner:
The original commit of govuk-docker included a how-to with a bunch of
'useful' commands. All of these can be found in other documentation or
'--help' output, and I'm not convinced it's helpful unless people have
a deeper understanding of how things work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,27 +200,6 @@ This will test whether or not your system meets the following requirements:
 * docker installed
 * docker-compose installed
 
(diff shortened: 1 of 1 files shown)


injms:
I'd much prefer this to stay in. As someone with very little knowledge of Docker, I've found these quick commands a useful reference and jumping off point to further investigation.

benthorner:
@injms fair enough - I'm glad they're useful :-).

PR #211, opened 2019-10-08
Add missing services to show what needs doing

benthorner:
https://trello.com/c/dewpTNKL/109-add-service-for-release-app


benthorner:
Sure, we can add more. These were just the ones Tijmen identified in the original Trello cards.

thomasleese:
I suspect not having Mapit will prevent us from being able to delete the dev VM, so I'll raise a PR to add it to the list.

PR #214, opened 2019-10-08
Add license-finder service

chao-xian:
The app needs the govuk_publishing_components updated and
was not added as a service before, hampering local development.

licence-finder dev gov uk_licence-finder_sectors (1)

## Trello cards

* https://trello.com/c/hL2TREil
* https://trello.com/c/auN1ZBy2

### DIFF
--- a/services/licence-finder/Makefile
+++ b/services/licence-finder/Makefile
@@ -1,0 +1,2 @@
+licence-finder: bundle-licence-finder static content-store
+	$(GOVUK_DOCKER) run $@-lite sh -c 'bin/rake db:migrate 2>/dev/null || bin/rake db:setup'
--- a/services/licence-finder/docker-compose.yml
+++ b/services/licence-finder/docker-compose.yml
@@ -1,0 +1,42 @@
+version: '3.7'
+
+x-licence-finder: &licence-finder
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: licence-finder
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/licence-finder
+
+services:
+  licence-finder-lite: &licence-finder-lite
+    <<: *licence-finder
+    privileged: true
+    depends_on:
+      - mongo
+      - elasticsearch6
+    environment:
+      MONGODB_URI: "mongodb://mongo/licence-finder_development"
+      ELASTICSEARCH_URI: http://elasticsearch6:9200
+
+  licence-finder-app:
+    <<: *licence-finder
+    depends_on:
+      - mongo
+      - redis
+      - static-app
+      - elasticsearch6
+      - content-store-app
+      - nginx-proxy-app
+    environment:
+      REDIS_URL: redis://redis
+      VIRTUAL_HOST: licence-finder.dev.gov.uk
+      MONGODB_URI: "mongodb://mongo/licence-finder_development"
+      ELASTICSEARCH_URI: http://elasticsearch6:9200
+      PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    command: bin/rails s --restart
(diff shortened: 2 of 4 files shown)


services/licence-finder/Makefile
```diff
@@ -0,0 +1,2 @@
+licence-finder: bundle-licence-finder static
+	$(GOVUK_DOCKER) compose run $@-lite sh -c 'bin/rake db:migrate 2>/dev/null || bin/rake db:setup'
```
benthorner: This looks like it needs a rebase?

services/licence-finder/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+  image: licence-finder
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/licence-finder
+
+services:
+  licence-finder-lite: &licence-finder-lite
+    <<: *licence-finder
+    privileged: true
```
benthorner: Are we sure this is necessary?

services/licence-finder/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    depends_on:
+      - mongo
+      - elasticsearch6
+    environment:
+      MONGODB_URI: "mongodb://mongo/licence-finder_development"
+      VIRTUAL_HOST: licence-finder.dev.gov.uk
+      ELASTICSEARCH_URI: http://elasticsearch6:9200
+      HOST: 0.0.0.0
```
benthorner: I don't think we need these for the 'lite' stack. Might be worth reading [here](https://github.com/alphagov/govuk-docker#stacks).

services/licence-finder/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    ports:
+      - "3000"
+
+  licence-finder-app:
+    <<: *licence-finder
+    depends_on:
+      - mongo
```
benthorner: I think we're missing an 'nginx-proxy' dependency here.

services/licence-finder/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+      - redis
+      - static-app
+      - elasticsearch6
+      - content-store-app
+    environment:
+      GOVUK_ASSET_ROOT: frontend.dev.gov.uk
```
benthorner: I'm not sure if this is right, especially as this doesn't depend on 'frontend'. Does it work without?

services/licence-finder/docker-compose.yml
chao-xian: Tests couldn't run as Chrome was not available. If iirc this was required for chromedriver.

services/licence-finder/docker-compose.yml
chao-xian: Again, tests until weren't running I added these in.

services/licence-finder/docker-compose.yml
chao-xian: Hmm this is part of static now but worth adding here to be explicit?

services/licence-finder/docker-compose.yml
chao-xian: This was un-needed. I'm not sure why I added it...

services/licence-finder/Makefile
chao-xian: Oh yeah thanks

services/licence-finder/docker-compose.yml
benthorner: Ah fair enough. I couldn't see where it was using it!

services/licence-finder/docker-compose.yml
benthorner: Yup. We should add things that this app depends on directly, to avoid coupling with other services.

services/licence-finder/docker-compose.yml
benthorner: I've managed to run the tests without them (`ports`, `VIRTUAL_HOST` and `HOST`).

PR #216, opened 2019-10-09
Use a simple bash script for day-to-day commands

benthorner:

Originally govuk-docker was a thin bash wrapper script, which
 met the following needs:

   - Reduce typing common flags, like `-f` to specify config files
   - Being able to run the environment from a project directory
   - Auto-inference of the service to start, based on the directory

Since that time, the script has been converted to Ruby and extended
with many other commands, but I feel this has come at the cost of
making the day-to-day commands simple to maintain and understand.
I think we should consider moving (back) to something simpler:

   - A (very) thin, non-extensible wrapper for day-to-day commands
     (`run`, `up`, `ps`, `logs`, which delegates entirely to `docker-compose`
   - A set of other scripts that continue to provide the other,
     less frequently used commands we added, such the `doctor` command


bin/govuk-docker
```diff
@@ -1,23 +1,11 @@
...
-end
-
-require "rubygems"
-require "bundler/setup"
-
-require_relative "../lib/govuk_docker"
-require_relative "../lib/govuk_docker/cli"
-GovukDocker::CLI.start(ARGV)
+echo "docker-compose -f [...] $@"
+docker-compose $COMPOSE_FLAGS "$@"
```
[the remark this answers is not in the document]
kevindew: Actually I had what might be a better idea. 

You could use the introduction of this as a move of govuk-docker from bin to exe and then could set up the govuk-docker in bin to explain what's moved and how to switch.

docs/installation.md
```diff
@@ -36,11 +36,34 @@ Now in the `~/govuk` directory, run the following commands.
...
+}
+
+function govuk_docker_startup () {
+  STACK=${1:-app}
+  govuk-docker up $(basename $(pwd))-$STACK
+}
+
+alias gd="govuk-docker"
+alias gdr="govuk_docker_run"
+alias gds="govuk_docker_startup"
```
issyl0: This alias conflict horribly with the `gds-cli`'s main entrypoint of `gds`. I'm not sure how many people use it around here (where's the :can-of-worms: emoji?) but re-defining this will be problematic for those in RE or who are RE-adjacent or who choose it as their tool of choice independently.

bin/govuk-docker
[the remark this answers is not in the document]
benthorner: Good point. I've splashed 'echos' around the `bin/` version.

barrucadu:
I'm generally in favour of shell scripts and feel that ruby can get very unwieldy at times (eg, the verbose AWS auth wrangling in the elasticsearch data sync PR), but wasn't the motive behind rewriting this in ruby in the first place that most people aren't so comfortable with shell scripts?

benthorner:
@barrucadu thanks for the upvote! I think the best answer to your question is to note that the new shell script in this PR is actually shorter than the one we've got with Ruby 😆. 

As I understand it, the reason we moved to Ruby was motivated by the desire to add commands like 'doctor' and 'setup', although the existing script was also getting a little long. The aim of this PR is to demo how we can address concerns with shell scripts by continuing to use Ruby for infrequent commands, while also being more frugal about the CLI, so it's just a wrapper.

bin/govuk-docker
```diff
@@ -1,23 +1,44 @@
...
+    echo 'https://github.com/alphagov/govuk-docker/blob/e00ab054d8d94132f1539b0027eefbe44e168f6f/README.md#usage'
+    exit 1
+    ;;
+esac
 
-require "rubygems"
-require "bundler/setup"
+echo '*** DEPRECATED: use exe/govuk-docker instead ***'
+echo 'https://github.com/alphagov/govuk-docker/blob/e00ab054d8d94132f1539b0027eefbe44e168f6f/docs/installation.md#setup'
+echo
```
kevindew: Should this also be in the case statement for an empty option (I say this with virtually no knowledge of bash case statements) to prevent you receiving two outputs if you have an argument.

bin/govuk-docker
```diff
@@ -1,23 +1,44 @@
...
+    echo 'https://github.com/alphagov/govuk-docker/blob/e00ab054d8d94132f1539b0027eefbe44e168f6f/docs/installation.md\#shortcuts'
+    ;;
+
+  prune)
+    echo '*** DEPRECATED: use exe/govuk-docker rm -sv instead ***'
+    echo 'https://github.com/alphagov/govuk-docker/blob/e00ab054d8d94132f1539b0027eefbe44e168f6f/README.md#how-to-clear-your-docker-containers'
+    ;;
+esac
+
+exit 1
```
kevindew: This looks really helpful for communicating the change 👍 

Couple of things: 
- we shouldn't use the terminology DEPRECATED here as this goes further than deprecation and has removed the functionality from these commands. Maybe replace with DEPRECATED with MOVED"?
- might be better to link to master branch rather than the commit SHA to reflect most up-to-date version of documentation.

docs/installation.md
```diff
@@ -36,11 +36,34 @@ Now in the `~/govuk` directory, run the following commands.
...
+Typing the [full commands](../README.md#usage) is likely to get tiring. We recommend adding the following functions (and aliases) to your shell config, which figure out the application you're running based on the name of the current directory, matching it with the corresponding docker-compose.yml in govuk-docker/services.
+
+```
+function govuk_docker_run () {
+  STACK=lite
+  govuk-docker run $(basename $(pwd))-$STACK "$@"
+}
+
+function govuk_docker_up () {
+  STACK=${1:-app}
```
kevindew: I think this should be prefixed with local so to not overwrite any existing global variables

docs/installation.md
```diff
@@ -36,11 +36,34 @@ Now in the `~/govuk` directory, run the following commands.
...
 You can now [clone and setup the apps you need](../README.md#Usage), after which you can do things like run tests and startup the app in your browser. If this doesn't work for whatever reason, follow the [instructions on how to resolve setup issues](#how-tos).
 
+### Shortcuts
+
+Typing the [full commands](../README.md#usage) is likely to get tiring. We recommend adding the following functions (and aliases) to your shell config, which figure out the application you're running based on the name of the current directory, matching it with the corresponding docker-compose.yml in govuk-docker/services.
+
+```
+function govuk_docker_run () {
+  STACK=lite
+  govuk-docker run $(basename $(pwd))-$STACK "$@"
```
kevindew: Do we need to set the STACK variable here? it might be simpler to have just `govuk-docker run $(basename $(pwd))-lite "$@"`

docs/installation.md
```diff
@@ -36,11 +36,34 @@ Now in the `~/govuk` directory, run the following commands.
...
+function govuk_docker_up () {
+  STACK=${1:-app}
+  govuk-docker up $(basename $(pwd))-$STACK
+}
+
+alias gd="govuk-docker"
+alias gdr="govuk_docker_run"
+alias gdu="govuk_docker_up"
+alias gdbe="govuk_docker_run bundle exec"
+```
```
kevindew: I'm pleased that only gd conflicts with the [oh my zsh git aliases](https://github.com/robbyrussell/oh-my-zsh/wiki/Cheatsheet#git)

docs/installation.md
```diff
@@ -36,11 +36,34 @@ Now in the `~/govuk` directory, run the following commands.
 git clone git@github.com:alphagov/govuk-docker.git
 cd govuk-docker
 bundle install
-govuk-docker setup
+bundle exec bin/setup
```
kevindew: Can we avoid the need to run these bins inside `bundle exec`?

As I understand from https://bundler.io/guides/bundler_setup.html we can add:

```
require 'rubygems'
require 'bundler/setup'
```

for that to work.

exe/govuk-docker
```diff
@@ -0,0 +1,21 @@
...
+
+# **************************************************************
+# DO NOT MODIFY THIS FILE
+#
+# This is meant to be a simple wrapper script to pass some '-f'
+# flags to docker-compose. Its behaviour should not depend on
+# its arguments. We do not want our own bespoke CLI, as this can
+# only lead to complexity, performance and maintenance overheads
+# on top of docker and docker-compose.
+# **************************************************************
```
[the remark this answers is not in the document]
benthorner: Yeah, that's fair. I was going all out with the initial comment! I've updated it to 'DO NOT ADD NEW STUFF'.

bin/govuk-docker
benthorner: Ah, so this is deliberate, as I think we should print out the generic bit about using `exe/govuk-docker` before talking about any specific sub-commands that have changed. In the first case, we're pointing to the new installation docs; in the specific cases, we're pointing to the alt thing to run.

bin/govuk-docker
benthorner: I was thinking the SHA would be more stable, as the stuff on master might change if this gets merged in, and it might be more helpful to link to something that was valid at a point in time. In some cases, like installation, the master version doesn't exist yet, so this seems like the way to go for the time being.

kevindew: > I was thinking the SHA would be more stable, as the stuff on master might change if this gets merged in, and it might be more helpful to link to something that was valid at a point in time. In some cases, like installation, the master version doesn't exist yet, so this seems like the way to go for the time being.

Yup, both approaches have downsides. The SHA hash seems fine if this doesn't hang around too long otherwise it runs the risk of pointed to superseded info.

kevindew:
Nice one - All the changes you've made look good to me and I'm happy to approve but feel we should wait until after the meeting next week.

bin/govuk-docker
```diff
@@ -1,23 +1,44 @@
-#!/usr/bin/env ruby
+#!/bin/bash
```
barrucadu: `/bin/bash` isn't portable, that should be `/usr/bin/env bash`

benthorner: Oops. I can't think why I made that change now - reverted!

benthorner:
@barrucadu can do. I notice it only covers '*.sh' files, of which there are none in the `exe/` directory. Should we do it preemptively, or change the files it looks for?

barrucadu:
I think if the line is changed to `$(SHELLCHECK) $(shell ls ${GOVUK_DOCKER_DIR}/bin/*.sh ${GOVUK_DOCKER_DIR}/exe/*)` it'll work fine

benthorner:
Dah, it made me write better bash. How annoying.

PR #217, opened 2019-10-10
Add SSH command

bilbof:
This enables users to open a shell in a running container in a given stack.

The use case is that one would like to do various things in a running container. There is the run command but it is nicer to have a shell to work in directly.

Usage:

```
govuk-docker ssh
```

This will start up a bash shell in the running container.

https://trello.com/c/bEEvYDFc/110-ssh-command

### DIFF
--- a/lib/govuk_docker/cli.rb
+++ b/lib/govuk_docker/cli.rb
@@ -4,6 +4,7 @@ require_relative "./commands/build"
 require_relative "./commands/compose"
 require_relative "./commands/prune"
 require_relative "./commands/run"
+require_relative "./commands/ssh"
 require_relative "./commands/startup"
 require_relative "./doctor/doctor"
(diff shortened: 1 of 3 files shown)


issyl0:
Yet Another SSH Tool. Erk. How does this interact with the currently draft #216?

benthorner:
Neat idea @bilbof! I think we need to be careful about the 'ssh' terminology here, though, as that's not what's actually happening and may address @issyl0's comment above. Given this should be an infrequent command, and it's just a one-liner, I'd suggest a "how-to" is a simpler and more discoverable approach here. 

In general, we should try to align as close as possible to docker/compose itself, since the 3 layers of CLI we have right now are confusing people. It strikes me that some regular Docker training would resolve a lot of issues! This may also be of interest: https://github.com/alphagov/govuk-docker/pull/216.

PR #219, opened 2019-10-16
Tidy up our the 'test' rule

benthorner:
https://trello.com/c/2bxWjW1c/95-we-shouldnt-have-binstubs-if-we-expect-the-bin-directory-to-be-in-a-users-path

This removes a test script in the bin/ directory and reproduces its
functionality by calling 'govuk-docker compose config' directly. We don't
particularly benefit from a friendly 'success message', and having a test
script in the bin/ directory is surprising/confusing.


Makefile
```diff
@@ -20,15 +20,12 @@ pull:
 	echo $(APPS) | cut -d/ -f3 | xargs -P8 -n1 ./bin/update-git-repo.sh
 
 test:
-	# Linting
-	bundle exec rubocop . --parallel
```
thomasleese: Shouldn’t we still keep this if we’re going to at least have some parts still written in Ruby?

thomasleese: Hmm, I've read the commit message.

Jenkins does `bundle exec govuk-lint-ruby --parallel --format html --format clang app lib spec test` and I'm concerned that if we start to move Ruby files out of those folders we might lose linting without realising. Other than just remembering that and changing [`rubyLintDirs`](https://github.com/alphagov/govuk-jenkinslib/blob/5fe7a435cc54f9f162d8167016d5238139977c4c/vars/govuk.groovy#L163) as necessary I'm not sure how we could prevent that from happening though.

thomasleese: If we think we'll retain the standard `lib` and `spec` dirs then I think it's fine.

PR #220, opened 2019-10-17
Add per-app data replication scripts

barrucadu:
This PR adds scripts to replicate databases.  **Unlike the dev VM, this does replication in terms of apps, not in terms of databases.**  There are a couple of big differences:

1. The old scripts download *all* databases of a given type (eg, all postgres databases), and use a command-line blacklist to prevent restoring certain databases.  These scripts download *one* database and only restore that one.

2. The old scripts required you to know the name of the database each app uses, as database backups are named after the database, not the app.  These scripts require you to know only the type of database uses (postgres, mysql, etc), and internally translates between the database name and the app name.

I think these changes make sense for most people, as we usually don't want all the data.  But there are some downsides:

1.  If you want to replicate all databases, you need to run the scripts multiple times.  This could be alleviated by having a helper script to do that.

2. There's extra complication in having the app name / database name mapping.  But we don't ever rename databases (to my knowledge), so there's not really a maintenance burden.  This could be alleviated by having app names and database names the same, but that would be a large change and probably warrant an RFC.

### To do:

- [x] MongoDB
- [x] Documentation
- [x] ShellCheck in CI

### Things I'd like feedback on:

- ~Where should replicated data be stored?  I've gone for `$HOME/govuk-data-sync`.  Should there be an environment variable for that?  Should it be somewhere inside `$GOVUK_DOCKER_DIR` or `$GOVUK_ROOT_DIR`?~ answered: `${GOVUK_DOCKER_REPLICATION_DIR:-$GOVUK_DOCKER_DIR/replication}`

- Do people want this "app-oriented" replication or would they prefer "database-oriented" replication like before?

- ~Do people like bash scripts for these, or would you prefer it to be Ruby even if it ended up being much longer?~ Ruby has been cast out (sort of)

- Would these scripts address your replication needs, or is there something else you want?

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,8 +126,6 @@ All the scripts, other than `replicate-elasticsearch.sh`, take the name of the a
 
 Draft data can be replicated with `replicate-mongodb.sh draft-content-store` and `replicate-mongodb.sh draft-router`.
 
-If the script hangs at "waiting for mongo..." (or whatever the database is), try stopping your govuk-docker containers and running the script again.  Data is cached, so it won't be downloaded a second time.
-
 ### How to: set environment variables
 
 While most environment variables should be set in the config for a service, sometimes it's necessary to set assign one or more variables at the point of running a command, such as a Rake task. This can be done using `env` e.g.
--- a/README.md
+++ b/README.md
@@ -117,6 +117,10 @@ gem 'govuk_publishing_components', path: '../govuk_publishing_components'
 
 There may be times when a full database is required locally.  The following sections give examples of how to replicate this data from integration.  All examples require pv, which can be installed on a Mac using Brew (`brew install pv`).
 
+#### Elasticsearch
+
+TODO
+
 #### MySQL
 
 1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/mysql/?region=eu-west-1&tab=overview)
--- a/README.md
+++ b/README.md
@@ -131,38 +131,7 @@ TODO
 
 #### MongoDB
 
-1.  Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/mongodb/daily/mongo/?region=eu-west-1&tab=overview)
-
-2. Unzip the archive, e.g. for Content Store:
-
-```
-gunzip mongodump-2019-08-12_0023.tgz
-```
-
-Or if it's a TAR file, you can extract a specific file or directory.  Using the Content Store as an example:
-```
-tar -xvzf mongodump-2019-08-12_0023.tar var/lib/mongodb/backup/mongodump/content_store_production -C directory_for_download
-```
-
-3. Update the `docker-compose.yml` file to mount your local directory into the VM, e.g.
-
-```
-  mongo:
-    image: mongo:2.4
-    volumes:
-      - mongo:/data/db
-      - /Path/To/Downloads/directory_for_download:/import
-    ports:
-      - "27017:27017"
-      - "27018:27018"
-```
-
-4. Import the backup files into the local Mongo database, e.g. for Content Store:
-
-```
-govuk-docker up -d mongo
-govuk-docker run mongo mongorestore --drop --db content-store /import/var/lib/mongodb/backup/mongodump/content_store_production/
-```
+TODO
 
 ### How to: set environment variables
 
--- a/README.md
+++ b/README.md
@@ -127,21 +127,7 @@ TODO
 
 #### PostgreSQL
 
-1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/postgres/?region=eu-west-1&tab=overview)
-
-2. Drop and recreate any existing database, e.g. for Publishing API:
-
-```
-govuk-docker up -d postgres
-govuk-docker run postgres /usr/bin/psql -h postgres -U postgres -c 'DROP DATABASE IF EXISTS "publishing-api"'
-govuk-docker run postgres /usr/bin/createdb -h postgres -U postgres publishing-api
-```
-
-3. Import the file into the local Postgres database, e.g. for Publishing API:
-
-```
-pv publishing_api_production.dump.gz  | gunzip | govuk-docker run postgres /usr/bin/psql -h postgres -U postgres -qAt -d publishing-api
-```
+TODO
 
 #### MongoDB
 
--- a/README.md
+++ b/README.md
@@ -123,21 +123,7 @@ TODO
 
 #### MySQL
 
-1. Download the relevant database dump from the [AWS S3 Bucket](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups/mysql/?region=eu-west-1&tab=overview)
-
-2. Drop and recreate any existing database, e.g. for Whitehall:
-
-```
-govuk-docker up -d mysql
-govuk-docker run mysql mysql -h mysql -u root --password=root -e "DROP DATABASE IF EXISTS whitehall_development"
-govuk-docker run mysql mysql -h mysql -u root --password=root -e "CREATE DATABASE whitehall_development"
-```
-
-3. Import the file into the local MySQL database, e.g. for Whitehall:
-
-```
-pv whitehall_production.dump.gz | gunzip | govuk-docker run mysql mysql -h mysql -u root --password=root whitehall_development
-```
+TODO
 
 #### PostgreSQL
 
--- a/README.md
+++ b/README.md
@@ -115,23 +115,18 @@ gem 'govuk_publishing_components', path: '../govuk_publishing_components'
 
 ### How to: replicate data locally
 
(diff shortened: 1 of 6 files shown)


benthorner:
@barrucadu thanks for putting this up. The first thing I noticed was how compact the replication scripts are. True, they're not 'small', but they're also pretty simple in their operation. I was expecting to see a load of complex AWS/DB manipulation, for some reason.

With the storage directory, I think it'd make more sense to use a `.gitignore`d directory in govuk-docker itself. The reason is simply that the files created are tied to these scripts, which are tied to govuk-docker. Placing it outside of govuk-docker could make its origin unclear.

I haven't actually tried this out yet, so these are just some superficial comments! I'll try and make some time to import a few databases and see what happens.

bin/replicate-mongodb.sh
```diff
@@ -0,0 +1,96 @@
...
+
+if [[ -e "$archive_path" ]]; then
+  echo "Skipping download - remove ${archive_path} to force"
+else
+  remote_file_name=$(aws --profile govuk-integration s3 ls "s3://${bucket}/mongodb/daily/${hostname}/" | tail -n1 | sed 's/^.* .* .* //')
+  aws --profile govuk-integration s3 cp "s3://${bucket}/mongodb/daily/${hostname}/${remote_file_name}" "$archive_path"
+fi
+
+cat_cmd='cat'
+if command -v pv &>/dev/null; then
```
benthorner: Minor: since this is in the Brewfile, can we assume it'll be present?

bin/replicate-mongodb.sh
```diff
@@ -0,0 +1,96 @@
...
+  rm -r "$extract_path"
+fi
+mkdir -p "$extract_path"
+
+"$cat_cmd" "$archive_path" | gunzip | tar -zx -f - -C "$extract_path" "var/lib/mongodb/backup/mongodump/${database}"
+
+container=$(govuk-docker compose run -d --rm -v "${extract_path}:/replication" mongo | tail -n1)
+trap 'docker stop "$container"' EXIT
+
+echo "CONTAINER: $container"
```
benthorner: What's this for?

.gitignore
```diff
@@ -1,2 +1,2 @@
-
+replication
```
benthorner: Can we also add a `.dockerignore` with this, as otherwise it'll become part of the build context.

bin/replicate-mysql.sh
```diff
@@ -0,0 +1,51 @@
...
+set -eu
+
+if [[ "$#" == "0" ]]; then
+  echo "usage: $0 \$app"
+  exit 1
+fi
+
+app="$1"
+
+if echo "$app" | grep -q '_'; then
```
benthorner: Same here.

bin/replicate-mysql.sh
```diff
@@ -0,0 +1,51 @@
...
+
+echo "Replicating mysql for $app"
+
+if [[ -e "$archive_path" ]]; then
+  echo "Skipping download - remove ${archive_path} to force"
+else
+  aws --profile govuk-integration s3 cp "s3://${bucket}/mysql/$(date '+%Y-%m-%d')/${archive_file}" "${archive_path}"
+fi
+
+cat_cmd='cat'
```
benthorner: Same here.

bin/replicate-mysql.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+replication_dir="${GOVUK_DOCKER_REPLICATION_DIR:-${GOVUK_DOCKER_DIR:-${GOVUK_ROOT_DIR:-$HOME/govuk}/govuk-docker}/replication}"
+
+bucket="govuk-integration-database-backups"
+archive_dir="${replication_dir}/mysql"
+archive_file="${app//-/_}_production.dump.gz"
+archive_path="${archive_dir}/${archive_file}"
+
+mkdir -p "$archive_dir"
```
benthorner: Same here.

bin/replicate-elasticsearch.sh
```diff
@@ -0,0 +1,51 @@
...
+    "compress": true,
+    "readonly": true,
+    "location": "/replication"
+  }
+}'
+
+sleep 5
+
+snapshot_name=$(curl "http://127.0.0.1:9200/_snapshot/snapshots/_all" | ruby -e 'require "json"; STDOUT << (JSON.parse(STDIN.read)["snapshots"].map { |a| a["snapshot"] }.sort.last)')
+curl -XPOST "http://127.0.0.1:9200/_snapshot/snapshots/${snapshot_name}/_restore?wait_for_completion=true"
```
benthorner: Just comparing with the other scripts: does this work if you run it again?

bin/replicate-postgresql.sh
```diff
@@ -0,0 +1,59 @@
...
+    archive_file="support_contacts_production.dump.gz"
+    ;;
+  *)
+    archive_file="${app//-/_}_production.dump.gz"
+    ;;
+esac
+
+archive_path="${archive_dir}/${archive_file}"
+
+mkdir -p "$archive_dir"
```
benthorner: Same here.

README.md
```diff
@@ -130,78 +130,20 @@ gem 'govuk_publishing_components', path: '../govuk_publishing_components'
...
-      - "27018:27018"
-```
-
-4. Import the backup files into the local Mongo database, e.g. for Content Store:
-
-```
-govuk-docker compose up -d mongo
-govuk-docker compose run mongo mongorestore --drop --db content-store /import/var/lib/mongodb/backup/mongodump/content_store_production/
-```
+If the script hangs at "waiting for mongo..." (or whatever the database is), try stopping your govuk-docker containers and running the script again.  Data is cached, so it won't be downloaded a second time.
```
benthorner: Minor: it might be helpful (since I experienced it) to mention doing a `govuk-docker compose down` beforehand, since this can lock DBs. Perhaps this should be part of the scripts?

bin/replicate-mongodb.sh
barrucadu: Ok, I'll remove this

bin/replicate-mongodb.sh
barrucadu: That *was* for debugging

bin/replicate-elasticsearch.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [[ -e "$archive_path" ]]; then
+  echo "Skipping download - remove ${archive_path} to force"
+else
+  aws --profile govuk-integration s3 sync "s3://${bucket}/" "${archive_path}/"
+fi
+
+# temporary config file because ES needs to be configured in advance
+# for filesystem-based snapshots
+cfg_path=$(mktemp '/tmp/govuk-docker-data-sync.XXXXX')
```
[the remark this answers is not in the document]
barrucadu: I'm not sure what you mean - we'll be creating a file on disk regardless of what it's called, so it should ideally be cleaned up afterwards.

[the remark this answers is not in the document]
benthorner: Ah, sorry - this is me not reading the manual for mktemp!

bin/replicate-elasticsearch.sh
barrucadu: Yes.  I just ran the script, deleted some indices, re-ran the script, and everything is back as it was.

bin/replicate-mongodb.sh
```diff
@@ -0,0 +1,96 @@
...
+set -eu
+
+if [[ "$#" == "0" ]]; then
+  echo "usage: $0 \$app"
+  exit 1
+fi
+
+app="$1"
+
+if echo "$app" | grep -q '_'; then
```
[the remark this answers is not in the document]
thomasleese: Could we automatically convert underscores into dashes (while telling the user what's happening)?

This feels like one of those times when the script understands what you were trying to do but refuses to do so anyway, similar to https://github.com/alphagov/fabric-scripts/pull/285.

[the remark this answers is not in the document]
benthorner: I'm cautious about this kind of thing. Although it's useful, it's also good to encourage people to know the correct names, for when there isn't a script to help them. Perhaps there's a case when it's particularly unintuitive?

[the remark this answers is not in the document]
thomasleese: I'm also generally cautious of these kinds of things, but I think in this case it's reasonable since the behaviour is very predictable. I think it's better than the previous situation where the script understands the error the user has made and how to correct it, but makes no attempt to correct it (which is a frustrating user experience).

Perhaps an alternative is that the script errors saying something like `Could not find database backup for "content_store" to replicate` after trying to download the backup, rather than doing a preemptive check. This has the benefit of showing the same error if someone misspells the app name or if they use an underscore.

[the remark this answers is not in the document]
benthorner: I agree - it's very similar to rubocop (not) autocorrecting stuff, which used to infuriate me. My issue with rubocop went away when I realised it was just telling me what I should be typing in the first place. But I understand that's not how everyone sees it. I like your suggestion, as that error handler sounds more useful in other cases (mis-spellings).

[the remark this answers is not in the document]
barrucadu: The `aws s3 cp` commands will already fail saying something about being unable to find the object, is a more specific error message needed?

barrucadu:
@thomasleese @benthorner This PR has been hanging around for a bit now, do you have any more comments or are you happy for this to be merged?

PR #221, opened 2019-10-22
Add authenticating-proxy to services

kevindew:
Depends on https://github.com/alphagov/authenticating-proxy/pull/167

This adds the authenticating-proxy app which can run in front of draft-router.
In order to run it against frontend apps they'll also need to be started
up as draft router doesn't have dependencies on them.

This resolves running authenticating-proxy in the default GDS SSO "mock"
mode. In order to run this with GDS SSO in "real" mode there needs to be
co-ordination between this authenticating-proxy and signon to share
oauth data, which was out of the scope of this change.

### DIFF
--- a/services/authenticating-proxy/docker-compose.yml
+++ b/services/authenticating-proxy/docker-compose.yml
@@ -1,0 +1,36 @@
+version: '3.7'
+
+x-authenticating-proxy: &authenticating-proxy
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: authenticating-proxy
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/authenticating-proxy
+
+services:
+  authenticating-proxy-lite:
+    <<: *authenticating-proxy
+    depends_on:
+      - mongo
+    environment:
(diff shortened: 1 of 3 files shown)


services/authenticating-proxy/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+
+services:
+  authenticating-proxy-lite:
+    <<: *authenticating-proxy
+    depends_on:
+      - mongo
+    environment:
+      MONGODB_URI: "mongodb://mongo/authenticating-proxy"
+      TEST_MONGODB_URI: "mongodb://mongo/authenticating-proxy-test"
+      GOVUK_UPSTREAM_URI: http://draft-origin.dev.gov.uk
```
ChrisBAshton: Can these be alphabetical? Or at least in the same order in the lite vs app stacks?

kevindew: Sure - I've switched them around.

I'd agree on alphabetical but it seems that shipped long sailed for the rest of these services environment variables and it'd be inconsistent with most other services to not have HOST following VIRTUAL_HOST for the app. It's a shame as I prefer them alphabetical too.

We made a more concerted effort for alphabetical in [publishing-e2e-tests](https://github.com/alphagov/publishing-e2e-tests/blob/master/docker-compose.yml) but that too failed. Maybe only an annoying linter is the way for consistency, but people may question the value of that.

PR #224, opened 2019-10-23
Add Transition

issyl0:
- I needed to run Transition locally. I didn't want to run a dev VM. The rest is here.

### DIFF
--- a/services/transition/Makefile
+++ b/services/transition/Makefile
@@ -1,0 +1,2 @@
+transition: bundle-transition
+	$(GOVUK_DOCKER) compose run $@-lite sh -c 'bin/rake db:migrate 2>/dev/null || bin/rake db:setup'
--- a/services/transition/docker-compose.yml
+++ b/services/transition/docker-compose.yml
@@ -1,0 +1,36 @@
+version: '3.7'
+
+x-transition: &transition
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: transition
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/transition
+
+services:
+  transition-lite:
+    <<: *transition
+    depends_on:
+      - postgres
+      - redis
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres/transition"
+      TEST_DATABASE_URL: "postgresql://postgres@postgres/transition-test"
+
+  transition-app: &transition-app
+    <<: *transition
+    depends_on:
+      - postgres
+      - redis
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres/transition"
+      REDIS_URL: redis://redis
+      VIRTUAL_HOST: transition.dev.gov.uk
+      HOST: 0.0.0.0
+    ports:
+      - "3000"
+    # Change to 'bin/rails --restart' when rails >= 5.2
+    command: /bin/sh -c "rm -f tmp/pids/server.pid && rails s"
--- a/services/transition/docker-compose.yml
+++ b/services/transition/docker-compose.yml
@@ -13,6 +13,7 @@ x-transition: &transition
 services:
   transition-lite:
     <<: *transition
+    privileged: true
     depends_on:
       - postgres
       - redis
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -70,7 +70,7 @@ The following apps are supported by govuk-docker to some extent.
       * JavaScript 404 errors when previewing pages, possibly [related to analytics](https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/init.js.erb#L28)
    - ✅ support
    - ✅ support-api
-   - ❌ transition
+   - ✅ transition
    - ✅ travel-advice-publisher
    - ⚠ whitehall
       * Who knows, really - several tests are failing, lots pass ;-)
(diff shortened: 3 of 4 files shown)


issyl0:
This works in govuk-docker if I also do `govuk-docker startup` in `nginx-proxy`. Apparently this is a quirk of a new service? I'm going to make this PR not draft. :-)

kevindew:
> This works in govuk-docker if I also do `govuk-docker startup` in `nginx-proxy`. Apparently this is a quirk of a new service? I'm going to make this PR not draft. :-)

You can resolve this if you add `nginx-proxy-app` to the list of dependencies.

issyl0:
The tests don't run cleanly in Docker because the app relies on `database_cleaner` gem which doesn't like remote database URLs. Either we add the config option to allow them, or we remove the gem in favour of RSpec config. The former feels like less effort. But it's out of scope of this PR either way as it's an app change. Just keeping everyone up to speed!

issyl0:
On the branch of https://github.com/alphagov/transition/pull/810, `govuk-docker be rake` runs and all the tests pass successfully.

benthorner:
> The tests don't run cleanly in Docker because the app relies on `database_cleaner` gem which doesn't like remote database URLs. Either we add the config option to allow them, or we remove the gem in favour of RSpec config. The former feels like less effort. But it's out of scope of this PR either way as it's an app change. Just keeping everyone up to speed!

Would this help? https://github.com/alphagov/govuk-docker/blob/master/services/whitehall/docker-compose.yml#L21

issyl0:
> Would this help? https://github.com/alphagov/govuk-docker/blob/master/services/whitehall/docker-compose.yml#L21

Thanks, @benthorner, good to know there's prior art. I've done this at an app level now though.

services/transition/Makefile
```diff
@@ -0,0 +1 @@
+transition: bundle-transition
```
benthorner: Can we setup the database here?

https://github.com/alphagov/govuk-docker/blob/master/services/content-publisher/Makefile#L2

services/transition/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+  working_dir: /govuk/transition
+
+services:
+  transition-lite:
+    <<: *transition
+    depends_on:
+      - nginx-proxy-app
+      - postgres
+      - redis
+      - transition-worker
```
benthorner: We shouldn't need this (or `nginx-proxy-app`) here. The dependencies for the `lite` stack should be only those necessary to get the tests passing.

https://github.com/alphagov/govuk-docker#stacks

services/transition/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+      DATABASE_URL: "postgresql://postgres@postgres/transition"
+      TEST_DATABASE_URL: "postgresql://postgres@postgres/transition-test"
+
+  transition-app: &transition-app
+    <<: *transition
+    depends_on:
+      - nginx-proxy-app
+      - postgres
+      - redis
+      - transition-worker
```
benthorner: I wonder if we need this. My (possibly extreme!) approach to this is to only add config that I can make use of. So if you're spinning up transition in your browser, and pages error because of the worker not being there (or something else), then we should add it. Otherwise, we could safely leave it out.

services/transition/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+      - "3000"
+    # Change to 'bin/rails --restart' when rails >= 5.2
+    command: /bin/sh -c "rm -f tmp/pids/server.pid && rails s"
+
+  transition-worker:
+    <<: *transition
+    depends_on:
+      - postgres
+      - redis
+      - nginx-proxy-app
```
benthorner: We shouldn't need `nginx-proxy-app` here, as it only applies to web app services.

services/transition/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: transition
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/transition
+
+services:
+  transition-lite:
```
benthorner: We should be able to get the cucumber tests passing with the `privileged` flag.

https://github.com/alphagov/govuk-docker/blob/master/services/content-publisher/docker-compose.yml#L16

services/transition/docker-compose.yml
issyl0: If we need it later then we can add it then. Good idea.

services/transition/docker-compose.yml
benthorner: At the moment, we've just got https://github.com/alphagov/govuk-developer-docs/pull/2081/files#diff-d21cd4fd2ed7b37c176270e625ad3014R258

issyl0:
🤔 Testing a fresh build and run,`make transition` works, `govuk-docker be rspec` works, `govuk-docker be cucumber` _doesn't_ entirely as it tries to find Redis at`127.0.0.1:6379`. Seems to come from [this file in the Transition repo from 2014](https://github.com/alphagov/transition/blob/master/config/redis.yml). 🤔

```
Failing Scenarios:
cucumber features/mapping_create_multiple.feature:31 # Scenario: Creating a large batch (that will be processed in the background)
cucumber features/mapping_import.feature:45 # Scenario: Successfully importing a larger batch of mappings
```

benthorner:
Thanks for checking 🥇. I think we could fix that, but it doesn't seem worth the effort right now. One thing I totally forgot, though, is to suggest updating the [compatibility doc](https://github.com/alphagov/govuk-docker/blob/master/docs/compatibility.md) for transition.

Would you mind doing that?

PR #228, opened 2019-10-24
Clarifies that command removes all containers

injms:
Edit to clarify that `govuk-docker compose rm -sv` removes all Docker containers, not just `govuk-docker` containers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ make pull
 
 ### How to: clear your Docker containers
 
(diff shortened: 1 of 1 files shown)


benthorner:
Does it? Trying myself, I can't see that. Can you give an example?

injms:
> Does it? Trying myself, I can't see that. Can you give an example?

It would seem that I can't - I'll chalk this one up to me being daft. Apologies!

PR #231, opened 2019-10-29
Run shellcheck in `make test`

barrucadu:
[shellcheck](https://www.shellcheck.net/) is a static analyser / linter for shell scripts.

We currently have one shellscript in this repo, #220 adds a bunch for data replication, and #216 replaces `govuk-docker` with another.  That feels to me like enough shell that we should be linting it.  We have some precedent for running shellcheck in CI - eg, in [govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/master/Jenkinsfile#L30).


benthorner:
Nice one @barrucadu 💯

PR #243, opened 2019-11-01
Fix shellcheck error for update-git-repo script

benthorner:
This declares the local 'out' variable before assigning it, as well
removing the trap/rm lines, since the 'rm' functionality should be
handled automatically by our use of 'mktemp'.

### DIFF
--- a/bin/update-git-repo.sh
+++ b/bin/update-git-repo.sh
@@ -20,9 +20,9 @@ truncate () {
 }
 
 catch_errors () {
-  local out=$(mktemp -t update-git.XXXXXX)
-  # shellcheck disable=SC2064
-  trap "rm -f '$out'" EXIT
+  local out
+  out=$(mktemp -t update-git.XXXXXX)
+  trap 'rm -f "$out"' EXIT
 
   if ! "$@" >"$out" 2>&1; then
     error "FAILED, with output:"


barrucadu:
> since the 'rm' functionality should be handled automatically by our use of 'mktemp'.

Is that right?  `mktemp` makes a file in whatever the temporary directory is, but I don't think it does any automated clean-up.  The file is "temporary" in that it's in a place probably mounted as volatile storage (like a ramdisk).

benthorner:
@barrucadu perhaps this is related to my understanding of how to treat tempfiles. Historically, I've viewed them as a system feature a program doesn't need to care about, from a cleanup perspective. I think you may be thinking about them in a more conceptual way i.e. "this file should be considered temporary". This article suggests there is some automated cleanup...

https://superuser.com/questions/187071/in-macos-how-often-is-tmp-deleted

...but perhaps it's good practice to do it manually as well?

barrucadu:
I guess it's not necessary---you're not going to run out of files with names matching `update-git.XXXXXX` unless you run this script a *lot* in 3 days---but I generally do clean up temporary files in the script which creates them.

benthorner:
@barrucadu that does seem like a good thing to do in general. My frustration with it was that I couldn't easily test it was necessary after fixing the linting error. Happy with it now, though.

PR #252, opened 2019-11-06
Fix incorrect '-f' flag in how-to

benthorner:
This removes an incorrect '-f' flag from one of the how-to suggestions
around re-building a service, as well as a couple of commands:

   - 'git pull' seems a little too obvious
   - 'run bash' is covered in the intro docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,16 +75,13 @@ Sometimes things go wrong or some investigation is needed. As govuk-docker is ju
 git pull
 
 # make sure the service is built OK
-make -f <service>
+make <service>
 
 # tail logs for running services
 govuk-docker logs -f
 
 # get all the running containers
 docker ps -a
-
-# get a terminal inside a service
-govuk-docker run <service>-lite bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,20 +71,14 @@ In these cases, you can swap out the URL for the relevant frontend (such as draf
 Sometimes things go wrong or some investigation is needed. As govuk-docker is just a bunch of docker config and a CLI wrapper, it's still possible to use all the standard docker commands to help fix issues and get more info e.g.
 
 ```
-# make sure govuk-docker is up-to-date
```
thomasleese: I'm not sure if this is so obvious. This is in a section about troubleshooting which (hopefully) doesn't happen very often. If people have a govuk-docker and it's working, I imagine they're unlikely to do `git pull` very often and it may not occur to them to be the first thing to do.

benthorner: Fair enough - it's back 👍

PR #253, opened 2019-11-07
Replace govuk-lint with rubocop-govuk

theseanything:
As govuk-lint is being deprecated, this commit replaces usage of govuk-lint with rubocop and rubocop-govuk.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,8 +1,6 @@
 inherit_gem:
-  govuk-lint: "configs/rubocop/all.yml"
-
-Rails:
-  Enabled: false
+  rubocop-govuk: 
+    - "config/default.yml"
 
 AllCops:
   Exclude:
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -6,6 +6,7 @@ node {
   govuk.buildProject(
     skipDeployToIntegration: true,
     rubyLintDiff: false,
+    sassLint: false,
```
benthorner: How come we need this now?

theseanything: I was thinking because scss-lint was removed as a dependency, the CI might fail. But it might be okay

theseanything: (Tho it is generally an unnecessary step in CI for this project - as there's no SASS)

theseanything: Looks like its fine to remove!

PR #257, opened 2019-11-11
Introduce a govuk-docker-version helper script

benthorner:
This introduces a new script to print version info about govuk-docker
and ties it into the main wrapper script, as a way to motivate people
to maintain the version and the associated CHANGELOG. Sometimes there
are breaking changes in this repo, for which a notification facility
would be helpful, since people are less likely to notice an update to
a CHANGELOG without an explicit call-to-action.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 replication
 stderr.log
+.version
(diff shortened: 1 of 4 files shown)


chao-xian:
Thanks for adding this Ben!

PR #259, opened 2019-11-21
Move nginx-proxy service to top-level

benthorner:
nginx-proxy has been treated like any other service since the start of
this repo, but it's actually quite different because it has no stack and
no Makefile. It's also a top-level service that's depended on by all of
the app stacks. This moves it to the top-level docker-compose.yml file
to make it more visible and simplify writing tests for (other) services.


chao-xian:
Do we have any docs on adding a service? It's not clear otherwise that you'd need to add the domain of an app to this list when someone in the future goes to add a new app/service. Be kind to your future devs!

benthorner:
@chao-xian we have [a spec](https://github.com/alphagov/govuk-docker/blob/master/spec/integration/compose/virtual_hosts_spec.rb) to ensure it gets done, and I'm currently writing the next "Learning GOV.UK Docker" tutorial that will cover nginx-proxy.

chao-xian:
@benthorner awesome

chao-xian:
@benthorner `The downside of this is having to 'shove' the 'proxy.conf' file in the 'config' directory.` if we needed to have a top level conf file for elasticsearch, would we need to do the same?

benthorner:
@chao-xian yup. I'm not a great fan of the 'config' directory because it's ambiguous what's inside. On the other hand, it doesn't seem to be causing any pain.

benthorner:
@chao-xian did you have any more comments?

PR #260, opened 2019-11-21
Add new spec to codify how we do stacks

benthorner:
This adds a new spec to ensure all services match the conventions described
in the README for stacks.


spec/integration/compose/service_stacks_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  service_names.each do |service_name|
+    it "configures #{service_name} with a lite or app stack" do
+      expect(compose_services(service_name).keys).to include(/lite/)
+        .or include(/app(-\w+)?$/)
+    end
+  end
+
+  compose_files = Dir.glob("services/**/docker-compose.yml")
+
+  compose_app_services = compose_files.flat_map do |filename|
```
ChrisBAshton: This definition is repeated in rails_debugger_spec and virtual_hosts_spec. Is it time to create a shared helper file for our tests? It would help to make this file more readable - current structure does not make very clear what is an expectation block and what is a 'setup' block.

benthorner:
@ChrisBAshton thanks for the suggestion. I did have another PR to do that, but this one is better since there's more evidence/experience to make us of in the refactoring.

benthorner:
@ChrisBAshton could you take another look at this?

spec/support/compose_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+      service_name =~ /lite$/
+    end
+  end
+
+  def self.all_services
+    @all_services ||= begin
+                        Dir.glob("services/**/docker-compose.yml")
+                          .map { |filename| YAML.load_file(filename) }
+                          .reduce({}) { |memo, config| memo.merge(config["services"]) }
+                      end
```
ChrisBAshton: Could this method not call the `self.services` method to DRY it up a bit? Something like:

```ruby
  @all_services ||= begin
    Dir.glob("services/**/docker-compose.yml")
      .map { |filename| self.services(filename) }
      .flatten
  end
```

spec/support/compose_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  def self.rails_services
+    app_services.select do |_service_name, service|
+      service["command"] =~ /rails/
+    end
+  end
+
+  def self.lite_services
```
ChrisBAshton: I'd be tempted to move this above `rails_services`, just because we seem to have 'app' and 'lite' stacks for apps, and 'rails services' seem a bit different.

spec/support/makefile_helper.rb
```diff
@@ -0,0 +1,7 @@
+module MakefileHelper
+  def self.dependencies(service_name)
+    filename = File.join("services", service_name, "Makefile")
+    rule = File.readlines(filename).first
+    rule.split(":")[1].scan(/[\w\-_]+/)
+  end
```
ChrisBAshton: Could we have a test for this helper? It's complicated enough that I think it would be worth it. Wouldn't bother for the ServicesHelper at this stage though.

spec/support/compose_helper.rb
benthorner: Oh yeah, updated 👍

spec/support/compose_helper.rb
benthorner: Indeed it is - I was wondering if it would confuse people or not. I've renamed to `rails_app_services` to help distinguish it.

spec/support/makefile_helper.rb
benthorner: Hmm...I think it would be a bit 'meta' to test the test helpers 😆. I've modified it a bit to aid understanding - does that help?

spec/support/compose_helper.rb
```diff
@@ -0,0 +1,32 @@
+module ComposeHelper
+  def self.app_services
+    @app_services ||= all_services.select do |service_name, _service|
+      service_name =~ /app(-\w+)?$/
+    end
+  end
+
+  def self.rails_app_services
```
ChrisBAshton: This is named better 👍 Thanks

PR #261, opened 2019-11-22
Stop ignoring mapit in one of the specs

benthorner:


spec/integration/compose/rails_debugger_spec.rb
```diff
@@ -2,7 +2,6 @@
 
 RSpec.describe "Compose rails debugger" do
```
issyl0: MapIt is not a Rails app, so this description is incorrect if we include MapIt again. I think the suspicion that this was a "Rails-only test" was the reason for the ignoring.

benthorner:
Ah, my bad @issyl0. I even called the file `rails_debugger_spec`! I've fixed to it be more specific, instead of ignoring specific services.

PR #263, opened 2019-11-25
Make it possible to download data without importing it

barrucadu:
We want to remove the old replication scripts, as they don't work with gds-cli (https://github.com/alphagov/govuk-puppet/pull/9866).  The dev VM is still around, so this means we'd ideally have a (possibly painful, manual) process to import data even so. 

This PR makes it possible to download data without importing it to govuk-docker databases, and will be step 1 of importing data to the dev VM.  Step 2 will be something like "look at what the import part of the script does and do that manually".


barrucadu:
```
/Users/michaelswalker/govuk/govuk-docker/bin/replicate-mongodb.sh: line 53: SKIP_IMPORT: unbound variable
```

Whoops...

PR #267, opened 2019-11-27
Add basic lite stack for local-links-manager

benthorner:
To help review: https://github.com/alphagov/local-links-manager/pull/538

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -46,7 +46,8 @@ The following apps are supported by govuk-docker to some extent.
    - ⚠ link-checker-api
       * Works in isolation but not in other services' `e2e` stacks, so must be run in a separate process.
         See https://github.com/alphagov/govuk-docker/issues/174 for details.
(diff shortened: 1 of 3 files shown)


docs/compatibility.md
```diff
@@ -46,7 +46,8 @@ The following apps are supported by govuk-docker to some extent.
    - ⚠ link-checker-api
       * Works in isolation but not in other services' `e2e` stacks, so must be run in a separate process.
         See https://github.com/alphagov/govuk-docker/issues/174 for details.
-   - ❌ local-links-manager
+   - ⚠  local-links-manager
+      * Only the `lite` stack is available, for running tests.
```
issyl0: Aside: Is it worth documenting somewhere why the things that aren't implemented aren't implemented? What things people have run into while adding them? Or is it the case that people just haven't yet?

benthorner: Fair point - I've updated the bullet. It's just because I have no reason to :-).

PR #268, opened 2019-11-27
Fix postgres database replication

barrucadu:
At some point the filename pattern changed: backups are now in the
"postgresql-backend" directory, and it's not consistent at all whether
things use hyphens or underscores.

I decided to use `aws s3 ls ... | grep` to find the app, rather than
hard code a list of app name -> backup file name mappings.  The list
would be long, and the `aws s3 ls` is needed anyway, as the backup
timestamps also now include the time the backup was taken, not just
the date.

### DIFF
--- a/bin/replicate-postgresql.sh
+++ b/bin/replicate-postgresql.sh
@@ -68,4 +68,4 @@ done
 database="$app"
 govuk-docker run postgres /usr/bin/psql -h postgres -U postgres -c "DROP DATABASE IF EXISTS \"${database}\""
 govuk-docker run postgres /usr/bin/createdb -h postgres -U postgres "$database"
-pv "$archive_path"  | gunzip | grep -v 'ALTER \(.*\) OWNER TO \(.*\);' | govuk-docker run postgres /usr/bin/psql -h postgres -U postgres -qAt -d "$database"
+pv "$archive_path" | govuk-docker run postgres /usr/bin/pg_restore -h postgres -U postgres -d "$database" --no-owner
--- a/bin/replicate-postgresql.sh
+++ b/bin/replicate-postgresql.sh
@@ -58,7 +58,7 @@ echo "stopping running govuk-docker containers..."
 govuk-docker down
 
 govuk-docker up -d postgres
-trap 'govuk-docker compose stop postgres' EXIT
+trap 'govuk-docker stop postgres' EXIT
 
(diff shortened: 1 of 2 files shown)


bin/replicate-postgresql.sh
```diff
@@ -31,7 +38,15 @@ if [[ -e "$archive_path" ]]; then
   echo "Skipping download - remove ${archive_path} to force"
 else
   mkdir -p "$archive_dir"
-  aws --profile govuk-integration s3 cp "s3://${bucket}/postgres/$(date '+%Y-%m-%d')/${archive_file}" "${archive_path}"
+  s3_file=$(try_find_file "$app")
+  if [[ -z "$s3_file" ]]; then
+    s3_file=$(try_find_file "${app//-/_}")
+  fi
+  if [[ -z "$s3_file" ]]; then
+    echo "couldn't figure out backup filename in S3 - this is a bug (or the app doesn't use postgres)."
```
issyl0:
```suggestion
    echo "couldn't figure out backup filename in S3 - if you're sure the app uses PostgreSQL, file an issue in alphagov/govuk-docker."
```

bin/replicate-postgresql.sh
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+function try_find_file {
+  set +e
+  out="$(aws --profile govuk-integration s3 ls "s3://${bucket}/postgresql-backend/" | grep "${1}_production.gz" | sed 's/.* //' | sort | tail -n1)"
```
issyl0: How does `aws --profile integration` interact with the [new local set up for accessing AWS](https://docs.publishing.service.gov.uk/manual/gds-cli.html)? Or is this separate? I guess I should try out running the scripts. :wink:

barrucadu: It probably won't interact - these scripts probably need to be `gds-cli`'d at some point

issyl0: Yeah - oops - my GDS CLI tutorial told people to delete their `~/.aws/` config files. 😬I'll test these out and do a separate PR for any updates based on `gds-cli` stuff. This is good to merge as-is though, I don't want this to block anything as I doubt everyone would have _actually_ deleted stuff, and this is Important. Also I don't know when I'll get to testing these out as it feels Large.

PR #269, opened 2019-11-27
Rewrite setup script to be a single file

benthorner:
Previously we had a Ruby script to setup GOV.UK Docker, which was hard to
understand because the setup spanned multiple files and methods.

This rewrites the setup into a single script. A couple of features of the
original script haven't been carried over:

  - Since all the changes are necessary and isolated, we don't ask before
doing them
 - We only expect people to run setup once in the repo, so no more
repo/pull stuff

This also adds an extra command to ensure the system resolver picks up the
new config in /etc/resolver, which has been a problem for some.

Supersedes and closes https://github.com/alphagov/govuk-docker/pull/160.


docs/installation.md
```diff
@@ -121,91 +121,46 @@ gem install bundler:X.Y.Z
...
-sudo brew services restart dnsmasq
-```
-
-To check whether dnsmasq name server at 127.0.0.1 can resolve subdomains of dev.gov.uk run `dig app.dev.gov.uk @127.0.0.1`. The response has to include the following answer section:
-
-```
-;; ANSWER SECTION:
-app.dev.gov.uk.		0	IN	A	127.0.0.1
-```
+You can also look at the command in the `bin/setup` to see what's changing.
```
issyl0:
```suggestion
You can also look at the command in `bin/setup` to see what's changing.
```

benthorner: Fixed 👍

benthorner:
@issyl0 so it does - I've updated the description to reflect that. We went from having no setup script at all to something quite sophisticated, and I'd like to see if we can get by with just this. Objectively, it's also quicker to setup GOV.UK Docker because that no longer requires rbenv+bundler.

Sorry, your tests were a necessary casualty :-).

benthorner:
I've also added an extra commit to fix a test error that popped up late-on.

issyl0:
I added "and closes", as "close" or "fix" is the keyword for auto-closing PRs/issues with the merging of other PRs.

issyl0:
You missed one:

```
╭─isabelllong@gds5567 ~/govuk/govuk-docker ‹rewrite-setup-command›
╰─ $ ./bin/doctor
Checking govuk-docker
Traceback (most recent call last):
	2: from ./bin/doctor:12:in `<main>'
	1: from /Users/isabelllong/govuk/govuk-docker/lib/govuk_docker/doctor/doctor.rb:5:in `messages'
/Users/isabelllong/govuk/govuk-docker/lib/govuk_docker/doctor/doctor.rb:71:in `dnsmasq_messages': undefined method `dnsmasq_conf' for GovukDocker::Paths:Class (NoMethodError)
```

benthorner:
Thanks @issyl0!

PR #271, opened 2019-11-28
More tweaks to installation docs

benthorner:

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -62,7 +62,7 @@ alias gdu="govuk-docker-up"
 alias gdbe="govuk-docker-run bundle exec"
 ```
 
-## Environment variables
+### Environment variables
 
 Both govuk-docker and the Makefile respect the following environment variables:
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -4,15 +4,15 @@
 
 govuk-docker has the following dependencies:
 
-- [brew](https://brew.sh/). (If you don't use a Mac, you'll need to dig into the `govuk-docker setup` command and manually install the things referenced).
+- [brew](https://brew.sh/)
 - [git](https://git-scm.com)
-- Ruby (whatever version is specified in [.ruby-version](https://github.com/alphagov/govuk-docker/blob/master/.ruby-version))
-  - Follow these [instructions to update Ruby using brew](#how-to-install-or-update-ruby)
+- [Ruby](#how-to-install-or-update-ruby)
 - A directory `~/govuk` in your home directory
 
 All other dependencies will be installed for you automatically.
 
 ### Docker settings
+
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -41,6 +31,14 @@ bin/setup
...
 You can now [clone and setup the apps you need](../README.md#Usage), after which you can do things like run tests and startup the app in your browser. If this doesn't work for whatever reason, follow the [instructions on how to resolve setup issues](#how-tos).
 
+### Docker settings
+
+Running GOV.UK applications can be resource intensive. To give Docker more resources, click the Docker whale icon in the macOS menu bar, select 'Preferences'. We suggest the following minimum resources:
+
+* 6 CPUs
+* 12 GB RAM
+* 64GB+ Disk
+
```
issyl0: Is it worth noting that by default if you run Docker on Linux, this doesn't apply as it can use all the available resources?

docs/installation.md
```diff
@@ -4,24 +4,14 @@
 
 govuk-docker has the following dependencies:
 
-- [brew](https://brew.sh/). (If you don't use a Mac, you'll need to dig into the `govuk-docker setup` command and manually install the things referenced).
+- [brew](https://brew.sh/)
+  - If you don't use a Mac, see the [Brewfile](../Brewfile)
```
issyl0: Homebrew also works on Linux and last time I tried, this Brewfile _does_ work apart from the Docker `cask` (I supressed the urge to include a `brew "docker" if OS.linux?`). But Linux users will have other problems (notably being constrained to the BYOD VPN). So it's best if we don't?

docs/installation.md
benthorner: Sounds reasonable - I've added a caveat.

docs/installation.md
benthorner: Hmm...I'm not sure. I don't think we should make the instructions more complicated, but they should also not be confusing for anyone who is using Linux. This seems like a reasonable compromise, since they are still motivated to ensure Docker is installed.

issyl0: There are other changes to make govuk-docker work for Linux, including dnsmasq and resolver configs. I have a branch somewhere from a few months ago. Everything's changed since then, and I stopped working on it when I realised that the BYOD VPN scuppered my use of it. I'm happy to revisit them if there's a user need, but for now I think there's not until Linux laptops become more usual in GOV.UK (which doesn't seem to be a Thing that we or IT are pushing for).

benthorner: Ah yes, I forgot about that. I've updated the commit/message to remove the Mac caveat, pending a proper set of instructions. Seems reasonable to leave the Docker settings bit, since it's separate and specific to Macs.

PR #273, opened 2019-11-29
Brewfile: Install Docker when on Linux too

issyl0:
- There was some discussion in #271 about making the installation
  instructions better for Linux.
- The Brewfile works on Linux, but for completeness and parity between
  the systems, let's install Docker when on Linux too (Casks are
  auto-skipped).
- There is still a long way to go to make GOV.UK Docker work on Linux,
  notably the GDS BYOD VPN and the dnsmasq configuration, but we can at
  least have everything installed for if ever Linux users want to figure
  it out.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,8 +1,11 @@
-tap "homebrew/cask"
-
 brew "dnsmasq"
-brew "docker-compose"
 brew "pv"
 brew "shellcheck"
 
-cask "docker"
+if OS.mac?
+  tap "homebrew/cask"
+  cask "docker"
+else
+  brew "docker"
+  brew "docker-compose"
+end


Brewfile
```diff
@@ -1,6 +1,7 @@
 tap "homebrew/cask"
 
 brew "dnsmasq"
+brew "docker" if OS.linux?
```
benthorner: Should this be `unless`?

issyl0: `unless OS.mac?` or `if OS.linux?` do the same thing for installing Linux-only dependencies.

On macOS, Docker is a cask (line 9).

benthorner: Hmm...still confused, sorry. We want to install docker via Brew if we're on a Mac? Or if we're on Linux?

issyl0: Linux (currently):

```
root@15a3d641e1df:/home/linuxbrew# brew bundle
Tapping homebrew/cask
Installing dnsmasq
Installing docker-compose
Installing pv
Installing shellcheck
Skipping cask docker (on Linux)
Homebrew Bundle complete! 5 Brewfile dependencies now installed.

root@15a3d641e1df:/home/linuxbrew# docker --version
bash: docker: command not found
```

Linux (with this change):

```
root@15a3d641e1df:/home/linuxbrew# brew bundle
Using homebrew/cask
Installing dnsmasq
Installing docker
Installing docker-compose
Installing pv
Installing shellcheck
Skipping cask docker (on Linux)
Homebrew Bundle complete! 6 Brewfile dependencies now installed.

root@15a3d641e1df:/home/linuxbrew# docker --version
Docker version 19.03.5, build 633a0ea
```

Does that help?

benthorner: Err, now I'm more confused :-|. My current understanding is:

   - `brew "docker"` blindly installs the `docker` package
   - `brew "docker" if OS.linux?` restricts installing Docker to Linux

But we also want to install Docker on Macs? I know some people have installed the desktop version, but in general I think we assume they won't have it. We don't tell people to install Docker any other way in the docs.

issyl0: Ah, I see the confusion! Docker is already installed on Macs with line 9, `cask "docker"`, with which people get the Docker for Mac GUI and the CLI tool. Hence why we're only installing the CLI separately on Linux, because the two packages would conflict if both installed on macOS.

benthorner: That makes more sense - we probably need a comment! Looks like `docker-compose` is also covered by the cask. After trying from scratch, I've also put up https://github.com/alphagov/govuk-docker/pull/274.

issyl0: I pushed a change that moves the section around so that it's separate. Does that work better?

benthorner: Looks good. By way of documentation, perhaps we could put the `cask` line in a `if OS.mac?` block. If that's really common brew knowledge, then we could just leave it as-is 👍.

issyl0: Casks are skipped by default on Linux, so the `if OS.mac?` would be a no-op.

benthorner: Indeed. However, I didn't know that before this PR, and maybe it would confuse others too. If it's common knowledge, then let's keep as-is 👍.

Another idea: can we do `if ... else`?

issyl0: How's that?

PR #275, opened 2019-12-02
Fix Rails 6 warnings about BINDING env var

benthorner:


benthorner:
@kevindew I forgot about that! Unfortunately we don't need my change anymore, since we're using the `--restart` flag now 😬

PR #278, opened 2019-12-03
Replace overloaded 'service' term with 'project'

benthorner:
Previously we used the word 'service' to describe the combination of
apps and libraries that make up GOV.UK. Unfortunately this is an
overloaded term: Docker Compose also uses the term 'service' to describe
something you can start/run. In this terminology, each GOV.UK service
had many Docker Compose services for each of its stacks. Confusing!

This renames all uses of the term '(GOV.UK) service' with 'project' as
a way to distinguish what we are referring to. I couldn't think of a
better word to describe 'apps + libraries' in a way that makes sense
when 'running' them, 'building' them, 'starting' them, etc.


kevindew:
@benthorner just noticed this has broken some make commands because we refer to images in services directory: https://github.com/alphagov/govuk-docker/blob/master/projects/asset-manager/docker-compose.yml#L6

benthorner:
@kevindew oops! Fix: https://github.com/alphagov/govuk-docker/pull/280

PR #282, opened 2019-12-09
Install rbenv using brew on setup

brucebolt:
This was an additional step we were expected to do manually, so adding rbenv to the Brewfile so it gets installed when `bin/setup` is run.


issyl0:
I pushed your branch to `origin` so that the tests run. 🙄

PR #283, opened 2019-12-09
Create the /etc/resolver directory

brucebolt:
`bin/setup` was writing a file to this directory, but it does not exist so the step was failing so the DNS was not correctly configured.


benthorner:
Dah, but I'm not in the org so I can't approve :-(.

issyl0:
I pushed this branch to `origin` so that the tests run. 🙄

PR #289, opened 2019-12-31
Fix phantomjs not working

benthorner:
This has been causing apps like finder-frontend to fail when running                                                                                                                                                
'rake' or 'rake spec:javascript' since it still uses phantomjs as the 
headless browser, and phantomjs stopped working.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=918727

Similar issue: https://github.com/grafana/grafana/issues/17588. 

Looks like the principle way to fix this is to replace PhantomJS!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+## 1.0.0
+
+* Signify that GOV.UK Docker is stable and ready for widespread use.
+
+## 0.0.2
+
+* Fix issue with Jasmine tests using PhantomJS (issue with OpenSSL)
+
+You’ll need to run 'govuk-docker build <project>-lite' to apply the fix.
+
 ## 0.0.1
 
 * Introduce versioning feature to support people when things change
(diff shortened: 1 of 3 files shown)


kevindew:
This looks good to me, but I'm not too clear what we're actually fixing in govuk-docker with this change? might be good to have that documented in this PR.

benthorner:
@kevindew sure, I've updated the commit/PR. Does that help?

issyl0:
Can we migrate away from PhantomJS in the affected apps?

benthorner:
> Can we migrate away from PhantomJS in the affected apps?

@issyl0 ideally, yes. However, I think that's beyond the scope of this PR. As a compromise, I've pulled back a (not-so-much) 'paid down' tech debt card about this...

https://trello.com/c/M7KeYcSp/83-phantomjs-development-has-been-abandoned

benthorner:
@ChrisBAshton updated with a CHANGELOG alert - would you like to 👀

benthorner:
@ChrisBAshton sure. I didn't realise `1.0.0` signified the 'stable/public' version. To avoid spamming people unduly, I'll increment it here - updated.

PR #296, opened 2020-01-21
Expose postgres port

bevanloon:
Allow connections from the host machine from 0.0.0.0:5432 to aid in
running the knowledge graph locally.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,8 @@ services:
     image: postgres:9.6
     volumes:
       - postgres:/var/lib/postgresql/data
+    ports:
+      - 5432:5432
 
   memcached:
     image: memcached


bevanloon:
So to run the knowledge graph I need both the content store and the publishing api exposed at the moment. I'm using `docker-compose up mongo postgres`. Can I expose port for postgres in that command? or an equivalent that brings both services up? (`run` doesn't allow more than one service at a time from what I can see).

Obviously it is a trade off - what's easiest/best for most people and what is the supported way of developing on GOV.UK. Do we have a large community of people who are running a postgres instance on their host machine as well?

I wonder whether there's something here around the "accepted" GOV.UK way - i.e. is it standard to have a postgres on the host and in docker or is it the standard to only have postgres in docker. Have we put any thought into that at all? 

I worry that if we try to cater to all setups, we end up catering to none of them very well.

benthorner:
Hmmm....tricky one. It's probably worth asking whether the need to run stuff on a local machine could be translated into a Docker container i.e. so the whole thing is run within Docker. I don't know what kind of programme/script/thing you're working with. Could it even be a service in its own right?

I agree we should avoid being overcautious. My concern is that it increases the number of potential issues people may run into, especially with apps like DGU Publish, which does require a local Postgres DB for the time being. In contrast, I'm not aware of anyone running a local Mongo for anything. A compromise might be to choose an alternative port that wouldn't clash with the default for Postgres - not ideal, but then if only a handful of people use it, I don't think that would be an issue.

benthorner:
Forgot to mention: I don't think there's any other way, apart from using a couple of `run` commands with the right flags.

bevanloon:
Closing. Did a small poll on slack - some developers do run a postgres instance on their host machine as well. Memorable quote from the poll "would ruin my life".

PR #297, opened 2020-01-22
Fix a couple of issues with initial setup

benthorner:


benthorner:
> LGTM 👍 I had to look up the `dirname $0` syntax, perhaps consider rephrasing the second commit message to be more explicit about what it is doing.

Sure - it's pretty standard for a bash script, but then it is bash! I've added a little to the commit.

PR #300, opened 2020-01-27
Whitehall specific Dockerfile

kevindew:
This creates a Dockerfile that is used for Whitehall builds. It adds the
ghostscript and mysql-client dependencies. Whitehall uses ghostscript as
a means to create image files from PDFs and it calls out to the MySQL
client during tests to test a sanitisation script.

These seem to be the missing steps needed to get a fully passing
Whitehall test suite passing with GOV.UK docker.

### DIFF
--- a/projects/whitehall/Dockerfile
+++ b/projects/whitehall/Dockerfile
@@ -1,0 +1,38 @@
+# Install packages for building ruby
+FROM buildpack-deps
+
+# Install chrome and its dependencies
+RUN apt-get update -qq && apt-get install -y libxss1 libappindicator1 libindicator7
+RUN wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb 2>&1 && \
+   apt install -y ./google-chrome*.deb && \
+    rm ./google-chrome*.deb
+
+# Install node / yarn
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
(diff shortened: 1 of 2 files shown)


projects/whitehall/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Install node / yarn
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update -qq && apt-get install -y yarn nodejs
+
+# Install phantomjs for some projects
+RUN export VERSION=2.1.1 && \
+    wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-$VERSION-linux-x86_64.tar.bz2 2>&1 && \
```
issyl0: PhantomJS is [very ☠️ ⚰️ as of 2018](https://github.com/ariya/phantomjs/issues/15344) - can we move away from it?

kevindew: Yup - apps are gradually losing it. In Whitehall it seems to be still part of the JS testing. Lots of things to improve still in Whitehall so not going to scratch that itch right now.

PR #304, opened 2020-02-05
Add Maslow to GOV.UK Docker

issyl0:
- Platform Health are upgrading this app to Rails 6, so I felt it was a
  good idea to make it runnable with GOV.UK Docker like the rest of the
  apps.

https://trello.com/c/1TKgm8xV/1743-add-maslow-to-govuk-docker

### DIFF
--- a/projects/maslow/docker-compose.yml
+++ b/projects/maslow/docker-compose.yml
@@ -1,0 +1,37 @@
+version: '3.7'
+
+x-maslow: &maslow
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: maslow
+  stdin_open: true
+  tty: true
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/maslow
+
+services:
+  maslow-lite:
+    <<: *maslow
+    privileged: true
+    depends_on:
+      - mongo
+    environment:
+      MONGODB_URI: "mongodb://mongo/maslow"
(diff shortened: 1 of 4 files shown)


issyl0:
I tested this with `govuk-docker run maslow-lite bundle exec rake` and all the tests passed. Running this in a browser, I get a GDS SSO error message, but that's to be expected as I've not replicated data yet so there are no users.

projects/maslow/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      - publishing-api-app
+    environment:
+      MONGODB_URI: "mongodb://mongo/maslow"
+      VIRTUAL_HOST: maslow.dev.gov.uk
+      HOST: 0.0.0.0
+    expose:
+      - "3000"
+    command: bin/rails s --restart
+
+  maslow-app-e2e:
```
benthorner: Does this need an `e2e` stack? We're trying to avoid them, as they're low usage and resource hungry.

issyl0: I don't think so - I copied the files from Publisher.

issyl0: Thanks. Removed.

benthorner:
With the web app, I think we should be using [the Content Store approach](https://github.com/alphagov/govuk-docker/blob/master/projects/content-store/Makefile#L3), since the DB setup needs to be a little different for Mongo - this should mean a 'dev user' gets created.

issyl0:
@benthorner Also done! That makes sense.

PR #309, opened 2020-02-24
Update Mapit setup

deborahchua:
This is part of the work to update Mapit to use Docker rather than the VM, specifically
to import data.

Trello card: https://trello.com/c/mJVTfjjd/1719-8-import-mapit-data-using-docker-%F0%9F%8D%90

### DIFF
--- a/projects/mapit/Dockerfile
+++ b/projects/mapit/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:2.7-buster
 
 RUN apt-get update -qq && apt-get install -y gettext \
-        postgresql-client libgdal-dev libgeos-dev musl-dev
+        postgresql-client libgdal-dev libgeos-dev musl-dev ruby-dev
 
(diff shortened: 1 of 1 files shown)


projects/mapit/Makefile
```diff
@@ -1,4 +1,5 @@
 mapit:
+	- $(GOVUK_DOCKER) run $@-app psql -U postgres -c "DROP DATABASE mapit;"
```
benthorner: We should avoid doing things like this, as these steps are meant to be idempotent: we often suggest people [run `make <project>` for troubleshooting](https://github.com/alphagov/govuk-docker#how-to-diagnose-and-troubleshoot), and it could be frustrating if that deletes dev data.

Since the purpose of this is for importing data, could the import scripts do this instead?

deborahchua: Ah I see, thanks @benthorner ! I have removed the commit now 👍

PR #310, opened 2020-02-28
Increase nginx proxy_read_timeout

chao-xian:
When assets are compiling in frontend apps, we tend to see 504s as
the process takes so long. Rather than hitting refresh again and
again, increasing the timeout should cut down waiting times.

### DIFF
--- a/nginx-proxy.conf
+++ b/nginx-proxy.conf
@@ -17,3 +17,5 @@ add_header 'Access-Control-Allow-Origin' '*';
 
 # Required for uploading files on various apps
 client_max_body_size 500m;
+
+proxy_read_timeout 200s;


nginx-proxy.conf
```diff
@@ -17,3 +17,5 @@ add_header 'Access-Control-Allow-Origin' '*';
 
 # Required for uploading files on various apps
 client_max_body_size 500m;
+
+proxy_read_timeout 200s;
```
issyl0: Is this the amount of time that compiling assets takes? 😱

chao-xian: The default is 60s and it times out. I followed a scientific formula of adding a zero to the end and not seeing 504s at the moment. Which is a huge relief.

benthorner: Sorry for the delayed comment. Could we add a comment to indicate the reason behind this setting, like the other lines in this file? Although it's in the commit, that information become less visible if someone decides to reorder the lines in this file; and it's just helpful to read it in-place, as the meaning isn't going to change.

PR #314, opened 2020-03-09
Fix auth for new Postgres 9.6 image

benthorner:
This forces the Postgres 9.6 service to use 'trust' authentication,
    to avoid having to specify a password everywhere. It seems the new
    Postgres 9.6 image no longer provides open access out-of-the-box.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,8 @@ volumes:
 services:
   postgres:
     image: postgres:9.6
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
     volumes:
       - postgres:/var/lib/postgresql/data
 


docker-compose.yml
```diff
@@ -12,6 +12,8 @@ volumes:
 services:
   postgres:
     image: postgres:9.6
+    environment:
+      POSTGRES_PASSWORD: notasecret
```
barrucadu: We also have to update every app's postgres URL to include this, see https://github.com/alphagov/publishing-e2e-tests/commit/bd8a6b897a3b3cb3d5345028e4f6adf8f1f77cfa

benthorner: Oops - lack of thinking there! I've updated with something (hopefully) better.

PR #316, opened 2020-03-10
Set mongodb to use smallfiles

chao-xian:
Seeing persistent errors that there's not enough space. This is a
[common issue in MongoDB prior to 3.0](https://github.com/dockerfile/mongodb/issues/9). Setting the --smallfiles
flag addressed the problem.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
     ports:
       - "27017:27017"
       - "28017:28017"
+    command: --smallfiles
 
   mysql:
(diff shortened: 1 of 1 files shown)


benthorner:
Could we give more detail about the problem we were seeing? I think it's somewhere in the linked issue, and it would be helpful to have our specific experience of it.

chao-xian:
I never got to the bottom of why I was experiencing this and it's not re-occurred since I switched back to master.

PR #320, opened 2020-03-16
Add Contacts Admin app to GOV.UK Docker

issyl0:
- All I wanted to do was run `bundle exec rubocop` to fix a bunch of linting errors so that RE can merge one of their PRs.
- Now, contacts-admin is in GOV.UK Docker! The app uses a MySQL database, which I don't have on my local machine, so `bundle install` was failing and it felt easier to do Docker for the benefit of everyone else who might ever need to develop this app.
- The Yak is taking longer to shave than I expected.

----

This is draft as it's not working yet (database woes), but I wanted to raise awareness.

### DIFF
--- a/projects/contacts-admin/docker-compose.yml
+++ b/projects/contacts-admin/docker-compose.yml
@@ -1,0 +1,43 @@
+version: '3.7'
+
+x-contacts-admin: &contacts-admin
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: contacts-admin
+  stdin_open: true
+  tty: true
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - home:/home/build
+  working_dir: /govuk/contacts-admin
+
+services:
+  contacts-admin-lite:
+    <<: *contacts-admin
+    privileged: true
+    depends_on:
+      - mysql
+      - redis
+    environment:
+      DATABASE_URL: "mysql2://root:root@mysql/contacts_development"
+      TEST_DATABASE_URL: "mysql2://root:root@mysql/contacts_test"
+      REDIS_HOST: redis
+
+  contacts-admin-app:
+    <<: *contacts-admin
+    depends_on:
+      - collections
+      - mysql
+      - nginx-proxy
+      - redis
+      - static
+      - whitehall
+    environment:
+      DATABASE_URL: "mysql2://root:root@mysql/contacts_development"
+      VIRTUAL_HOST: contacts-admin.dev.gov.uk
+      HOST: 0.0.0.0
+      REDIS_HOST: redis
+    expose:
+      - "3000"
+    command: bin/rails s --restart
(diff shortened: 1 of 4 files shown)


issyl0:
```
/Users/issylong/govuk/govuk-docker/exe/govuk-docker run contacts-admin-lite sh -c 'bin/rake db:migrate 2>/dev/null || bin/rake db:setup'
docker-compose -f [...] run contacts-admin-lite sh -c bin/rake db:migrate 2>/dev/null || bin/rake db:setup
Starting govuk-docker_mysql_1 ... done
Starting govuk-docker_redis_1 ... done
Database 'contacts_development' already exists
Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
Couldn't create 'contacts_test' database. Please check your configuration.
rake aborted!
Mysql2::Error::ConnectionError: Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
```

🤔

benthorner:
Looks like the DB config is trying to use the local machine. Probably missing an env var, or the config is too hard coded.

issyl0:
Signon is another app that uses a MySQL database, so I copied that and crossed my fingers. :-)

benthorner:
@issyl0 sorry, I was rushing my comment. I'll explain a bit more here. We've had some issues where the DB config for an app assumes the DB is available on the local machine, for certain environments. That looks to be the case here: the dev environment allows for a customisable URL, but the test environment is just defaulting to localhost. See [here](https://github.com/alphagov/contacts-admin/blob/master/config/database.yml#L20).

issyl0:
It works!

projects/contacts-admin/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      - redis
+    environment:
+      DATABASE_URL: "mysql2://root:root@mysql/contacts_development"
+      TEST_DATABASE_URL: "mysql2://root:root@mysql/contacts_test"
+      REDIS_HOST: redis
+
+  contacts-admin-app:
+    <<: *contacts-admin
+    depends_on:
+      - mysql
```
benthorner: I think we're missing some dependencies here, in comparison to the Makefile. For example, there's no point `make`ing Whitehall if we don't fire it up anywhere. If it's too inconvenient to test the `-app` stack, we don't have to add it here - someone else can do that when they need to. We could make a note of that in the compatibility list.

issyl0: I'm just going by what I know is still relevant from [the contacts-admin README](https://github.com/alphagov/contacts-admin#dependencies). I last touched this app in 2016. 😬 The thing I was trying to do on Friday is done now anyway thanks to this working `-lite` stack: https://github.com/alphagov/contacts-admin/pull/731. 😅

issyl0:
I don't have time to work on this right now. If people want it (ie when Platform Health come to do a Rails upgrade), feel free to re-open. I'll leave the branch around.

PR #335, opened 2020-04-24
Add data.gov.uk Publish ('lite' stack only)

benthorner:
"The time has come."

Depends on: https://github.com/alphagov/datagovuk_publish/pull/957


issyl0:
I thought that Publish wasn't used...

benthorner:
It is used: it's the conduit through which stuff published through CKAN makes it into the ES DB for the frontend, which is called "DGU Find". Simples.

(It is also an incomplete web app, which is unused, but was intended to replace CKAN.)

PR #336, opened 2020-04-26
Use GitHub Actions for CI, not Jenkins

issyl0:
- We don't need to use Jenkins here as govuk-docker is not a deployed
  app on GOV.UK infrastructure.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,13 @@
+name: govuk-docker
+
+on: [push, pull_request]
```
benthorner, line 3: Do we just need `push` here? Otherwise, I think this will run the tests twice, every time someone pushes.

issyl0:
```suggestion
on: push
```

PR #349, opened 2020-05-15
ci: Adhere to the styleguide

issyl0:


issyl0:
Huh, just noticed that this PR came from my fork. Presumably when I was testing Actions out for forks for the RFC, I forgot to delete it. And `gh pr create` pushed to the last known-good remote?

PR #350, opened 2020-05-15
Add a Dockerfile linter (Hadolint) to this repo

issyl0:
- [Hadolint](https://github.com/hadolint/hadolint) is a Dockerfile linter that also bundles `shellcheck`.
- I've seen it used in other open source projects with Dockerfiles, and it was mentioned as a thing in one of the GOV.UK replatforming ideas docs. So I thought I'd see how it performed here.

----

So far I've only linted `Dockerfile.govuk-base`. But if people like this (that's why it's a draft - comments welcome) then I'll extend CI to lint everything and fix all the failures. The other ones are pretty much the same as the `Dockerfile.govuk-base` ones as they're copies with extra stuff per-app.

----

I've had to globally ignore some lints here:

- [DL3008](https://github.com/hadolint/hadolint/wiki/DL3008) - pinning versions in `apt-get install` - feels like quite a lot of maintenance overhead for every patch version that we'd need to update in Dockerfile.govuk-base and all its downstream consumers. I'm going to wager that we don't want the dependency management of having to check each of these Dockerfiles periodically for the latest version numbers of packages and have to update them.
  - However, I can see that for some, this would be useful. We've had several times recently in local dev where `docker build` has failed because of updated dependency versions that other things we use can't cope with.
  - If someone wants to volunteer to keep all of these up-to-date - and, indeed, find a known-good state in the first place - feel free to suggest to unignore this one!

- [DL3006](https://github.com/hadolint/hadolint/wiki/DL3006) is a little less controversial. We don't tag versions of these Dockerfiles, so we can't satisfy this lint.


Dockerfile.govuk-base
```diff
@@ -6,7 +6,8 @@
 FROM buildpack-deps
 
 # Install chrome and its dependencies
-RUN apt-get update -qq && apt-get install -y libxss1 libappindicator1 libindicator7 --no-install-recommends
+RUN apt-get update -qq && apt-get install -y libxss1 libappindicator1 libindicator7 --no-install-recommends && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
```
cbaines, line 10: Are you sure that `apt-get clean` is necessary given `/etc/apt/apt.conf.d/docker-clean`?

Dockerfile.govuk-base
```diff
@@ -6,7 +6,7 @@
 FROM buildpack-deps
 
 # Install chrome and its dependencies
-RUN apt-get update -qq && apt-get install -y libxss1 libappindicator1 libindicator7
+RUN apt-get update -qq && apt-get install -y libxss1 libappindicator1 libindicator7 --no-install-recommends
```
cbaines: I know Docker favours using difficult and unmaintainable Dockerfiles, but if you wanted to avoid passing `--no-install-recommends` each time you invoke apt, I believe it's possible to effect the same change by amending the apt configuration.

Dockerfile.govuk-base
issyl0: :shrug: I just did what the robot told me to do.

cbaines: I think that says everything about the danger to thinking of "linters" as a set of tools.

issyl0: Yeah, I admit I didn't really weigh up pros and cons.

issyl0:
In light of the above thread of "why does this need to happen", maybe it doesn't.

PR #367, opened 2020-06-18
Better describe Whitehall compatability

cbaines:
Remove the uncertianty, I've heard from people that say most tests
pass, but some fail consistently, and this will be down to an issue
with the environment provided by govuk-docker (given I know they all
pass on the ci-agents, and for me locally when not using
govuk-docker).

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -82,6 +82,6 @@ The following apps are supported by govuk-docker to some extent.
    - ✅ transition
    - ✅ travel-advice-publisher
    - ⚠ whitehall
-      * Who knows, really - several tests are failing, lots pass ;-)
+      * Most tests pass, but several tests fail consistently due to issues with govuk-docker
       * Rake task to [create a test taxon](https://github.com/alphagov/whitehall/blob/master/lib/tasks/taxonomy.rake#L11) for publishing is not idempotent
       * Placeholder images don't work as missing proxy for [/government/assets](https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/news_article_presenter.rb#L133)


docs/compatibility.md
```diff
@@ -82,6 +82,6 @@ The following apps are supported by govuk-docker to some extent.
    - ✅ transition
    - ✅ travel-advice-publisher
    - ⚠ whitehall
-      * Who knows, really - several tests are failing, lots pass ;-)
+      * Most tests pass, but several tests fail consistently due to issues with govuk-docker
```
benthorner, line 85:
```suggestion
      * Most tests pass, but several tests fail consistently and we haven't investigated why yet
```

cbaines: That's also true, but I was trying here to give direction should anyone decide to investigate.

At least when debugging the test failures for govuk-guix, I found several tests had quite specific requirements on the environment, so my guess is that something is up with the Dockerfile that's in use.

ChrisBAshton, line 85:
```suggestion
      * `make whitehall` requires that you have `publishing-api` running in another shell.
      * It also errors at the `taxonomy:populate_end_to_end_test_data` stage, but you can still `govuk-docker up`.
      * Most tests pass, but several tests fail consistently... (however you want to phrase this)
```

^ Based on a lot of trial-and-erroring this morning while trying to set up smart-answers (which somewhat confusingly has whitehall as a dependency, and kept breaking at the `make` stage as a result. But I was able to start it up regardless).

cbaines:
To back up my claim that this is an issue with the Dockerfile, I've tried running the tests myself. 

With the one change here [1], given the state of the universe as it was when I evaluated the Dockerfile to build the image, I was able to run the Rails, Cucumber and Teaspoon tests and there are no tests that consistently fail.

1: https://github.com/alphagov/govuk-docker/pull/369

kevindew:
Do we know what the tests are that consistently fail? 

The other day when I ran the full suite of unit tests (`rake parallel:test`) the ones I had fail were related to https://github.com/alphagov/whitehall/commit/ec98c5045eee19e17ac1e7edb82dc86cb8eeb15c where I need to run an assets:precompile first, once done all passed. I didn't run js or cucumber tests though.

cbaines:
> Do we know what the tests are that consistently fail?
> 
> The other day when I ran the full suite of unit tests (`rake parallel:test`) the ones I had fail were related to [alphagov/whitehall@ec98c50](https://github.com/alphagov/whitehall/commit/ec98c5045eee19e17ac1e7edb82dc86cb8eeb15c) where I need to run an assets:precompile first, once done all passed. I didn't run js or cucumber tests though.

I think I've had the exact same experience as you: https://github.com/alphagov/govuk-docker/pull/369#issuecomment-647996517

I don't know why I haven't experienced this when not using govuk-docker, I don't particularly remember running in to this problem. I can see now that for the Jenkins jobs, the assets are precompiled before running the tests.

This does suggest I was wrong about it being something to do with the image being used, although it's perfectly possible that people in the past weeks and months were generating different images that did have problems. I got the felling thought that no one was having success running the Whitehall tests with govuk-docker, but maybe that is just an issue with assets?

benthorner:
I'm going to close this for now. We can always reopen if we discover a specific compatibility issue to document.

PR #369, opened 2020-06-22
Add missing imagemagick dependency to the Whitehall Dockerfile

cbaines:
Whitehall uses the mini_magick Ruby library, and I believe this
Dockerfile is written with the intention of using bundler/rubygems to
install mini_magick.

Unfortunately, bundler/rubygems aren't capable enough to sort out
installing imagemagick, which is a problem, as it's needed to make
mini_magick work. Therefore, install the Debian package for
imagemagick through the Dockerfile.

### DIFF
--- a/projects/whitehall/Dockerfile
+++ b/projects/whitehall/Dockerfile
@@ -31,7 +31,7 @@ RUN /rbenv/plugins/ruby-build/install.sh
 ENV PATH /rbenv/bin:$PATH
 
 # Install Whitehall specific dependencies
-RUN apt-get update -qq && apt-get install -y default-mysql-client ghostscript
+RUN apt-get update -qq && apt-get install -y default-mysql-client ghostscript imagemagick
 
 RUN useradd -m build
 ENV PATH /home/build/.rbenv/shims:${PATH}


kevindew:
This shouldn't be necessary as imagemagick is already installed via the buildpack-deps image. For example:

```
➜  whitehall git:(master) govuk-docker-run apt-cache policy imagemagick
docker-compose -f [...] run whitehall-lite apt-cache policy imagemagick
Starting govuk-docker_mysql_1 ... done
Starting govuk-docker_redis_1 ... done
imagemagick:
  Installed: 8:6.9.7.4+dfsg-11+deb9u7
  Candidate: 8:6.9.7.4+dfsg-11+deb9u7
  Version table:
 *** 8:6.9.7.4+dfsg-11+deb9u7 500
        500 http://deb.debian.org/debian stretch/main amd64 Packages
        500 http://security.debian.org/debian-security stretch/updates/main amd64 Packages
        100 /var/lib/dpkg/status
```

benthorner:
Then having 

> This shouldn't be necessary as imagemagick is already installed via the buildpack-deps image. For example:
> 
> ```
> ➜  whitehall git:(master) govuk-docker-run apt-cache policy imagemagick
> docker-compose -f [...] run whitehall-lite apt-cache policy imagemagick
> Starting govuk-docker_mysql_1 ... done
> Starting govuk-docker_redis_1 ... done
> imagemagick:
>   Installed: 8:6.9.7.4+dfsg-11+deb9u7
>   Candidate: 8:6.9.7.4+dfsg-11+deb9u7
>   Version table:
>  *** 8:6.9.7.4+dfsg-11+deb9u7 500
>         500 http://deb.debian.org/debian stretch/main amd64 Packages
>         500 http://security.debian.org/debian-security stretch/updates/main amd64 Packages
>         100 /var/lib/dpkg/status
> ```

Hmm...looks like you're right. I was going to suggest something with the version, but it appears to be the latest one too. @cbaines are you sure the tests failed before making this change?

cbaines:
> Hmm...looks like you're right. I was going to suggest something with the version, but it appears to be the latest one too. @cbaines are you sure the tests failed before making this change?

They did, specifically these tests failed:

```
test/unit/helpers/govspeak_helper_test.rb:455
test/unit/helpers/govspeak_helper_test.rb:461
test/unit/helpers/govspeak_helper_test.rb:440
```

I was working with the assumption that something was missing from the Dockerfile, those tests relate to images for fractions, and I couldn't see imagemagick in the Dockerfile, so that's why I added it.

The other thing I remember doing was running `rake assets:precompile`, as I think those tests require an asset manifest to work. I've attempted to undo this change locally, and run the tests again, and they still all pass, so I think Kevin's right, I should have checked the source material for the buildpack-deps image or just if imagemagick was already installed before jumping in and changing this Dockerfile to install it.

I'll close this now.

PR #376, opened 2020-07-21
Add datagovuk_find application

pixeltrix:
Docker container runs in privileged mode to allow using Google Chrome in headless mode for feature specs.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,7 @@ volumes:
   mongo-3.6:
   go:
   elasticsearch-6:
+  elasticsearch-7:
 
 services:
   postgres-9.6:
@@ -53,6 +54,16 @@ services:
       - ES_JAVA_OPTS=-Xms1g -Xmx1g
     volumes:
       - elasticsearch-6:/usr/share/elasticsearch/data
+
+  elasticsearch-7:
+    image: elasticsearch:7.9.3
+    environment:
+      - http.host=0.0.0.0
+      - transport.host=127.0.0.1
+      - xpack.security.enabled=false
+      - ES_JAVA_OPTS=-Xms1g -Xmx1g
+    volumes:
+      - elasticsearch-7:/usr/share/elasticsearch/data
 
   nginx-proxy:
     image: jwilder/nginx-proxy:latest
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -33,7 +33,8 @@ These are repos that can be started as a some kind of process, such as a web app
    - ✅ content-publisher
    - ✅ content-store
    - ✅ content-tagger
-   - ❌ datagovuk_find
+   - ⚠ datagovuk_find
+      * **TODO: Missing support for a webserver stack**
    - ⚠ datagovuk_publish
       * **TODO: Missing support for a webserver stack**
    - ✅ email-alert-api
--- a/projects/datagovuk_find/Makefile
+++ b/projects/datagovuk_find/Makefile
@@ -1,0 +1,2 @@
+datagovuk_find: bundle-datagovuk_find
+	$(GOVUK_DOCKER) run $@-lite yarn
--- a/projects/datagovuk_find/docker-compose.yml
+++ b/projects/datagovuk_find/docker-compose.yml
@@ -1,0 +1,27 @@
+version: '3.7'
+
+volumes:
+  datagovuk_find-node-modules:
+  datagovuk_find-tmp:
+
+x-datagovuk_find: &datagovuk_find
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: datagovuk_find
+  stdin_open: true
+  tty: true
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - root-home:/root
+    - datagovuk_find-node-modules:/govuk/datagovuk_find/node_modules
+    - datagovuk_find-tmp:/govuk/datagovuk_find/tmp
+  working_dir: /govuk/datagovuk_find
+
+services:
+  datagovuk_find-lite:
+    <<: *datagovuk_find
+    depends_on:
+      - elasticsearch-7
+    environment:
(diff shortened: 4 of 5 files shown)


benthorner:
@pixeltrix did you want someone to review this?

pixeltrix:
> @pixeltrix did you want someone to review this?

Yes, please - I wasn't sure whether this needs to run through govuk-docker or it is meant to use can docker

benthorner:
We've got DGU Publish in here, just so we can do basic stuff with the app if needed. (The CKAN Docker repo was also not in existence at the time, so it's partly historical.)

I've got the build running locally, but most of the tests fail due to WebMock and unexpected requests. I think you'll need to add something like `allow: "elasticsearch6"` to the `spec_helper.rb` in the app.

cdccollins:
> We've got DGU Publish in here, just so we can do basic stuff with the app if needed. (The CKAN Docker repo was also not in existence at the time, so it's partly historical.)
> 
> I've got the build running locally, but most of the tests fail due to WebMock and unexpected requests. I think you'll need to add something like `allow: "elasticsearch6"` to the `spec_helper.rb` in the app.

I've opened a PR for this 😄 https://github.com/alphagov/datagovuk_find/pull/824

benthorner:
@cdccollins just wondering, how's this going?

pixeltrix:
@benthorner that change to allow elasticsearch6 has been merged now so hopefully the tests will pass now

benthorner:
@pixeltrix ah great! Are you able to confirm the tests pass yourself, or do you need some help?

pixeltrix:
I can confirm myself but have a day full of meetings so won't be able to get to it until tomorrow

benthorner:
No worries @pixeltrix, and thanks for trying it out 👍. I've tested locally and found a couple of issues:

- The [fix for the WebMock hosts](https://github.com/alphagov/datagovuk_find/pull/824) will break when @kevindew merges [a set of major changes](https://github.com/alphagov/govuk-docker/pull/394). Can we use a regex to allow any "elasticsearch*" host instead?

- When I run the tests locally I get failures because the `ES_HOST` variable is a URL, and only seems to work when I change it to `elasticsearch6:9200`. Do you get that problem yourself? (Note, I needed to clear out my containers with `govuk-docker down` in order for the change to be reflected.)

pixeltrix:
@kevindew here's a fix for the WebMock hosts: https://github.com/alphagov/datagovuk_find/pull/857/files

Investigating the local ES_HOST issue now

pixeltrix:
@benthorner two things I've found in my investigation:

1. When using a regex for `allow:` it matches against the full URI and not just the host - a string is just checked against a host and not the port, protocol, etc. so I'll need to create a new PR for that.

2. The datagovuk_find repo has been updated to Elasticsearch 7 in a way which is backwards incompatible with Elasticsearch 6 (the mappings in spec/support/elastic_helper.rb have changed). Not sure what you want to do? I can add an option for Elasticsearch 7 to govuk-docker but not sure if this is the best choice given @kevindew's changes for 2.0. Also the datagovuk_publish repo has been upgraded to Elasticsearch 7 as well so I suspect the existing govuk-docker project file is broken.

benthorner:
@pixeltrix thanks for the update.

> When using a regex for allow: it matches against the full URI and not just the host

Oops. Thanks for checking - let me know if you need that follow-up PR reviewing.

> I can add an option for Elasticsearch 7 to govuk-docker

Yes, that would be compatible with @kevindew's changes. Specifically, adding a `elasticsearch-7` service and a corresponding volume.

> so I suspect the existing govuk-docker project file is broken

Hmm...tricky. It has occurred to me that we may be duplicating effort if [docker-ckan also supports Find](https://github.com/alphagov/docker-ckan/pull/41). I'd be happy with removing Publish/Find and updating the compatibility doc to point towards docker-ckan, as it's own DGU-specific ecosystem. You may be better placed to make a call here: do you think there's an advantage in having DGU Publish/Find supported in GOV.UK Docker, as well as the DGU-specific version?

pixeltrix:
> do you think there's an advantage in having DGU Publish/Find supported in GOV.UK Docker, as well as the DGU-specific version?

One reason might be ease of doing upgrades, especially when there might be a conflict or error from the dependabot created PRs

benthorner:
@pixeltrix ah sorry, I should be clearer: would it be possible to use the docker-ckan repo to do all things involving DGU Publish/Find, instead of GOV.UK Docker? Is that something you've tried doing?

Perhaps the simplest thing to do here is to raise an issue, so this doesn't block the PR - assuming you're willing to get it to the point where we can merge it with all the tests passing.

pixeltrix:
> @pixeltrix ah sorry, I should be clearer: would it be possible to use the docker-ckan repo to do all things involving DGU Publish/Find, instead of GOV.UK Docker? Is that something you've tried doing?

The app has been upgraded to Ruby 2.7.2 and Rails 6.0 so this PR isn't really necessary to complete the upgrade work. If the recommendation is to use docker-ckan for DGU then we should close this PR and remove datagovuk_publish from govuk-docker - happy to create a new PR for that if you decide that's what you want.

benthorner:
@pixeltrix it's a tricky one: not people develop with DGU (me included), so it's not clear to me what would work best in practice. That said, I've had a read of [the docker-ckan repo](https://github.com/alphagov/docker-ckan), and noted that it's not really optimised for running Publish/Find - in particular, it makes no mention of running tests for them, and it looks like it requires a lot of setup just to run one of them in isolation. Since there doesn't seem to be much opinion either way, we may as well just get this in, noting that:

- It's got a small footprint, so should be relatively safe and easy to change in future.
- We can always revisit this decision if it gets out-of-sync again in the future.

pixeltrix:
@benthorner I've added the elasticsearch-7 service and confirmed that both DGU Find and Publish tests pass once alphagov/datagovuk_find#860 is merged.

benthorner (approved):
Nice - all tests pass for me with both apps (with the ES fix branch of DGU Find). Thanks for taking the time to finish this 🏅.

PR #378, opened 2020-07-22
Change govuk_publishing_components to use yarn

kevindew:
This reflects the change made in https://github.com/alphagov/govuk_publishing_components/pull/1610


projects/govuk_publishing_components/Makefile
```diff
@@ -1,2 +1,2 @@
 govuk_publishing_components: bundle-govuk_publishing_components
-	$(GOVUK_DOCKER) run $@-lite npm install
+	$(GOVUK_DOCKER) run $@-lite yarn install
```
andysellick, line 2: Don't we also need to run `npm install` or is that covered by yarn?

kevindew: It's covered by yarn. Yarn is essentially a drop in replacement for npm.

Issue #381, opened 2020-08-05
contacts-admin fails to build


huwd:
## Steps to reproduce

`make contacts-admin`

## Stack trace

```
/home/huw/govuk/govuk-docker/exe/govuk-docker run contacts-admin-lite sh -c 'bin/rake db:migrate 2>/dev/null || bin/rake db:create db:migrate'
docker-compose -f [...] run contacts-admin-lite sh -c bin/rake db:migrate 2>/dev/null || bin/rake db:create db:migrate
Starting govuk-docker_mysql_1 ... done
Starting govuk-docker_redis_1 ... done
Database 'contacts_development' already exists
Database 'contacts_test' already exists
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateContacts < ActiveRecord::Migration[4.2]
/govuk/contacts-admin/db/migrate/20130814072057_create_contacts.rb:1:in `<main>'

Caused by:
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateContacts < ActiveRecord::Migration[4.2]
/govuk/contacts-admin/db/migrate/20130814072057_create_contacts.rb:1:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
make: *** [/home/huw/govuk/govuk-docker/projects/contacts-admin/Makefile:2: contacts-admin] Error 1
```

## Thoughts

Interesting, I would have thought this would have come up in the contacts app by now if this was failing, also seems that in whitehall we have classes that inherit from `ActiveRecord::Migration` without a version, and i don't think this error is occuring there... a red herring perhaps?

Can anyone else give this a go and verify this is happening for them too?

## Notes
 
Why was I running this?
`¯\_(ツ)_/¯` wanted to see what would happen if I ran `make all-apps`, this was one of the things that happened

kevindew:
[...]

PR #382, opened 2020-08-06
Fix DB setup step to allow contacts-admin to build

huwd:
Resolves:  #318 

Tested with `make contacts-admin` - :+1:  it works!

The previous makefile was attempting to set up the initial DB by running
migrations which is not the [recommended appraoch for building the db of
a Rails app](https://guides.rubyonrails.org/v3.2/migrations.html#schema-dumping-and-you) as Kevin mentioned in [Issue #381](https://github.com/alphagov/govuk-docker/issues/381)

So I've brought this into line with [the whitehall makefile](https://github.com/alphagov/govuk-docker/blob/a9e00d16c613f42159897ea003e20b9ff3769c3e/projects/whitehall/Makefile#L2). This mirrors recent changes to [signon](https://github.com/alphagov/signon/pull/1458)

Co-authored-by: Kevin Dew <kevin-dew@digital.cabinet-office.gov.uk>


huwd:
Ah great spot thanks Kevin, I've recently switched to gpg-agent, so I suspect i've left something badly configured in the switchover

Issue #223, opened 2019-10-23
finder-frontend fails to build without publishing-api running


Tetrino:
**Impact**: Low
**Workaround**: Run publishing-api and static docker containers before building and running finder-frontend

After the implementation of the dynamic list work, finder-frontend will error when using `govuk-docker build` unless publishing-api is already built and running prior: 

```bash
/Users/jonathonshire/govuk/govuk-docker/bin/govuk-docker compose run search-api-setup rake publishing_api:publish_eu_exit_business_finder
docker-compose -f [...] run search-api-setup rake publishing_api:publish_eu_exit_business_finder
Starting govuk-docker_postgres_1 ... done
Starting govuk-docker_nginx-proxy-app_1 ... done
Starting govuk-docker_mongo_1           ... done
Starting govuk-docker_rabbitmq_1        ... done
Starting govuk-docker_redis_1           ... done
Starting govuk-docker_router-app-draft_1 ... done
Starting govuk-docker_router-api-app-draft_1 ... done
Creating govuk-docker_router-app_1              ... done
Creating govuk-docker_content-store-app-draft_1 ... done
Creating govuk-docker_router-api-app_1          ... done
Creating govuk-docker_content-store-app_1       ... done
Creating govuk-docker_publishing-api-worker_1   ... done
Creating govuk-docker_publishing-api-app-e2e_1  ... done
I, [2019-10-23T12:00:59.342136 #1]  INFO -- sentry: ** [Raven] Raven 2.11.3 configured not to capture errors: DSN not set
I, [2019-10-23T12:01:01.635643 #1]  INFO -- sentry: ** [Raven] Raven 2.11.3 configured not to capture errors: DSN not set
I, [2019-10-23T12:01:03.719706 #1]  INFO -- : Publishing Find Brexit guidance for your business (a finder_email_signup content item)
rake aborted!
GdsApi::HTTPBadGateway: URL: http://publishing-api.dev.gov.uk/v2/content/2818d67a-029a-4899-a438-a543d5c6a20d
Response body:
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.17.3</center>
</body>
</html>
```

It appears we're not running the app via the finder-frontend call at this point, which is resulting in the 502. Manually starting publishing-api before attempting to build and run finder-frontend solves the problem.

It's likely just a small fix involving what starts up in the build process, I just don't have time to look at it right now.

[...]

PR #385, opened 2020-08-07
Remove data insertion step for Search API

benthorner:
Closes: https://github.com/alphagov/govuk-docker/issues/223

This can/should be achieved with data replication, so it's not a
necessary part of the build process.


huwd:
Sorry as I merged #388 I've introduced a merge conflict, I think this change would also have resolved that issue.
If you're happy to rebase, this change makes sense to me and I'm happy to approve.

benthorner:
Thanks @huwd - I've rebased now, and I see you've approved already.

PR #386, opened 2020-08-14
Remove draft stacks for most apps

benthorner:
Refers to: https://github.com/alphagov/govuk-docker/pull/368

Previously these were added as part of the stretch goal of creating
this repo: to simulate end-to-end publishing on a local machine. In
reality, we don't use the local 'draft' stacks, so let's remove them.


benthorner:
@huwd good spot. Totally forgot about Nginx - we have [a check for adding things](https://github.com/alphagov/govuk-docker/blob/master/spec/integration/compose/virtual_hosts_spec.rb), but not removing them.

Updated.

PR #394, opened 2020-08-28
Breaking changes for version 2.0

kevindew:
## GOV.UK Version 2

This PR contains a number of breaking changes for GOV.UK Docker which are intended to go out as a version 2 release. These changes are:

- Ruby projects now run as a root docker user
- database services and mounts now have version numbers in their names
- PhantomJS is removed from the stack

There are further details about these changes in the commits.

## How to upgrade

You have two approaches you can take to upgrading to GOV.UK Docker 2. The simplest of these is the "Scorched earth" approach, which essentially deletes everything and rebuilds GOV.UK Docker. It then takes ~30 minutes to re build (dependent on which project you're working with). If you want to keep your data and/or save time you can take the more complicated "Upgrade in place" approach which can be done in about 5-10 mins.

### 1. Scorched earth

This is a nuke everything and start again approach. Good if you like a clean slate and to retrieve disk space, but it is time consuming. This is the simpler option and ideal for anyone who isn't very confident with Docker.

These steps assume that you only use docker for GOV.UK Docker, it will delete beyond just your GOV.UK Docker containers.

- Stop all containers: `docker ps -a -q | xargs docker rm -f && docker container prune` 
- Remove all image data `docker system prune -a`
- Remove all volumes `docker volume prune`
- Start apps by following initial usage guidance: https://github.com/alphagov/govuk-docker#usage

### 2. Upgrade in place 

This involves re-running the make command before using any project. This is good if you value your time, allows you to maintain database data but means you may get a tricky to diagnose error if you forget to `make` a project before running it and that you'll have old images/volumes to clean up later. This one is better if you are confident with Docker.

- Stop all your existing docker containers: `docker ps -a -q | xargs docker rm -f` (if you use docker for more than GOV.UK Docker you'll probably want to take a more delicate approach that only stops the GOV.UK Docker containers)
- If you wish to maintain data you can copy the previous database volumes you had with the following commands:
  - `docker volume create govuk-docker_postgres-9.6 && docker run --rm -i -t -v govuk-docker_postgres:/from -v govuk-docker_postgres-9.6 alpine ash -c "cd /from ; cp -av . /to"`
  - `docker volume create govuk-docker_mongo-3.6 && docker run --rm -i -t -v govuk-docker_mongo:/from -v govuk-docker_mongo-3.6 alpine ash -c "cd /from ; cp -av . /to"`
  - `docker volume create govuk-docker_mysql-5.5 && docker run --rm -i -t -v govuk-docker_mysql:/from -v govuk-docker_mysql-5.5 alpine ash -c "cd /from ; cp -av . /to"`
  - `docker volume create govuk-docker_elasticsearch-6 && docker run --rm -i -t -v govuk-docker_elasticsearch6:/from -v govuk-docker_elasticsearch-6 alpine ash -c "cd /from ; cp -av . /to"`
- for each project you wish to use run `make <project name>` from govuk-docker root directory before use (otherwise you'll get a cryptic error)
- you can now go about using GOV.UK Docker as normal, you'll likely have some alerts about orphaned containers you can remove these by running `govuk-docker down --remove-orphans` which will stop all currently running containers and remove any orphaned ones. You can clear old volumes with `docker volume --prune`.

## Experiencing problems?

If you're having any problems I recommend opening a [GitHub issue](https://github.com/alphagov/govuk-docker/issues/new) and assigning it to me. 

If you want/need to return to the previous version of GOV.UK Docker you can do this by checking out an earlier commit of GOV.UK (for example `git checkout b2fb90d`). You will then need to run the upgrade steps again - if you took the "Upgrade in place" approach you'll need to work out how to reverse the volume copies. It isn't advised to return to the previous version, so you should only do this if you hit a serious issue.


benthorner (commented):
Nice one - works for me! 🚀

Steps I did to test locally:

- Checkout this branch and run `make content-publisher`, which fails (as expected) when it gets to Publishing API
- Run `govuk-docker-run yarn` in the `content-publisher` directory
- Switch the Gemfile to use the `chrome-sandbox` branch of `govuk_test`
- Run `govuk-docker-run bundle exec rake` (succeeds)
- Run `govuk-docker-up` and open in browser (succeeds, and quickly too!)

For making the breaking change, you could take advantage of [the versioning system](https://github.com/alphagov/govuk-docker/blob/master/CONTRIBUTING.md).

benthorner:
I'm trying this out as a beta tester. I went for option one and then built Email Alert * and ran tests for Static.

Only one issue: stopping all containers command had a parse error; used `docker ps -a -q | xargs docker rm -f` instead.

kevindew:
> docker ps -a -q | xargs docker rm -f

Great thanks for that. That seems a better command anyway as it can handle no processes running which I means I can drop the conditional aspect.

benthorner (approved):
Thanks again for all your work on this @kevindew ⭐ 🌟 🤩. I've left a few minor comments around a theme: documentation. One of the teething problems with this repo has been getting longterm consistency, and some tests would be helpful there.

I'm still wondering about the `node_modules` volume. How could we provide guidance on that, since it's relatively rare among our projects? Perhaps an ADR around this would be the most pragmatic place, since the performance gains are fairly low for this volume.

projects/specialist-publisher/Dockerfile
```diff
@@ -1,33 +0,0 @@
...
-    rm -vr phantomjs-$VERSION-linux-x86_64
-
-# Install rbenv to manage ruby versions
-RUN git clone https://github.com/sstephenson/rbenv.git /rbenv
-RUN git clone https://github.com/sstephenson/ruby-build.git /rbenv/plugins/ruby-build
-RUN /rbenv/plugins/ruby-build/install.sh
-ENV PATH /root/.rbenv/shims:/rbenv/bin:$PATH
-ENV BUNDLE_SILENCE_ROOT_WARNING 1
-
-RUN apt-get update -qq && apt-get install -y qt5-default libqt5webkit5-dev gstreamer1.0-plugins-base gstreamer1.0-tools gstreamer1.0-x
```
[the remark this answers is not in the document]
kevindew: As I understand that is part of the qt webkit installation that isn't needed now this uses chrome.

projects/asset-manager/docker-compose.yml
```diff
@@ -10,6 +13,7 @@ x-asset-manager: &asset-manager
   volumes:
     - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
     - root-home:/root
+    - asset-manager-tmp:/govuk/asset-manager/tmp
```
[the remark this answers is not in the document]
kevindew: Good idea - I'll check that out. I've not worked with the tests across this.

projects/support-api/Dockerfile
```diff
@@ -1,35 +0,0 @@
-# Install packages for building ruby
-FROM buildpack-deps
```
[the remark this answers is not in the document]
kevindew: Oops probably slipped in with something else. I'll fix

kevindew:
Thanks for the review Ben, lots of good points. An ADR does sound a good idea. 

node_modules should rather rapidly become the norm as I apply stylelint across the apps. I did have a hacky idea on how to test it, could use the presence of yarn in an applications makefile to suggest it should be there, does make me a little nauseous at the hackiness though but would probably work.

kevindew:
@benthorner I've put a couple of tests in if you'd like to have a look. I'd not really looked at the other tests before and notice they're all rather circumstantial so the rather delicate node one I've put in feels consistent with existing approaches - I can see it's rather a challenge to work out where things are assertable.

spec/integration/compose/expected_volumes_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "spec_helper"
+
+RSpec.describe "Expected volumes" do
+  ProjectsHelper.rails_projects.each do |project_name|
```
benthorner, line 4: Nice test - good to pick up the other volumes to 💯.

I found the helper quite confusing, though:

- There's now a bit of a cycle between `ProjectsHelper` and `ComposeHelper`.
- The `rails_projects` is only subtly different to `rails_app-services`.

An alternative might be to have `rails_projects` inline with the test, like we do [the rather hairy makefile test](https://github.com/alphagov/govuk-docker/blob/master/spec/integration/make/dependencies_spec.rb). It's been a tricky task working out what helpers to have and how to compose them. Having Rails projects computed inline would at least reduce the effort digging in to the call stack.

kevindew: Sure, I've switched both `rails_projects` and `node_projects` to inline. We can deal with specificities if the helpers are needed elsewhere. Since the methods need to be available outside of an example they're sensitive to ordering.

spec/integration/compose/expected_volumes_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.rails_projects
+    rails_service_names = ComposeHelper.rails_app_services.keys
+
+    ProjectsHelper.all_projects.select do |project_name|
+      project_service_names = ComposeHelper.services(project_name).keys
+      rails_service_names.intersection(project_service_names).any?
+    end
+  end
+
+  rails_projects.each do |project_name|
```
benthorner, line 20: Minor: you could reduce the nesting around the examples by having the helper return `rails_services`.

benthorner (approved):
Thanks for adding the tests for those other volumes as well 💯. I think this is good to go. Perhaps we can make a note somewhere to write an ADR about the change? It could be an opportunity to reaffirm the reasons for not having pre-built images for everything, since that would have been another way to improve performance (but not really). I'd be willing to write it, if we want to extend the scope that far.

spec/integration/compose/expected_volumes_spec.rb
kevindew: Do you mean such that the loop is `rails_projects.each do |project_name, service_name, service|` ? 

If so I toyed with it but found it a pain to give the methods a nice name e.g `def services_for_rails_projects` and thought `rails_projects` didn't imply you'd get services. So I'm open to it if you have suggestions of how it can be cleaner.

benthorner: Oh yes, I see what you mean - I didn't clock that `project_name` is needed as well.

I can't think of a clearer way to articulate these helpers as-is. Perhaps something more OO would be the way to go in future, so you could (for example) ask a service object about its name, project and config. Similarly, we could have an object to represent Makefile rules - bet there's a gem for that.

Definitely out-of-scope here!

huwd (approved):

huwd:
@kevindew just rebased this against master as some members of my team need both this and the accounts stuff.
Hope that's ok?

kevindew:
> @kevindew just rebased this against master as some members of my team need both this and the accounts stuff.
> Hope that's ok?

No problem, thanks

Issue #389, opened 2020-08-17
GDS API Adapters fails to build


huwd:
In this episode of `Will it build`:

# Steps to reproduce

Join our plucky cast as they attempt to run:
`make gds-api-adapters`

## Stack trace

And encounter the mysterious case of the missing Argument 

```
/home/huw/govuk/govuk-docker/exe/govuk-docker run gds-api-adapters-lite rbenv install -s || (/home/huw/govuk/govuk-docker/exe/govuk-docker build --no-cache gds-api-adapters-lite; /home/huw/govuk/govuk-docker/exe/govuk-docker run gds-api-adapters-lite rbenv install -s)
docker-compose -f [...] run gds-api-adapters-lite rbenv install -s
/home/huw/govuk/govuk-docker/exe/govuk-docker run gds-api-adapters-lite sh -c 'gem install --conservative --no-document bundler -v $(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'
docker-compose -f [...] run gds-api-adapters-lite sh -c gem install --conservative --no-document bundler -v $(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)
grep: Gemfile.lock: No such file or directory
ERROR:  While executing gem ... (OptionParser::MissingArgument)
    missing argument: -v
make: *** [Makefile:41: bundle-gds-api-adapters] Error 1
```

## Also note

So... I suspect this is because the central `Makefile` [has an expectation there'll be a Gemfile.lock coming with each package](https://github.com/alphagov/govuk-docker/blob/master/Makefile#L41).
But in this case gds-api-adapters appears to be a gem itself.

So i guess in this case something more custom in the makefile, (or perhaps another task that focuses on the steps for Gems?) is the route?

kevindew:
Oh interesting, I guess that happens with most gems since they'll often not have a Gemfile.lock on initial cloning. I guess the options we'd want to consider is either only running `gem install bundler` when Gemfile.lock exists or to just install the latest version of bundler in that scenario. I think the former situation should be fine since Ruby bundles bundler in recent versions. Do you know if this commands works for you if you just comment out that line?

huwd:
Yep that works great if i comment that out.
So... humm we could either wrap this in an If statement, so IF there's a Gemfile.lock then run this command?
Or is there a more graceful failure state? Like if you run it and fail then continue?

my bash fu is a little lacking, if you had any starting point on syntax i'd be happy to go test until there's something PR-able?

[...]

PR #396, opened 2020-09-09
Only bundle install when a Gemfile.lock exists

huwd:
A response to issue #389, govuk-docker contains some gems that are
needed to use with other applications that tend not to have a
Gemfile.lock when initialy cloned. This was causing make (eg. `make
gds-api-adapters`) to fail.

Here we make the depdency explicit and skip the step if no Gemfile.lock
exists, a step that hasn't proved disruptive to using the adapters in
other applications.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ all-apps: $(APPS)
 bundle-%: clone-%
 	$(GOVUK_DOCKER) build $*-lite
 	$(GOVUK_DOCKER) run $*-lite rbenv install -s || ($(GOVUK_DOCKER) build --no-cache $*-lite; $(GOVUK_DOCKER) run $*-lite rbenv install -s)
-	$(GOVUK_DOCKER) run $*-lite sh -c 'gem install --conservative --no-document bundler -v $$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'
+	if [ -f Gemfile.lock ]; then $(GOVUK_DOCKER) run $*-lite sh -c 'gem install --conservative --no-document bundler -v $$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'; fi
 	$(GOVUK_DOCKER) run $*-lite bundle
 
 clone-%:


Makefile
```diff
@@ -38,7 +38,7 @@ all-apps: $(APPS)
 bundle-%: clone-%
 	$(GOVUK_DOCKER) build $*-lite
 	$(GOVUK_DOCKER) run $*-lite rbenv install -s || ($(GOVUK_DOCKER) build --no-cache $*-lite; $(GOVUK_DOCKER) run $*-lite rbenv install -s)
-	$(GOVUK_DOCKER) run $*-lite sh -c 'gem install --conservative --no-document bundler -v $$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'
+	if [ -f Gemfile.lock ]; then $(GOVUK_DOCKER) run $*-lite sh -c 'gem install --conservative --no-document bundler -v $$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'; fi
```
benthorner, line 41: What happens if we haven't installed bundler for a gem?

huwd: Interesting, I'd not considered that.
I'm going to go through the rest of the repos that have a gemspec and see what the state of play is. Then I'll feed back here if there _are_ any that don't have bundler installed. I should be able to see if this step trips up any others we try to make.

If there are not, then perhaps we can have a chat about best pratice and what's good enough for now?

If there are some, then perhaps I'll take this line and abstract it off into it's own make step or bash function that can choose if we install bundler with slightly more complex logic.

benthorner: Sure, thanks. I only recently learnt that Ruby now comes with bundler pre-installed, but it would be good to verify all the gemspec projects we support actually use a sufficient version for that. If not, they could just have their own custom Ruby/bundle thing in their Makefile, since they're intent on being special (or we could just upgrade Ruby!).

huwd: Ah interesting!
Looks like bundler has been in standard lib [since ruby 2.6.0](https://ruby-doc.org/core-2.6/doc/standard_library_rdoc.html).
I'll take a quick look now for anything we've got in GOV.UK docker with a gemspec, and see if it's >2.6.0.

huwd: so I think i've found five gems that are supported by `govuk-docker`.
From my `~/govuk` if i run:
```
cat govspeak/.ruby-version && \
cat plek/.ruby-version && \
cat gds-api-adapters/.ruby-version && \
cat govuk_publishing_components/.ruby-version && \
cat govuk_app_config/.ruby-version
```
i get:

```
2.6.3
2.6.6
2.6.6
2.6.6
2.6.5
```

So currently I believe all supported gems are a version above 2.6.x and so should come with bundler in standard lib! :tada:

huwd: So for the moment you happy to approve with this if statement?
If we have a need for more gems in future, or to support earlier ruby versions I'm happy to look at something more elegant?

benthorner: Nice, thanks for investigating! I'm glad we can keep it simple.

benthorner (approved):
🎉

PR #401, opened 2020-09-17
Refresh compatibility list

benthorner:
https://trello.com/c/3vrf07uj/190-complete-govuk-docker-compatibility-list

This came out of a suggestion in the Dev Tools Survey. In
general, we should clarify the scope of what GOV.UK Docker
could support. Please see the commits for more details.


docs/compatibility.md
```diff
@@ -1,5 +1,7 @@
 # Compatibility
 
+## Executable "apps"
+
 The following apps are supported by govuk-docker to some extent.
```
kevindew, line 5: This line should change (or be removed) if we're also including things that aren't supported.

docs/compatibility.md
```diff
@@ -1,5 +1,7 @@
 # Compatibility
 
+## Executable "apps"
```
kevindew, line 3: Should we use the terminology "projects" around this since the differentiator for this list and the following one is whether it's a project or not as I understand.

docs/compatibility.md
```diff
@@ -83,8 +82,26 @@ The following apps are supported by govuk-docker to some extent.
...
+   - ❌ transition-config
    - ✅ travel-advice-publisher
    - ⚠ whitehall
       * Who knows, really - several tests are failing, lots pass ;-)
       * Rake task to [create a test taxon](https://github.com/alphagov/whitehall/blob/master/lib/tasks/taxonomy.rake#L11) for publishing is not idempotent
       * Placeholder images don't work as missing proxy for [/government/assets](https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/news_article_presenter.rb#L133)
+
+## Generic Ruby libraries
+
+The following apps are supported as part of a "generic-ruby-library" project.
```
kevindew, line 94: Similar point about "supported". It also be useful to define what qualifies for being a generic ruby project.

kevindew (commented):
Looks good, I like the consolidation.

docs/compatibility.md
ChrisBAshton: And also a brief line about what generic-ruby-library actually does / how it's useful. 👍

docs/compatibility.md
benthorner: I've rewritten this based on the other comments.

docs/compatibility.md
benthorner: The use of "app" here means the project is executable in some way. Everything is a project. Would it help to say `Executable "app" projects` instead?

docs/compatibility.md
benthorner: Sure. I've clarified this to explain in more detail.

benthorner:
Thanks for taking a look @ChrisBAshton and @kevindew. I ended up rushing this a little, as I didn't expect the decision about what to/not include to become nuanced. Your comments speak to that, and I've added some explanations in the compatibility doc to try specify the scope more clearly. Could you take another look?

ChrisBAshton (approved):
LGTM 👍

kevindew (approved):

PR #402, opened 2020-09-21
Fix for Specialist Publisher failing make command

Tetrino:
Couldn't run the Make command because of missing updates in the apt-get
library triggering 404 errors. Added an apt-update to make sure it can
grab the needed resources correctly.

### DIFF
--- a/projects/specialist-publisher/Dockerfile
+++ b/projects/specialist-publisher/Dockerfile
@@ -27,7 +27,7 @@ RUN /rbenv/plugins/ruby-build/install.sh
 ENV PATH /rbenv/bin:$PATH
 
 # Dependencies required for Capybara Webkit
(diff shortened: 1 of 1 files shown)


thomasleese (approved):

huwd (commented):
Interesting, why do we think this isn't getting updated with [line 5](https://github.com/alphagov/govuk-docker/blob/aedc90f36e53ea78cd0b23ce4f0c5dd3fabe113a/projects/specialist-publisher/Dockerfile#L5) in this file?

I can't see us adding a source for capybara between there and this line, am i missing something?

benthorner:
> am i missing something?

This is because [the image automatically does an `apt-get clean` after any `apt-get` update](https://github.com/docker-library/buildpack-deps/pull/41). This is evidences by us already having multiple `apt-get update`s.

huwd:
Ah ha, great to know thanks for that link too!

huwd (approved):
LGTM

Tetrino:
Was away for the best part of a week. Sorry for the delay!

PR #403, opened 2020-09-22
Add whitehall-frontend.dev.gov.uk as a virtual host

maxgds:
The existing host for whitehall is whitehall-admin.dev.gov.uk which
is not intuitive when trying to run the app to view frontend pages.
This allows either url to work on the frontend parts of whitehall.

### DIFF
--- a/projects/whitehall/docker-compose.yml
+++ b/projects/whitehall/docker-compose.yml
@@ -37,7 +37,7 @@ services:
     environment:
       DATABASE_URL: "mysql2://root:root@mysql/whitehall_development"
       REDIS_URL: redis://redis
-      VIRTUAL_HOST: whitehall-admin.dev.gov.uk
+      VIRTUAL_HOST: whitehall-admin.dev.gov.uk, whitehall-frontend.dev.gov.uk
       HOST: 0.0.0.0
     expose:
       - "3000"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,6 +107,7 @@ services:
           - support.dev.gov.uk
           - transition.dev.gov.uk
           - travel-advice-publisher.dev.gov.uk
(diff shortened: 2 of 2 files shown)


projects/whitehall/docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
     environment:
       DATABASE_URL: "mysql2://root:root@mysql/whitehall_development"
       REDIS_URL: redis://redis
-      VIRTUAL_HOST: whitehall-admin.dev.gov.uk
+      VIRTUAL_HOST: whitehall-admin.dev.gov.uk, whitehall.dev.gov.uk
```
kevindew, line 40: I'd suggest going with whitehall-frontend.dev.gov.uk here. That would be consistent with the host whitehall is given within our system and the one that we used to use in the development VM. 

If I remember right there is also a host restriction for accessing admin so whitehall.dev.gov.uk may imply being able to access both sides but I don't think that will let you into the admin parts.

kevindew (commented):
This seems a good idea 👍  Just a counter suggestion on the hostname.

projects/whitehall/docker-compose.yml
maxgds: fair point, I was going to add both whitehall and whitehall-frontend but thought it was overkill

benthorner:
Thanks for raising this @maxgds ! Have you tried this out locally? I think you may need to add the new hostname as [another alias](https://github.com/alphagov/govuk-docker/blob/master/docker-compose.yml#L110) and restart the `nginx-proxy` container.

maxgds:
> Thanks for raising this @maxgds ! Have you tried this out locally?

I did (with whitehall before switching to whitehall-frontend) and it ran fine.

benthorner (approved):
Nice one 👍

PR #416, opened 2020-10-07
Remove all-apps from Makefile

### LINKED ISSUE #407: Generic Ruby Library will not build (and shouldn't try to!)
## Steps to reproduce

`make all-apps`

## Issue

[We've talked about removing all-apps in the past.](https://github.com/alphagov/govuk-docker/issues/384#issuecomment-669862676)
It currently tries to build the following list:
```
asset-manager authenticating-proxy cache-clearing-service calculators collections collections-publisher contacts-admin content-data-admin content-data-api content-publisher content-store content-tagger datagovuk_publish email-alert-api email-alert-frontend email-alert-service feedback finder-frontend frontend generic-ruby-library government-frontend govuk_crawler_worker govuk-developer-docs govuk_publishing_components hmrc-manuals-api imminence info-frontend licence-finder link-checker-api local-links-manager manuals-frontend manuals-publisher mapit maslow miller-columns-element publisher publishing-api release router-api router search-admin search-api service-manual-frontend service-manual-publisher short-url-manager signon smart-answers specialist-publisher special-route-publisher static support-api support transition travel-advice-publisher whitehall
```

So it will try and build generic-ruby-library.

However running `make generic-ruby-library` fails because the makefile there has a set of more specifc rules.

## Solution

We should remove `all-apps` or ensure that it does not include `generic-ruby-library` or provide a separate make step to cover everything in that makefile.

NB: This also means that the five generic ruby files do not appear in ALL_APPS, if we're trying to test builds that way.

huwd:
Closes #407

The usefulness of this command has declined over time, and since we moved the generic ruby libraries to their own directory this command will fail as it tries to build based on folder name.

As this is not a command most users might be expected to run, we're removing it here.

If you have a need to build your own apps, you'd be better off doing so by running your own command to list relevant directories.

The following takes the existing command and does an inverted grep to create a blocklisti that could prove useful
`ls projects/*/Makefile | xargs -L 1 dirname | xargs -L 1 basename | grep -v generic-ruby-library`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ APPS ?= $(shell ls ${GOVUK_DOCKER_DIR}/projects/*/Makefile | xargs -L 1 dirname 
 
 # Best practice to ensure these targets always execute, even if a
 # file like 'clone' exists in the current directory.
-.PHONY: clone pull test all-apps
+.PHONY: clone pull test
 
 default:
 	@echo "Run 'make APP-NAME' to set up an app and its dependencies."
@@ -30,10 +30,6 @@ test:
 
 	# Validate shell scripts
 	$(SHELLCHECK) $(shell ls ${GOVUK_DOCKER_DIR}/bin/*.sh ${GOVUK_DOCKER_DIR}/exe/*)
-
-# This will be slow and may repeat work, so generally you don't want
-# to run this.
-all-apps: $(APPS)
 
 bundle-%: clone-%
 	$(GOVUK_DOCKER) build $*-lite


benthorner (approved):
Thanks @huwd! I think this was originally a replacement for "building a whole dev VM", which (unsurprisingly) no one actually wants to do, so it's good to get rid of it at last.

By the way, you can catch 'em all (including the generic ones) with:

```
cat projects/*/Makefile | grep ': ' | cut -d ':' -f 1 | xargs make
```

PR #419, opened 2020-10-08
[DO NOT MERGE] GOV.UK account discovery | Turn on GOV.UK account feature

huwd:
Passes the feature flag to finder frontend allowing for GOV.UK account correspndance with the brexit checker.

Also makes govuk-accounts a depdendency of both the app and live stack


brucebolt (approved):
It'd be prudent to let the other teams working on finder-frontend know they'll need to install new dependencies next time they run it locally.

huwd:
Decided to go with a slightly different approach, if this is approved: https://github.com/alphagov/govuk-docker/pull/431
Then it will supercede this branch.

PR #421, opened 2020-10-09
Add http to finder-frontend VIRTUAL_HOST

brucebolt:
In https://github.com/alphagov/finder-frontend/pull/2239, we have allowed non-http to be used as the virtual host, so now need to specify the protocol in the environment variable.

### DIFF
--- a/projects/finder-frontend/docker-compose.yml
+++ b/projects/finder-frontend/docker-compose.yml
@@ -32,7 +32,7 @@ services:
       - nginx-proxy
     environment:
       ASSET_HOST: finder-frontend.dev.gov.uk
-      VIRTUAL_HOST: finder-frontend.dev.gov.uk
+      VIRTUAL_HOST: http://finder-frontend.dev.gov.uk
       BINDING: 0.0.0.0
       MEMCACHE_SERVERS: memcached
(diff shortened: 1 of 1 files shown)


projects/finder-frontend/docker-compose.yml
```diff
@@ -32,7 +32,7 @@ services:
       - nginx-proxy
     environment:
       ASSET_HOST: finder-frontend.dev.gov.uk
-      VIRTUAL_HOST: finder-frontend.dev.gov.uk
+      VIRTUAL_HOST: http://finder-frontend.dev.gov.uk
```
kevindew, line 35: Changing this seems like a bit of a smell where this environment variable is being used beyond it's intention and probably indicates that finder-frontend should not be using it.

This env var is for [nginx-proxy docker container](https://github.com/nginx-proxy/nginx-proxy#docker-compose) and shouldn't be something the application is aware of. 

Hopefully you can use Rails request context to not need to know about the hostname and allow it to redirect based on the incoming request using same hostname and protocol.

brucebolt: Thanks Kevin.  I changed my mind immediately after creating the PR (hence why I closed it) and will solve the problem I'm having in another way.

PR #422, opened 2020-10-09
Fix image named after gds-api-adapters

kevindew:
This shouldn't have the name of a particular project, I'm assuming it was just something missed with the change.


benthorner (approved):
Oops, thanks for spotting! My "refactoring" is taking a bit of a hammering in various PRs 😬

PR #431, opened 2020-11-11
Create app-account stack, add to finder-frontend + collections

huwd:
## What

- Creates a new `app-*` stack called `app-account`
- Adds it to collections (so that the transition header will work with the feature flag)
- Addsit to finder-frontend (so the Brexit transition checker will work with accounts)

## Why
Finder frontend is currently being used as part of the trial / prototype
of GOV.UK accounts. These integrations are likely to be tested with
other apps in the near future.

A re-usable account sstack would therefore be a useful shorthand to
developers giving the expectation that it will:
- spin up dependent apps required to make account features work
- provide all environment variables (eg: keys, feature flags) to ensure
things run smoothly.

## New dependencies here
For finder frontend, that adds the additional dependencies of
email-alert-api (used to get subscription titles to send to the account
in the JWT) and the account manager prototype itself.

I've also moved environment variables out of other stacks, the expecation
of using a vanilla `app` stack should be that accounts will not be enabled.
Which should also provide a useful way to test locally that services fail gracefully
given the absence of a dependent service and enviornment variables.

### DIFF
--- a/projects/finder-frontend/docker-compose.yml
+++ b/projects/finder-frontend/docker-compose.yml
@@ -35,11 +35,6 @@ services:
       VIRTUAL_HOST: finder-frontend.dev.gov.uk
       BINDING: 0.0.0.0
       MEMCACHE_SERVERS: memcached
-      GOVUK_ACCOUNT_OAUTH_CLIENT_ID: transition-checker-id
-      GOVUK_ACCOUNT_JWT_KEY_PEM: "-----BEGIN EC PRIVATE KEY-----\nMHcCAQEEID+UE4jB5QtUFOtekHJliaSB8K2xQU2AS4xczOLAqvCUoAoGCCqGSM49\nAwEHoUQDQgAEjMgE/d6Bdu1K17BTjNycDvIXKSmUgZ5YYoayU3gGqTLhgefuK2qb\no99Wx+SuvdW/GRvIlUIW1ooNRdQ3QFwwCw==\n-----END EC PRIVATE KEY-----\n"
-      GOVUK_ACCOUNT_JWT_KEY_UUID: "898d62b7-eed9-464a-a4ae-9d9e08bd9bee"
-      GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET: transition-checker-secret
-      PLEK_SERVICE_ACCOUNT_MANAGER_URI: "http://www.login.service.dev.gov.uk"
     expose:
       - "3000"
     command: bin/rails s --restart
@@ -56,8 +51,24 @@ services:
       PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
       VIRTUAL_HOST: finder-frontend.dev.gov.uk
       BINDING: 0.0.0.0
+
+  finder-frontend-app-account:
+    <<: *finder-frontend-app
+    depends_on:
+      - router-app
+      - nginx-proxy
+      - email-alert-api-app
+      - static-app
+      - memcached
+      - govuk-account-manager-prototype-app
+    environment:
+      ASSET_HOST: finder-frontend.dev.gov.uk
+      VIRTUAL_HOST: finder-frontend.dev.gov.uk
+      BINDING: 0.0.0.0
+      MEMCACHE_SERVERS: memcached
       GOVUK_ACCOUNT_OAUTH_CLIENT_ID: transition-checker-id
       GOVUK_ACCOUNT_JWT_KEY_PEM: "-----BEGIN EC PRIVATE KEY-----\nMHcCAQEEID+UE4jB5QtUFOtekHJliaSB8K2xQU2AS4xczOLAqvCUoAoGCCqGSM49\nAwEHoUQDQgAEjMgE/d6Bdu1K17BTjNycDvIXKSmUgZ5YYoayU3gGqTLhgefuK2qb\no99Wx+SuvdW/GRvIlUIW1ooNRdQ3QFwwCw==\n-----END EC PRIVATE KEY-----\n"
       GOVUK_ACCOUNT_JWT_KEY_UUID: "898d62b7-eed9-464a-a4ae-9d9e08bd9bee"
       GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET: transition-checker-secret
       PLEK_SERVICE_ACCOUNT_MANAGER_URI: "http://www.login.service.dev.gov.uk"
+      FEATURE_FLAG_ACCOUNTS: "enabled"
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Variations on the `app` stack are allowed where necessary such as:
 
   - **app-draft**: used for testing the [authenticating-proxy](https://github.com/alphagov/govuk-docker/tree/master/projects/authenticating-proxy) against a draft version of the [router](https://github.com/alphagov/govuk-docker/tree/master/projects/router) app.
   - **app-live**: used to test a read-only frontend app against live GOV.UK APIs (avoids having to replicate data locally).
+  - **app-account**: used to enable integrations with a local [account manager app prototype](https://github.com/alphagov/govuk-account-manager-prototype) and [attribute store](https://github.com/alphagov/govuk-attribute-service-prototype/). Currently a part of the [GOV Accounts trial](https://gds.blog.gov.uk/2020/09/22/introducing-gov-uk-accounts/)
 
 Some `app` stacks also depend on a `worker` stack, to run asynchronous tasks [[example](https://github.com/alphagov/govuk-docker/blob/d286748e0300df8f0d1ed618086d4f8f951e752a/projects/content-publisher/docker-compose.yml#L46)].
 
--- a/projects/collections/Makefile
+++ b/projects/collections/Makefile
@@ -1,1 +1,1 @@
-collections: bundle-collections content-store router static
+collections: bundle-collections content-store router static govuk-account-manager-prototype
--- a/projects/collections/docker-compose.yml
+++ b/projects/collections/docker-compose.yml
@@ -44,3 +44,14 @@ services:
       PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
       VIRTUAL_HOST: collections.dev.gov.uk
       HOST: 0.0.0.0
+
+  collections-app-account:
+    <<: *collections-app
+    depends_on:
+      - nginx-proxy
+      - govuk-account-manager-prototype-app
+    environment:
+      FEATURE_FLAG_ACCOUNTS: "enabled"
+      ASSET_HOST: collections.dev.gov.uk
+      VIRTUAL_HOST: collections.dev.gov.uk
+      HOST: 0.0.0.0
--- a/projects/finder-frontend/Makefile
+++ b/projects/finder-frontend/Makefile
@@ -1,2 +1,2 @@
-finder-frontend: bundle-finder-frontend content-store static search-api router
+finder-frontend: bundle-finder-frontend content-store static search-api router email-alert-api govuk-account-manager-prototype
 	$(GOVUK_DOCKER) run $@-lite yarn


projects/finder-frontend/docker-compose.yml
```diff
@@ -56,8 +51,24 @@ services:
...
+    <<: *finder-frontend-app
+    depends_on:
+      - nginx-proxy
+      - email-alert-api-app
+      - govuk-account-manager-prototype-app
+    environment:
+      PLEK_SERVICE_SEARCH_URI: https://www.gov.uk/api
+      PLEK_SERVICE_WHITEHALL_FRONTEND_URI: https://www.gov.uk
+      GOVUK_WEBSITE_ROOT: https://www.gov.uk
+      PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
```
kevindew, line 65: These live services seem a bit strange to me, it seems like this has become half live? I'm guessing this is desired for convenience but it would be good to be clear that what in an account stack is live or not.

It'd feel more logically consistent if these weren't live dependencies though as that'd keep it fully distinct from the live service and feel more like an extension of the app stack.

huwd: ah true, that's just a mistake on my part - will remove those.
:thinking: given this inherits from app will it inherit:
```
      ASSET_HOST: finder-frontend.dev.gov.uk
      VIRTUAL_HOST: finder-frontend.dev.gov.uk
      BINDING: 0.0.0.0
      MEMCACHE_SERVERS: memcached
```

or do we have to declare these each time?
I see the `app` stack and `app-live` stack already have `BINDING` duplicated, do we know if that is a requirement?

huwd: Thanks @kevindew [amended here](https://github.com/alphagov/govuk-docker/pull/431/commits/c36f5b3d5ebc1303aa2912c515e7e060426ce311)

kevindew: No it doesn't inherit any of the environment variables you have to declare them each time. It'll only inherit if we use a yaml merge within an object.

Do you need the other dependencies of the app: https://github.com/alphagov/govuk-docker/blob/c36f5b3d5ebc1303aa2912c515e7e060426ce311/projects/finder-frontend/docker-compose.yml#L26-L32 as they're not inherited either?

There is a part of me that wonders if it'd be simpler having accounts as just part of the app stack - do we anticipate that to be massively inconveniencing? I'm just wondering if we're prematurely worried and creating a snowflake.

huwd: oh! I thought those were inherited too.
so for the checker I should need:
- router-app
- static-app
(:thinking: I did test this so I wonder how I was getting to any pages without a router...)

I suspect I can run all elements of the checker without:
- search-api-app
- content-store-app

and will need to check on
- memcached

This presents an interesting dilemma though, with different dependencies `app-account` would not be an enhancement to the `app` stack... instead it would accept that instead of the entire app you're only interested in the accounts bit... so anyone running `app-accounts` who goes over to a proper finder or search, would find that not working... is that ok...? :thinking:

huwd: > There is a part of me that wonders if it'd be simpler having accounts as just part of the app stack - do we anticipate that to be massively inconveniencing? I'm just wondering if we're prematurely worried and creating a snowflake.

Yeah I'm open to this idea, and also bit concerned my snow-flake-y-ness too. The pros as I see them are:

- I strongly suspect our trial will continue to other applications at some point soon (new or existing tbc)
- This should result in less disruption for those who are also working in these apps (e.g. Brexit Transition folks working on the checker will not neet to get accounts set up if they're still running the traditional `app` stack)... so it's a concious opt-in to accounts which seems nice to me...
- For apps that have a `live` stack i think it clearly sets these out as something that has no overlap... so you can't have an `app-live` with working accounts (I'm assuming it would be confusing to get real data from the content store... but be speaking to a local development account)

Does that sway you at all? Or do you think we should put these back in `app` and then deal with this when there's a more presing need?

kevindew: Thanks Huw. Accounts does present an interesting dilemma for the live stack where before it represented everything being live bar the application, whereas it will now mean almost everything live 🤔  unless we're able to work out how apps can integrate without accounts without needing secrets for accounts (is that a goal? would be ace)

I'm not massively opinionated either way so can be persuaded that an app-account stack is the way to go. I am leaning more towards just putting accounts in the general app stack though as it feels like this is more conventional with existing practices and opens up less ambiguities. It feels like something that should be considered further if evidence suggests it does inconvenience people, and if that's the case it might be a question of "is accounts unnecessarily hard to run?"

huwd: I suspect dev-live will be a consistent issue - the applications that speak to accounts need to be "registered" with the account manager and given the right shared secret keys to prove they're authorised to do the things they're requesting - and to prevent anyone and their cat from firing off production requests.

So unless we had something where the "live" stack could speak to a remote server (staging?) and we were happy for those staging secrets to leak into dev (... which I can't get a really good use case for right now...) then I suspect there'll continue to be a rift between live and development setups.

This is probably a larger thing for us to consider, perhaps at the upcoming tech fortnightly? I guess many of our tools have been predicated on us having unauthenticated APIs in production and transactional services. Perhaps Accounts is the question of what do we do when that assumption no longer holds... (and how likely is that at this point?)

Interesting interesting...
I will catch up with my team and get their review here, and if you are flexible, then I'll do one of two things:
- If my team likes `app-account` I'll merge this and then keep an eye on it, at the end of our trial I'm happy to revert this PR and then address the larger question of accounts vs our dev tools when we've committed to services that aren't in alpha?
- If my team isn't excited about `app-account` then I'll follow your suggestion, move this back into `app` and put a shout out to anyone else working in `finder-frontend` that if they're pulling master they'll need our two repos too.

I think those questions about how easy/difficutly accounts might be to setup are interesting ones, throw them our way during out acconts tech fortnightly update?

kevindew: Great stuff! main ask if you go with an app-accounts stack is that you're clear about what should and shouldn't go in it to try avoid confusion over some apps being live APIs and others not.

Yeah with regards to how it'll work in live stacks, we have dealt with similar with signon by apps using a gem to interact and the mock environment for development. Maybe some inspiration could be taken from how those problems were solved if it's applicable?

barrucadu: > Accounts does present an interesting dilemma for the live stack where before it represented everything being live bar the application, whereas it will now mean almost everything live 🤔 unless we're able to work out how apps can integrate without accounts without needing secrets for accounts (is that a goal? would be ace)

There's perhaps some ideas we can take from gds-sso here, where when running in development mode the user gets signed into the first account.  Maybe we can make some secrets optional:

- `GOVUK_ACCOUNT_JWT_KEY_PEM` and `GOVUK_ACCOUNT_JWT_KEY_UUID` - I think we could make signing the JWT optional in dev mode.  Currently we only allow one cipher, but we could allow that + (if dev) no cipher.
- `GOVUK_ACCOUNT_OAUTH_CLIENT_ID` and `GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET` - these are trickier because the client ID identifies which app the user is coming from, but maybe we could make these more consistent and able to have a sensible default value in dev?

How exactly GOV.UK apps are going to integrate with accounts when it gets rolled out to more things is still fairly unknown.  We've currently got special-case code in finder-frontend and collections, but if we make a new gem, or move the common code to govuk_app_config, then we could put this dev secrets management logic there.  I think we should write an RFC for that though, as it'll be a big new thing (ramifications for caching, session management, etc).

I'm not sure how making the live stack work with live accounts would work though, unless we maybe do something like add an OAuth app (to the production accounts system) for dev.gov.uk.

---

> If my team likes `app-account`

I think I've got a slight preference towards a separate stack for just accounts.  It keeps this stuff isolated to our team for now, which is good while we're still mulling over topics like how secrets should work in dev.

The way we're mostly doing things at the moment is using the live stack + an accounts system running in govuk-docker, so it'd be good to keep that if possible - for example, collections won't render /transition without a content item.  Needing to replicate data (or just edit the docker-compose file) to try out changes to that page would be kind of a pain.

---

> tech fortnightly

We've put a 45 minute slot in for the 1st of December tech fortnightly to have a bit more of an in-depth introduction to the accounts system and the tech it's built on, which will hopefully be useful to people.

huwd: I think for now, we've avoided this one becuase we're looking at the checker that stores all it's own data... but there's interest in what wider personalisation would look like...

... which now i think about it will likely require a lot of content flying around when we're sense checking our development...
Good to have this on the radar early. Thanks!

huwd: > The way we're mostly doing things at the moment is using the live stack + an accounts system running in govuk-docker, so it'd be good to keep that if possible - for example, collections won't render /transition without a content item. Needing to replicate data (or just edit the docker-compose file) to try out changes to that page would be kind of a pain. Or maybe we can live without that if it'll be confusing...

Yeah `/transition` is our first challenge here. I have an instinctive aversion to the idea of chaining these... so if we avoid a solution to this being anything like collections-app-account-live`.

I'm tempted to say shrug a bit about what the answer is for `collections` at the moment as the impact is small and we can test it in other ways, but if we start seeing greater needs to adapt that page then we should consider a dev setup early.

I suspect we'll still be working with a single content item for `/transition` for the forseeable, so perhaps we could find a way to just replicate that item locally with an `app-account` stack for any testing there?

brucebolt (commented):
This looks good to me, and I like having the separate stack for accounts.

One concern I have is that (as far as I can see), the Makefiles require lots of account/email specific dependencies to be built regardless of the stack.  Is there any way we can stop this happening so non-accounts developers don't need to have lots of applications built unnecessarily?

huwd:
Humm that's a good queston @brucebolt 
I guess I'd been thinking about it as an extension of app:

- `app` - everything you need to make the entire thing work
- `app-accounts` - everything from app + what you need to integrate an account...

But now I think about it `app-live` already blurrs that line slightly, it chooses _not_ to spin up some services because it provides the real URLs for live ones instead, but the full functionality is always there...

So when you are working in the accounts stack, should that automatically turn off some features that are "not relevant to accounts" in a local app? Will that fit with a user expectation?

I think within our team... it probably would, and it being faster is also nice.
So I'll modify this, but would appreciate anyone chipping in if we think that kind of feature subset is not a good pattern to follow.

huwd:
Sorry @brucebolt  re-read this and realised you were talking about the Makefile and not the docker-compose dependencies themselves.

Humm, so i think the answer is not currently.

Each Makefile has a single command currently that represents building everything that app might need across any stack.

We could add a second rule to the Makefile along the lines of:

```Makefile
finder-frontend: bundle-finder-frontend content-store static search-api router
	$(GOVUK_DOCKER) run $@-lite yarn

finder-frontend-app-account: bundle-finder-frontend email-alert-api govuk-account-manager-prototype
	$(GOVUK_DOCKER) run $@-lite yarn
```

This could also lighten the amount make finder-frontend has to do.

But I guess I'm not sure if users would expect everything on the stack to be there with finder-frontend or not?

huwd:
Humm ok this change breaks the current expectations of the tests.
I could go modify those but I'm beginning to think a change to multiple forms of make in an app should be more widely discussed.

I'm going to revert this to the former setup of a single make command that covers all possible stacks.
I'll then keep an eye out for any complaints that this has added undue friction and perhaps we can optimise from there?

brucebolt (approved):
Looks good to me 👍 .  We should check in with other users of finder-frontend just to make them aware of these changes.

huwd:
Yeah I'll give the brexit folks a shout and put a line in the govuk developers slack channel.

The upshot of it being a separate stack is that it shouldn't disrupt anyone outside of our team.
They should be able to keep running `govuk-docker up finder-frontend-app` with no changes.

The only odd bit will be that without accounts they don't get the `/save-your-results` page as appeas on live

PR #432, opened 2020-11-13
Restructure installation and how-to guidance

benthorner:
https://trello.com/c/i3Y2vAua/188-documentation-on-how-to-troubleshoot-apps-failing-to-start-in-govuk-docker

This is a somewhat ambitious PR to simplify the different kinds of guidance in GOV.UK Docker, so that it's easier to find and extend. The aims of this PR are:

- A clear and minimal README that covers all the key points to get started.
- A single, dedicated manual for troubleshooting, which I aim to extend in a future PR.
- A single, dedicated manual for how-tos, with common ones linked in the README.

One of the positive side effects of making these changes has been adding some guidance around testing, which was previously unclear / fragmented, since it overlaps with getting the `bin/doctor` script to work.

This is quite a big and opinionated change, so I'm open to breaking it down into smaller PRs, so we can make some quick progress on less contentious parts. Please see the commits for more details.


CONTRIBUTING.md
```diff
@@ -2,6 +2,30 @@
...
+
+**First make sure you have the following dependencies.**
+
+- [rbenv](https://github.com/rbenv/rbenv#installation)
+
+**Next install Ruby and the gems required for testing.**
+
+```sh
+rbenv install
+gem install bundler
```
kevindew, line 15: This step should be unnecessary given we're using Ruby >= 2.6

CONTRIBUTING.md
```diff
@@ -2,6 +2,30 @@
...
+bundle install
+```
+
+**Now use one of the following commands to run tests.**
```
kevindew, line 19: Do we need to give people both options? could leave the `bundle exec rspec` equivalent as an advanced step that people can find themselves.

README.md
```diff
@@ -12,7 +12,42 @@ The aim of govuk-docker is to make it easy to develop any GOV.UK app. It achieve
 
 ## Installation
 
-[Instructions for how to install and setup govuk-docker are here.](docs/installation.md)
+**First make sure you have the following dependencies.**
+
+- [brew](https://brew.sh/)
+- [git](https://git-scm.com)
+- A directory `~/govuk` in your home directory
```
kevindew, line 19: Minor: the tilde already implies this is in your home directory. If we're wanting to make it clearer for those who don't understand tilde we could just switch it to `govuk`.

README.md
```diff
@@ -12,7 +12,42 @@ The aim of govuk-docker is to make it easy to develop any GOV.UK app. It achieve
...
+
+**Next, add the following line to your bash/zsh config.**
+
+```
+export PATH=$PATH:~/govuk/govuk-docker/exe
+```
```
kevindew, line 25: You may want a step telling them to run `source ~/.zshrc` or `source ~/.bashrc` so the path is updated for the session.

kevindew (commented):
Looks great - nice job Ben. I left a few comments before I realised that they weren't on new items but things that had moved. Happy to defer resolving them if they're broadening the scope.

CONTRIBUTING.md
benthorner: Oh yes, I keep forgetting about that. Removed.

README.md
benthorner: Good point. I've reworded this.

CONTRIBUTING.md
benthorner: That's fair - someone can look at the Makefile to understand the whole test suite. I've reworded this section.

README.md
benthorner: Good idea. I've added some pointers about that.

benthorner:
Thanks for the review @kevindew! I think I've addressed all your comments. Could you take another look?

kevindew (approved):
LGTM 👍

PR #434, opened 2020-11-17
Revise troubleshooting guidance

benthorner:
https://trello.com/c/i3Y2vAua/188-documentation-on-how-to-troubleshoot-apps-failing-to-start-in-govuk-docker

This PR is a review of the general troubleshooting guidance
for apps, which we know is unhelpful for some people.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -65,7 +65,11 @@ scutil --dns
 # reach    : 0x00030002 (Reachable,Local Address,Directly Reachable Address)
 ```
 
-You can also look at the command in `bin/setup` to see what's changing.
+* Re-run the setup script and look for errors
+
+```
+bin/setup
+```
 
 ## Troubleshoot: Installation
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,8 +1,6 @@
 # Troubleshooting
 
-## Troubleshoot: Usage
-
-### Diagnose common issues when setting up GOV.UK Docker
+## Diagnose common issues when setting up GOV.UK Docker
 
 Run the following command in `~/govuk/govuk-docker`. Since this script makes use of Ruby Gems, you will need to [install some additional dependencies](../CONTRIBUTING.md#testing) in order to do this.
 
@@ -10,7 +8,7 @@ Run the following command in `~/govuk/govuk-docker`. Since this script makes use
 bin/doctor
 ```
 
-### Diagnose general issues with a project/app not working
+## Diagnose general issues with a project/app not working
 
 * Make sure you run all commands via GOV.UK Docker.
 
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -71,17 +69,7 @@ scutil --dns
...
 
-## Troubleshoot: Installation
-
-### Diagnose common issues when setting up GOV.UK Docker
-
-Run the following command in `~/govuk/govuk-docker`. Since this script makes use of Ruby Gems, you will need to [install some additional dependencies](../CONTRIBUTING.md#testing) in order to do this.
-
-```
-bin/doctor
-```
```
ChrisBAshton, line 66: May be worth linking to https://github.com/alphagov/govuk-docker/issues/435 in your commit, as I wouldn't have known why this section was removed if I hadn't already read your issue..

docs/troubleshooting.md
```diff
@@ -12,21 +12,33 @@ bin/doctor
...
+```
+
+* Try cleaning up and running your command again.
+
+```
+# stop all apps and their dependencies
+govuk-docker down
 
-# try clearing all containers / volumes
-govuk-docker rm -sv
```
ChrisBAshton, line 29: I know you want to encourage people to diagnose and fix errors 'properly', but sometimes if you're in a bind and need to get something working quickly, this is just the easiest option. I'd really like you to leave it in, even if it's caveated with "You really shouldn't have to do this...!"

ChrisBAshton (commented):
Have had a read and have a couple of suggestions below.

An aside: one note I made from yours & Huw's talk the other week was `govuk-docker-up app-live` for looking at live data. That may not be the exact command or purpose, but it was new to me and I wrote it down - can't see any further detail about it in the repo though. Is it documented somewhere? :)

docs/troubleshooting.md
benthorner: I agree there is a need to have a clean slate. The `govuk-docker down` command I've suggested is actually equivalent to this, but has the advantage of being easier to explain and write down.

Does that help address your concern?

docs/troubleshooting.md
benthorner: Ah, so I was wondering whether to be explicit about this in the commit: the section still exists at the top - it's just de-duplicated. Depending on what happens with #435, we may yet remove or edit it, but I agree that would be premature here.

Is that OK?

benthorner:
Thanks for the review @ChrisBAshton. I think I may have addressed your comments with a bit more explanation, but let me know if you'd like me to clarify the commits, for example.

The `app-live` stack has a slightly awkward history, because I think some devs still rely on the `startup.sh` scripts in each repo, which replicate the functionality. Anyway, the functionality in GOV.UK Docker is documented [in the README](https://github.com/alphagov/govuk-docker#the-app--stacks), and should  also now be a little more prominent, since I rewrote it.

docs/troubleshooting.md
ChrisBAshton: Ah, didn't spot that. Makes sense. Thanks

docs/troubleshooting.md
ChrisBAshton: Sure, thanks for explaining.

ChrisBAshton (approved):
LGTM 👍

PR #436, opened 2020-11-18
ADR to document shared Dockerfile approach

kevindew:
This is a decision that was made long ago and I felt we should document
this before we add further ADRs. Hopefully I've remembered it all.

I somewhat used:
https://github.com/alphagov/govuk-docker/blob/master/docs/adr/0001/DECISIONS.md#docker-dockerfiles
to inform this, but I skipped of the reasons that I didn't find
particularly convincing.

### DIFF
--- a/docs/adr/0002-docker-images-for-ruby-applications.md
+++ b/docs/adr/0002-docker-images-for-ruby-applications.md
@@ -1,0 +1,113 @@
+# 2. Docker images for Ruby Applications
+
+Date: 2020-11-17
+
+## Context
+
+In order to run GOV.UK applications in a Docker environment we need
+[Docker images][]. GOV.UK had previously created Docker images for various
+applications to enable the [publishing-e2e-tests][] suite, which runs a
+containerised version of the GOV.UK publishing stack. These images were
+defined with a Dockerfile in the root of a project's repository
+([example][example-dockerfile]) and, as they are the only Dockerfile for a
+project, were essentially the defacto Docker images for GOV.UK applications.
+
+Attempting to reuse these images in the GOV.UK Docker development environment
+revealed a number of problems:
+
+1. Any time a Ruby project's Gem dependencies changed the image would need to
+   be rebuilt which [required re-installing][gem-install] every single gem,
+   which could be a slow process.
+2. As Gem dependencies were stored within a project's image no Gems could be
+   shared across applications - this meant that the initialisation process
+   for multiple GOV.UK apps could be frustratingly slow with the same Gems
+   installed in each different image, it also resulted in high disk usage.
+3. The images were only exercised in publishing-e2e-tests which ran them
+   [in production mode][production], this meant they didn't necessarily have
+   sufficient dependencies for development usage  (e.g. Google Chrome for UI
+   testing).
+4. The images [embedded the application's files][embed-files] - this meant
+   that by default any edits to files would require the image to be rebuilt.
+
+Thus we considered whether it would be appropriate to modify these images to
+solve the problems or to create different images for GOV.UK Docker.
+
+[Docker images]: https://docs.docker.com/get-started/overview/#images
+[publishing-e2e-tests]: https://github.com/alphagov/publishing-e2e-tests
+[example-dockerfile]: https://github.com/alphagov/publishing-api/blob/8b6bc39a3f6b73c2c3e73def6a1c2a6e7687d789/Dockerfile
+[embed-files]: https://github.com/alphagov/specialist-publisher/blob/33a6f3a5d7c8b5ea79d0cd5a8232ed40391db1ef/Dockerfile#L26
+[gem-install]: https://github.com/alphagov/government-frontend/blob/51d86566de33c122aea20f30bcc0fd0d36572752/Dockerfile#L16
+[production]: https://github.com/alphagov/publishing-e2e-tests/blob/3c0e50f66c2c61a7bb98eb07cacd4fcf6233fb51/docker-compose.yml#L14
+
+## Decision
+
+We decided that we should not reuse the existing images and that we would
+resolve the problems with a different image approach for GOV.UK Docker. While
+some of the problems could be resolved we reasoned that the existing images
+represent [standard industry practice][docker-rails] for web application images
+and that this practice was optimised for running a production instance of an
+application. We felt that the needs of production and development were
+sufficiently distinct that it would be simpler to have separate images than try
+to consolidate.
+
+We instead [created a Dockerfile][base-dockerfile] that is stored in
+GOV.UK Docker and this is used by the majority of GOV.UK Docker Ruby projects.
+This base Dockerfile contains a collection of common dependencies used by a
+significant number of projects. Projects that have additional dependencies
+have their [own Dockerfiles][whitehall-dockerfile]. This allowed us to ensure
+that projects had their necessary development dependencies.
+
+We resolved the problems for Gem installations with a [shared Docker
+mount][home-volume]. Using this allowed the installation, and thus updating,
+of Ruby versions and Gems to be done [outside a Docker build
+process][ruby-install], meaning these tasks can be achieved without rebuilding
+the image. This shared mount is used across projects and allows Ruby and Gem
+dependencies to be reused.
+
+We resolved the problem of embedded files with a [mount for the ~/govuk
+directory][govuk-mount]. This mount allows containers to access application
+code and for this code to be modified without requiring an image rebuild. Using
+this broad directory allows projects to access other projects when necessary
+which assists working with [govuk-content-schemas][] and
+[local versions of Gems][local-gems].
+
+[docker-rails]: https://docs.docker.com/compose/rails/
+[base-dockerfile]: https://github.com/alphagov/govuk-docker/blob/b2fb90dd62a7579976cc3adaa9c783cd92cbd7e7/Dockerfile.govuk-base
+[whitehall-dockerfile]: https://github.com/alphagov/govuk-docker/blob/b2fb90dd62a7579976cc3adaa9c783cd92cbd7e7/projects/whitehall/Dockerfile
+[home-volume]: https://github.com/alphagov/govuk-docker/blob/b2fb90dd62a7579976cc3adaa9c783cd92cbd7e7/docker-compose.yml#L4
+[ruby-install]: https://github.com/alphagov/govuk-docker/blob/b2fb90dd62a7579976cc3adaa9c783cd92cbd7e7/Makefile#L27-L31
+[govuk-mount]: https://github.com/alphagov/govuk-docker/blob/b2fb90dd62a7579976cc3adaa9c783cd92cbd7e7/projects/publishing-api/docker-compose.yml#L11
+[govuk-content-schemas]: https://github.com/alphagov/govuk-content-schemas
+[local-gems]: https://bundler.io/man/gemfile.5.html#PATH
+
+## Status
+
+Accepted
+
+## Consequences
+
+With most GOV.UK Docker projects lacking an individual Dockerfile there is
+substantially less Docker configuration to manage and there is a reduced risk
+of inconsistencies. This also reduces the difficulty in adding new projects to
+GOV.UK Docker.
+
+The containers in GOV.UK Docker lack the property of being a [unit of software
+that packages up all code and dependencies][docker-container] which may be
+surprising. However, it is unclear that container principles can be easily
+applied to Ruby-on-Rails development environments without the problems we
+experienced.
+
+By storing the Dockerfile in GOV.UK Docker there is only a very weak coupling
+between the minimal set of system packages installed on the image, and those
+used by the application. This will mean that changes to system packages need
+coordinating in both the application repository and GOV.UK Docker. This is
+however a rare occurrence, due to the minimal number of system packages in
+the Dockerfile used by GOV.UK Docker.
+
+GOV.UK will have multiple Dockerfiles for the same application raising a risk
+of confusion for developers. While [there remains][remove-e2e] a Dockerfile in
+the root of repositories developers may expect that this is somewhat related
+to the GOV.UK Docker project.
+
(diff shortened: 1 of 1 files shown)


benthorner:
Thanks for raising this @kevindew 🥇. Several people have asked about this, and it will be helpful to have a single, dedicated document to point them to, as (I think) the reasons are all still valid.

I found myself a bit confused between the context, decision and consequences, as we seem to be arguing for the decision in all three. Perhaps the part to focus on is the last big paragraph of the context, where we're arguing in terms of time and space, but quite generically. I also think we're missing an argument against the E2E images, since they're optimised for production. Rather than make loads of suggestions, it may help to summarise the steps I took to make the decisions I did, which were all prompted by me wanting to do a specific action and finding it difficult.

- The [first decision](https://github.com/alphagov/govuk-docker/blob/master/docs/adr/0001/DECISIONS.md#docker-dockerfiles) was to avoid using the existing E2E images. The E2E images are fundamentally not suitable for development, because they are optimised for a production environment, so [lack test dependencies and packages](https://github.com/alphagov/static/blob/master/Dockerfile#L19). Making this decision meant we needed to build custom images for each app, that would at least _work_ in development.

- The [next decision](https://github.com/alphagov/govuk-docker/blob/master/docs/adr/0001/DECISIONS.md#docker-lockfiles) was to optimise the images for development. This is less around storage (although that's a good supporting argument) and more around convenience of development: no one wants to wait to reinstall all of their gems every time someone merges a Dependabot PR (that was the exact frustration I had).

- The final decision was really a no-brainer: keep the single, DRY Dockerfile separate to the rest of GOV.UK. This was partly for the reason you described - it makes changes easier - but also historically because GOV.UK Docker was built to have minimal-to-no impact on other GOV.UK repos, which has the advantage of very low coupling.

There are also a minor decision related to the second one (again, optimising for development): [govuk-docker uses itself as the build context for all images](https://github.com/alphagov/govuk-docker/blob/master/docs/adr/0001/DECISIONS.md#docker-context). Anyway, I hope that's helpful. I appreciate I made all these decisions on my own, and despite them being written, it's hard to pick up on them all and to be expected that not all of them will be commonly agreed. Let me know if you'd like to chat about anything here!

kevindew:
Thanks Ben, that was really helpful. I've re-structured this document following your feedback. Let me know your thoughts

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,96 @@
...
+The vast majority of built GOV.UK Docker images do not contain all of the
+dependencies an application requires. Instead, [additional build steps][] will
+be needed to run an application.
+
+By storing the Dockerfile in GOV.UK Docker there is not a coupling between
+a Dockerfile and a version of an applications code. This will mean that changes
+to system dependencies need co-ordinating in both the application repositiory
+and GOV.UK Docker. This is however a rare occurrence.
+
+GOV.UK will have different Dockerfiles for development and production runtimes,
```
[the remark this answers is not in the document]
kevindew: Yeah, this point is also a bit broader than E2E tests as [RFC-117](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-117-container-runtime.md) would mean that in time we'll have different production containers whether the e2e ones are reused or not.

Replatforming are currently using them but they know about the deletion, I think they'll build their own following that - they'll have some fun things to consider there.

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,96 @@
+# 2. Docker images for Ruby Applications
+
+Date: 2020-11-17
+
+## Context
+
+GOV.UK is built as a [microservice architecture][] comprising of over 50
+applications, the majority of these being built built with Ruby on Rails.
+A challenge for developing on GOV.UK is that you frequently need to
```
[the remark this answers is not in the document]
kevindew: I think this veered into a far too broad context that I should have edited down much earlier. I've cut it significantly back now.

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,104 @@
...
+
+Date: 2020-11-17
+
+## Context
+
+In order to run GOV.UK applications in a Docker environment we need
+[Docker images][]. GOV.UK had previously created Docker images for various
+applications to enable the [publishing-e2e-tests][] suite, which runs a
+containerised version of the GOV.UK publishing stack. These images were
+defined with a Dockerfile in the root of a projects repository
```
benthorner, line 11:
```suggestion
defined with a Dockerfile in the root of a project's repository
```

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,104 @@
...
+Attempting to reuse these images in the GOV.UK Docker development environment
+revealed a number of problems:
+
+1. The images [embedded the applications files][embed-files] - this meant that
+   any time a file is changed the image needed to be rebuilt, which is a
+   frequent occurrence when developing.
+2. Any time a Ruby project's Gem dependencies changed the image would need to
+   be rebuilt which [required re-installing][gem-install] every single gem,
+   which could be a slow process.
+3. As Gem dependencies were stored within a projects image no Gems could be
```
benthorner, line 24:
```suggestion
3. As Gem dependencies were stored within a project's image no Gems could be
```

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,104 @@
...
+applications to enable the [publishing-e2e-tests][] suite, which runs a
+containerised version of the GOV.UK publishing stack. These images were
+defined with a Dockerfile in the root of a projects repository
+([example][example-dockerfile]) and, as they are the only Dockerfile for a
+project, were essentially the defacto Docker images for GOV.UK applications.
+
+Attempting to reuse these images in the GOV.UK Docker development environment
+revealed a number of problems:
+
+1. The images [embedded the applications files][embed-files] - this meant that
```
benthorner, line 18: This is relatively easy to overcome by mounting the local files over the top of those in the volume. We could say that, while there are workarounds for some of the issues here, on the whole using the existing images was impractical.

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,104 @@
...
+2. Any time a Ruby project's Gem dependencies changed the image would need to
+   be rebuilt which [required re-installing][gem-install] every single gem,
+   which could be a slow process.
+3. As Gem dependencies were stored within a projects image no Gems could be
+   shared across applications - this meant that the initialisation process
+   for multiple GOV.UK apps could be frustratingly slow with the same Gems
+   installed in each different image, it also resulted in high disk usage.
+4. The images were only exercised in publishing-e2e-tests which ran them
+   [in production mode][production], this meant they didn't necessarily have
+   sufficient dependencies for development usage.
```
benthorner, line 30:
```suggestion
   sufficient dependencies for development usage (e.g. the Chrome browser for UI testing).
```

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,104 @@
...
+it would be simpler to have separate images than try to consolidate.
+
+We instead [created a Dockerfile][base-dockerfile] that is stored in
+GOV.UK Docker and this is used by the majority of GOV.UK Docker Ruby projects.
+This base Dockerfile contains a collection of common dependencies used by a
+significant number of projects. Projects that have additional dependencies
+have their [own Dockerfiles][whitehall-dockerfile].
+
+This enabled us to resolve the aforementioned problems. The need to embed
+files is replaced by a [shared Docker mount][home-volume] that provides the
```
benthorner, line 59: Are there two points here?

- We use a shared mount to install dependencies.
- We use a shared mount for app files.

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,104 @@
...
+## Consequences
+
+With most GOV.UK Docker projects lacking an individual Dockerfile there is
+substantially less Docker configuration to manage and there is a reduced risk
+of inconsistencies. This also reduces the difficulty in adding new projects to
+GOV.UK Docker.
+
+The containers in GOV.UK Docker lack the property of being a [unit of software
+that packages up all code and dependencies][docker-container] which may be
+surprising. However, it is unclear how that container principles can be easily
```
benthorner, line 87:
```suggestion
surprising. However, it is unclear that container principles can be easily
```

benthorner (commented):
Much clearer, thanks @kevindew 👍. There's still one part of the decision that I think could with expanding on, and we could perhaps point out that we (/I) did try to work around the problems with the E2E images. I've left comments about those things, and otherwise I think it's good to go. 

Do you think it's worth asking someone with less context to review it too?

docs/adr/0002-docker-images-for-ruby-applications.md
kevindew: Oh sure, I hadn't realised that - but seems pretty obvious now 🤦 . So rather than a total pain it's more of a redundant action that can get in a bit of a mess for files like Gemfile or package.json which impact container and runtime.

I've moved this concern from the top to the bottom of list to reflect it not being a deal breaker.

docs/adr/0002-docker-images-for-ruby-applications.md
kevindew: I've rephrased this paragraph now.

kevindew:
Thanks @benthorner - do you want to check the last commit?

Yes, let's see if @huwd would like to review?

docs/adr/0002-docker-images-for-ruby-applications.md
benthorner: Thanks. I realise I wasn't specific above. This is actually a separate mount:

```
${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
```

Note that a minor problem this solves is the need to have files for _other_ projects available, in order to test against gems (+ govuk-content-schemas). I'm not sure we need to go to this level of detail here, but I thought I'd mention it in case it fits nicely into any changes we make here.

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,106 @@
...
+Thus we considered whether it would be appropriate to modify these images to
+solve the problems or to create different images for GOV.UK Docker.
+
+[Docker images]: https://docs.docker.com/get-started/overview/#images
+[publishing-e2e-tests]: https://github.com/alphagov/publishing-e2e-tests
+[example-dockerfile]: https://github.com/alphagov/publishing-api/blob/8b6bc39a3f6b73c2c3e73def6a1c2a6e7687d789/Dockerfile
+[embed-files]: https://github.com/alphagov/specialist-publisher/blob/33a6f3a5d7c8b5ea79d0cd5a8232ed40391db1ef/Dockerfile#L26
+[gem-install]: https://github.com/alphagov/government-frontend/blob/51d86566de33c122aea20f30bcc0fd0d36572752/Dockerfile#L16
+[production]: https://github.com/alphagov/publishing-e2e-tests/blob/3c0e50f66c2c61a7bb98eb07cacd4fcf6233fb51/docker-compose.yml#L14
+[volume]
```
benthorner, line 41:
```suggestion
```

docs/adr/0002-docker-images-for-ruby-applications.md
kevindew: Oh yes, of course. Will sort.

kevindew: It's updated now.

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,113 @@
...
+   which could be a slow process.
+2. As Gem dependencies were stored within a project's image no Gems could be
+   shared across applications - this meant that the initialisation process
+   for multiple GOV.UK apps could be frustratingly slow with the same Gems
+   installed in each different image, it also resulted in high disk usage.
+3. The images were only exercised in publishing-e2e-tests which ran them
+   [in production mode][production], this meant they didn't necessarily have
+   sufficient dependencies for development usage  (e.g. Google Chrome for UI
+   testing).
+4. The images [embedded the applications files][embed-files] - this meant
```
benthorner, line 29:
```suggestion
4. The images [embedded the application's files][embed-files] - this meant
```

docs/adr/0002-docker-images-for-ruby-applications.md
```diff
@@ -0,0 +1,113 @@
...
+
+We resolved the problems for Gem installations with a [shared Docker
+mount][home-volume]. Using this allowed the installation, and thus updating,
+of Ruby versions and Gems to be done [outside a Docker build
+process][ruby-install], meaning these tasks can be achieved without rebuilding
+the image. This shared mount is used across projects and allows Ruby and Gem
+dependencies to be reused.
+
+We resolved the problem of embedded files with a [mount for the ~/govuk
+directory][govuk-mount] which is expected to project repositories. This mount
```
benthorner, line 68:
```suggestion
directory][govuk-mount]. This mount
```

I'm not sure what this meant, and what follows makes sense without it.

benthorner (approved):
Thanks again @kevindew 🇮🇹

docs/adr/0002-docker-images-for-ruby-applications.md
kevindew: Ah I missed a bit. It was meant to say "which is expected to store projects' repositories" - I've gone with your removed suggestion.

kevindew:
@huwd would you mind reviewing this please? Re: https://github.com/alphagov/govuk-docker/pull/436#pullrequestreview-534626405

thomasleese (approved):
Makes sense to me!

PR #437, opened 2020-11-23
ADR to document volume usage and running as root

kevindew:

This documents some of the decisions introduced in https://github.com/alphagov/govuk-docker/pull/394

### DIFF
--- a/docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
+++ b/docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
@@ -9,12 +9,12 @@ A common problem developers have experienced with GOV.UK Docker is that the
 common that users [were served 504 responses][504-fix] on initial attempts to
 start an application. This slowness in GOV.UK Docker appeared to be an
 amplification of [existing performance][govuk-frontend-performance] issues,
-however when GOV.UK apps began upgrading
-from the end-of-life Ruby Sass to Sassc we began to see problems that pointed
-to a performance issue within GOV.UK Docker.
+however when GOV.UK apps began upgrading from the end-of-life Ruby Sass to
+Sassc we also began to see problems that pointed to a performance issue
+within GOV.UK Docker.
 
 Sassc is known to [significantly improve the compilation time][sassc-speed] of
-Sass compared to Ruby Sass. When we updated [Content Publisher][] to Sassc
+Sass compared to Ruby Sass. When we updated [Content Publisher][] to Sassc we
 found that the [application.scss][] file compiled in 20% of the time (~20s to
 4s) when running natively on macOS. However when this was run on GOV.UK Docker
 the performance was significantly less impressive, compiling in 62% of the
@@ -22,15 +22,19 @@ time (~31s to ~19s).
 
 We discovered that the source of this performance issue was the
 [performance overhead of reading][docker-mac-read] from the
-[shared mount][govuk-mount] of application files on macOS. We considered a number
-of options of how this could be improved:
+[shared mount][govuk-mount] of application files on macOS. In particular we
+found that there were directories, `tmp` and `node_modules`, which could be
+frequently accessed during Sass compilation, but did not contain
+files that a developer would need to edit during development. Bearing this in
+mind, we considered a number of options as to how we could improve the
+performance:
 
-1. Use [Docker volumes][volume] to store non-application code (such as caches or
-   dependencies) to benefit from faster I/O access.
-2. [Configure GOV.UK Docker to use NFS][nfs-docker] for improved I/O access to
-   shared mounts.
-3. Use [docker-sync.io](http://docker-sync.io/) as a means for improved I/O
-   access to shared mounts.
+1. Use [Docker volumes][volume] to store non-application code (such as `tmp`
+   and `node_modules`) to benefit from faster I/O access.
+2. [Configure GOV.UK Docker to use NFS][nfs-docker] (an alternative driver for
+   shared mounts) for improved I/O access to shared mounts.
+3. Use [docker-sync.io](http://docker-sync.io/) (a file-syncing daemon) as a
+   means for improved I/O access to shared mounts.
 
 [asset pipeline]: https://guides.rubyonrails.org/asset_pipeline.html
 [504-fix]: https://github.com/alphagov/govuk-docker/pull/310
@@ -45,10 +49,12 @@ of options of how this could be improved:
 
 ## Decision
 
-We decided to use Docker volumes for non-application code.
+We decided to mount project-specific Docker volumes on top of the existing
+`~/govuk` shared mount, in order to store non-application code (caches or
+dependencies).
 
-When trying this on Content Publisher with the `tmp` directory and
-`node_modules` directory as Docker volumes we saw a significant improvement
+When trying this on Content Publisher, with the `tmp` directory and
+`node_modules` directory as Docker volumes, we saw a significant improvement
 in Sass compilation time on macOS:
 
 | Environment                                           | Ruby Sass | Sassc | Decrease |
@@ -64,7 +70,9 @@ machine to create the NFS directory which would complicate installation and
 usage of GOV.UK Docker.
 
 We decided not to use docker-sync.io as that required additional dependencies
-and a large amount of additional configuration.
+and a large amount of additional configuration. We felt that, given Docker
+volumes provided us with near native performance, it would not be worth
+pursuing an unconventional, configuration heavy, approach.
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Context
+
+A common problem developers have experienced with GOV.UK Docker is that the
+[asset pipeline][] compilation steps can be slow, to the point that it was once
+common that users [were served 504 responses][504-fix] on initial attempts to
+start an application. This slowness in GOV.UK Docker appeared to be an
+amplification of [existing performance][govuk-frontend-performance] issues,
+however when GOV.UK apps began upgrading
+from the end-of-life Ruby Sass to Sassc we began to see problems that pointed
```
benthorner, line 13:
```suggestion
from the end-of-life Ruby Sass to Sassc we also began to see problems that pointed
```

(They were two separate issue, after all.)

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+to a performance issue within GOV.UK Docker.
+
+Sassc is known to [significantly improve the compilation time][sassc-speed] of
+Sass compared to Ruby Sass. When we updated [Content Publisher][] to Sassc
+found that the [application.scss][] file compiled in 20% of the time (~20s to
```
benthorner, line 18:
```suggestion
we found that the [application.scss][] file compiled in 20% of the time (~20s to
```

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+time (~31s to ~19s).
+
+We discovered that the source of this performance issue was the
+[performance overhead of reading][docker-mac-read] from the
+[shared mount][govuk-mount] of application files on macOS. We considered a number
+of options of how this could be improved:
+
+1. Use [Docker volumes][volume] to store non-application code (such as caches or
+   dependencies) to benefit from faster I/O access.
+2. [Configure GOV.UK Docker to use NFS][nfs-docker] for improved I/O access to
```
benthorner, line 30:
```suggestion
2. [Configure GOV.UK Docker to use NFS][nfs-docker] (an alternative driver for shared mounts) for improved I/O access to
```

Minor: we could give a bit more indication what this alternative means in practice.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+   shared mounts.
+3. Use [docker-sync.io](http://docker-sync.io/) as a means for improved I/O
```
benthorner, line 32:
```suggestion
3. Use [docker-sync.io](http://docker-sync.io/) (a file-syncing daemon) as a means for improved I/O
```

Minor: we could give a bit more indication what this alternative means in practice.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+[Content Publisher]: https://github.com/alphagov/content-publisher
+[application.scss]: https://github.com/alphagov/content-publisher/blob/2f91d43bcd1eb1f2d30f9d1ff9e556cd4d0de2a6/app/assets/stylesheets/application.scss
+[docker-mac-read]: https://github.com/docker/for-mac/issues/77
+[govuk-mount]: https://github.com/alphagov/govuk-docker/blob/cb124a3a1d3353e777d4f777d77f03f93415d415/projects/content-publisher/docker-compose.yml#L11
+[volume]: https://docs.docker.com/storage/volumes/
+[nfs-docker]: https://www.jeffgeerling.com/blog/2020/revisiting-docker-macs-performance-nfs-volumes
+
+## Decision
+
+We decided to use Docker volumes for non-application code.
```
benthorner, line 48:
```suggestion
We decided to mount project-specific Docker volumes on top of the existing `~/govuk` shared mount, in order to store non-application code (caches or dependencies).
```

We can be more specific here, since we already use Docker volumes for non-application code (e.g. rbenv). Some of the decision points:

- It's conceptually more confusing to have volumes on top of volumes.

- It's more config overhead to specify volumes or a per-project basis.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+| Environment                                           | Ruby Sass | Sassc | Decrease |
+|-------------------------------------------------------|-----------|-------|----------|
+| Native macOS                                          | ~20s      | ~4s   | 80%      |
+| GOV.UK Docker (with shared `~/govuk` mount)           | ~31s      | ~19s  | 38%      |
+| GOV.UK Docker (with `tmp` and `node_modules` volumes) | ~16s      | ~4s   | 75%      |
+
+We saw less impressive improvements by switching the shared mount to use NFS.
+This performed approximately 3 times slower than the Docker volume approach.
+This approach was also not desirable as it required configuration on the host
+machine to create the NFS directory which would complicate installation and
```
benthorner, line 63: This is quite a detailed explanation relative to the intro in the previous section. I've suggested we add a little more to explain it there, to try and match the depth here.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+usage of GOV.UK Docker.
+
+We decided not to use docker-sync.io as that required additional dependencies
```
benthorner, line 66: After I read this I still wondered: but what if it was worth it? Perhaps we can make the argument more explicit: the shared volume approach is _almost_ as good a native, and requires only a small amount of standard config; conversely, the other two approaches require significantly more setup.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+Sassc is known to [significantly improve the compilation time][sassc-speed] of
+Sass compared to Ruby Sass. When we updated [Content Publisher][] to Sassc
+found that the [application.scss][] file compiled in 20% of the time (~20s to
+4s) when running natively on macOS. However when this was run on GOV.UK Docker
+the performance was significantly less impressive, compiling in 62% of the
+time (~31s to ~19s).
+
+We discovered that the source of this performance issue was the
+[performance overhead of reading][docker-mac-read] from the
+[shared mount][govuk-mount] of application files on macOS. We considered a number
```
benthorner, line 25:
```suggestion
[shared mount][govuk-mount] of application files on macOS, specifically the `tmp` and `node_modules` directories, which are both heavily in Sass compilation. We considered a number
```

These directories appear rather suddenly in the next section. We should indicate why they're relevant - the `node_modules` one is reasonably surprising, I think.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Status
+
+Accepted
+
+## Consequences
+
+We have added Docker volumes to the configuration of all GOV.UK Rails projects
+so that the `tmp` directory is a volume. This provides improved I/O access to
+this directory and offers performance benefits outside of Sass compilation.
```
benthorner, line 77:
```suggestion
this directory and offers performance benefits outside of Sass compilation (e.g. [for ActiveStorage](https://github.com/alphagov/content-publisher/blob/2f91d43bcd1eb1f2d30f9d1ff9e556cd4d0de2a6/config/storage.yml#L3).
```

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+To ensure this is applied consistently we have [added a test][tmp-test].
+
+For projects that make use of npm modules we will add a `node_modules` volume
```
benthorner, line 80:
```suggestion
For projects that make use of npm modules we have added a `node_modules` volume
```

Sudden change in tense here.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+to the project. This will mean that these installs can no longer be shared
+between a host machine and GOV.UK Docker - which is beneficial for any native
```
benthorner, line 82: We should acknowledge the (unlikely) potential for confusion, since some of the directories seen on the local machine will no longer reflect those in use by the container. In particular, anyone who previously got away with running `npm/yarn install` will now have to run `govuk-docker-run npm/yarn install`.

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+modules that are compiled for the installing OS. This is also
```
benthorner, line 83:
```suggestion
modules by removing the potential for compatibility issues when compiling modules that were coupled to the specific OS. This is also
```

Minor: it took a couple of readings to understand this (I'm not aware it's been a problem).

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+[enforced by a test][node-test].
+
+In order to make increased usage of Docker volumes we had to change the base
+Dockerfile from [running as a specific user][specific-user] to the Docker
+default of running as root. This is because it is
+[difficult and complex][volume-root] to define volumes as owned by a different
+user. This changes meant that all users of GOV.UK Docker needed to rebuild
+their images to accommodate this change. It is also increased the
+[complexity in running Google Chrome][chrome-root] as part of an applications
```
benthorner, line 92:
```suggestion
[complexity in running Google Chrome][chrome-root] as part of an application's
```

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
```diff
@@ -0,0 +1,102 @@
...
+between a host machine and GOV.UK Docker - which is beneficial for any native
+modules that are compiled for the installing OS. This is also
+[enforced by a test][node-test].
+
+In order to make increased usage of Docker volumes we had to change the base
+Dockerfile from [running as a specific user][specific-user] to the Docker
+default of running as root. This is because it is
+[difficult and complex][volume-root] to define volumes as owned by a different
+user. This changes meant that all users of GOV.UK Docker needed to rebuild
+their images to accommodate this change. It is also increased the
```
benthorner, line 91:
```suggestion
their images to accommodate this change. It is also slightly increased the
```

(Arguably it's actually _less_ complex, since we've removed all the surprising `privileged` flags everywhere.)

benthorner (commented):
Really helpful summary @kevindew. Thanks for the effort summarising such a complex and lengthy change so succinctly. The motivation is really clear. I've got a couple of more major comments around clarifying things like the decision and alternatives, especially for people who aren't so familiar with all this - I can see us pointing to this when someone turns up late to the migration party!

docs/adr/0003-use-docker-volumes-for-improved-macos-performance.md
kevindew: Argh, when will I ever get my possessives correct.

kevindew:
Thanks @benthorner, lots of good points. I've tried to action them all in the most recent commit.

benthorner (approved):
🇱🇮 

I've had another read and it looks good to me - thanks for addressing all the comments.

PR #438, opened 2020-11-23
Document database version number decision

kevindew:

This documents the database service version decisions implemented in https://github.com/alphagov/govuk-docker/pull/394

### DIFF
--- a/docs/adr/0004-use-versioned-database-services.md
+++ b/docs/adr/0004-use-versioned-database-services.md
@@ -1,0 +1,58 @@
+# 1. Use versioned database services
+
+Date: 2020-11-23
+
+## Context
+
+GOV.UK Docker has services defined for various databases, such as PostgreSQL,
+MongoDB and Elasticsearch. These have their data persisted in
+[Docker volumes][volume] defined [for each service][defined-volumes].
+
+When GOV.UK Docker updated from [MongoDB 2.4 to 3.6][mongodb-update] users
+began receiving cryptic errors when trying to run applications that depended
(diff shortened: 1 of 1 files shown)


docs/adr/0004-use-versioned-database-services.md
```diff
@@ -0,0 +1,53 @@
...
+## Status
+
+Accepted
+
+## Consequences
+
+When we next update the version of a database service in GOV.UK Docker users
+will find they have an empty volume for that database rather than one with
+incompatible data. This will make it easier for developers to work with the
+service as they can run the familiar `make <service>` task to re-initialise it.
```
benthorner, line 46:
```suggestion
service as they can run the familiar `make <service>` task to re-initialise it (or refill it with replica data, using [the replication scripts](https://github.com/alphagov/govuk-docker/tree/master/bin)).
```

docs/adr/0004-use-versioned-database-services.md
```diff
@@ -0,0 +1,53 @@
...
+
+Advanced users will have the option to maintain their existing data by
+migrating data from the old volume to the new volume.
```
benthorner, line 49:
```suggestion
migrating data from the old volume to the new volume manually. We decided not to provide guidance for this, as we think most local data is low value and not worth the effort to maintain.
```

Minor: we can be more specific about how this would work in practice.

benthorner (approved):
Nice - persuasive extra/future argument to end with too ⭐

kevindew:
Thanks @benthorner 👍  - I'll merge once I've got the other 2 ADRs through.

PR #439, opened 2020-11-27
Add database versions to replication scripts

edwardkerry:
As govuk-docker is now [using versioned database services](https://github.com/alphagov/govuk-docker/blob/master/docs/adr/0004-use-versioned-database-services.md) the data replication scripts need to include these versions.


benthorner (approved):
Oops, thanks for fixing ⭐. It's a bit awkward that we may in future have some apps whose DB is different to the one here, but we'll just have to address that when it happens.

thomasleese (approved):

PR #440, opened 2020-12-03
Fix broken link to installation

alex-ju:
Possibly a result of moving doc files around


benthorner (approved):
Oops - thanks for spotting and fixing ⭐.

PR #442, opened 2020-12-08
Add yarn/node_modules for repos that use yarn for linting

kevindew:
Trello: https://trello.com/c/MeG1zc2m/195-roll-out-stylelint-config-gds-across-govuk

We're gradually migrating GOV.UK apps to do JS and Sass linting via node
based tools. This is a collection of apps that have been migrated to
this approach. This change adds yarn as part of the applications setup
and use a node_modules volume for the modules.

I've prepared this to be merged after I merge in the changes to the various apps that I have open PRs for. It should be pretty benign for apps to not have this change, it'll just mean installing yarn dependencies manually.

### DIFF
--- a/projects/calculators/Makefile
+++ b/projects/calculators/Makefile
@@ -1,1 +1,2 @@
 calculators: bundle-calculators static content-store router
+	$(GOVUK_DOCKER) run $@-lite yarn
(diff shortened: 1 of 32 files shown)


thomasleese (approved):
Makes sense to me 👍

benthorner (approved):
Looks good to me too 👍

PR #448, opened 2021-01-14
Use a shared cache for yarn

kevindew:
Trello: https://trello.com/c/MeG1zc2m/195-roll-out-stylelint-config-gds-across-govuk

This can speed up installs quite significantly when packages have been
installed by a different project. I saw Specialist Publisher go from
~30s install to ~10s - both with an empty node_modules.


benthorner (approved):
Nice find! I see it default to some `/usr/local...` directory. We primarily use `/root` as a place to cache things, so this is consistent with that 👍.

PR #450, opened 2021-01-19
Fix using unsupported "HOST" environment variable

benthorner:
Previously we mainly set "HOST" to "0.0.0.0" so that a Rails server
would listen on all network interfaces, in particular, the virtual
one used by Docker for communication with the host machine. Rails 6
began a transition to use "BINDING" instead of "HOST" [1], and since
6.1 support for "HOST" has been removed [2]. All of our apps seem
to be on Rails 6+ now, so this change should be non-breaking.

[1]: https://github.com/rails/rails/commit/37b373a8d2a1cd132bbde51cd5a3abd4ecee433b
[2]: https://github.com/rails/rails/commit/c0728ad3213d1e582871860e37c131b503372b2d#diff-f188e2ac461d2bcd6f3e4a3b27302df22ad2c696793637667a1f6ef59b00d2ae

### DIFF
--- a/projects/asset-manager/docker-compose.yml
+++ b/projects/asset-manager/docker-compose.yml
@@ -38,7 +38,7 @@ services:
       MONGODB_URI: "mongodb://mongo-3.6/asset-manager"
       REDIS_URL: redis://redis
       VIRTUAL_HOST: asset-manager.dev.gov.uk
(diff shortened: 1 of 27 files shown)


ChrisBAshton (approved):
Thanks for jumping on this, Ben!

PR #451, opened 2021-01-20
Remove REDIS_HOST environment variables

kevindew:
These apps have all been migrated to use REDIS_URL so they no longer
need this environment variable. There's a chance this might cause some
inconveniences for people working with old branches of apps, but I think
it's worth doing this sooner rather than risking let it linger.


benthorner (approved):

ChrisBAshton (approved):
Thanks Kevin!

Issue #452, opened 2021-01-20
Test DB not created when running "rake db:setup"


benthorner:
This isn't specific to GOV.UK Docker, but seems to affect all repos running Rails 6.1+. A workaround for this is to run:

```
govuk-docker-run env RAILS_ENV=test bundle exec rake db:setup
```

While it's something of a regression, it appears to be intentional and related to [this Rails issue](https://github.com/rails/rails/issues/27299). Next steps:

- Find the change in Rails that lead to this change in behaviour. Hopefully this will lead to documentation about the recommended approach, or we could refer to [the mainstream docs](https://guides.rubyonrails.org/v3.2/testing.html).

- Agree the fix and apply this to each affected app. For example, we may agree to add the above workaround to each `Makefile`. Ideally we should preserve the old behaviour, as that was simpler.

ChrisBAshton:
Did a little digging earlier, I think these two links might be relevant:
- https://github.com/rails/rails/pull/39027
- https://github.com/rails/rails/commit/6ca9031ba3c389f71366c3e6abf069c6924c5acf

ChrisBAshton:
Just tried the steps in the mainstream docs. `rake db:test:prepare` in [step 1](https://guides.rubyonrails.org/v3.2/testing.html#the-three-environments), as well as `rake db:migrate` and `rake db:test:load` in [the preparation section](https://guides.rubyonrails.org/v3.2/testing.html#preparing-your-application-for-testing) all fail with `ActiveRecord::NoDatabaseError: Unknown database 'contacts_test'`.

benthorner:
Thanks for investigating @ChrisBAshton! I looked a bit deeper and found that it's due to [this PR](https://github.com/rails/rails/pull/37351). It should only affect people running apps with GOV.UK Docker, as opposed to directly on their machine.

Some options:

1. We could move away from `DATABASE_URL` entirely. This is used by the deployed versions of our apps, so it would be quite a difficult change to make overall.

2. We could manually set the `url` for the development environment in each app, so that it reads from e.g. `ENV["DEV_DATABASE_URL"]`, much like we do for test ([here](https://github.com/alphagov/email-alert-api/blob/59e9ec81ac1caee1c85e6109d9b28a9e8a96045c/config/database.yml#L20)).

3. We could live with the regression and add the workaround to each `Makefile`.

[...]

PR #454, opened 2021-01-27
Simplify and enhance DB setup for Rails 6.1

benthorner:
Fixes: https://github.com/alphagov/govuk-docker/issues/452

This simplifies the existing DB setup steps before adding another to
cope with a breaking change in Rails 6.1. While not all of our repos 
use Rails 6.1 yet, this change is backwards-compatible; we may as 
well fix the problem for everything preemptively if we can Please see 
the commits for more details.


kevindew (approved):
Looks good - thanks for fixing it.

ChrisBAshton (approved):
Thanks Ben, LGTM too! 👍

Issue #444, opened 2020-12-14
Compatibility documentation incorrect for some apps


benthorner:
The [compatiblity doc](https://github.com/alphagov/govuk-docker/blob/master/docs/compatibility.md) should be corrected:

- Clarify what support there is for "frontend".
- Clarify what support there is for "whitehall".
- Clarify what support there is for "calculators".

PR #455, opened 2021-01-27
Refresh compatibility list

benthorner:

Fixes: https://github.com/alphagov/govuk-docker/issues/444

Depends on: https://github.com/alphagov/calculators/pull/1030

This switches to a more practical / positive approach to compatibility
for some services, and clarifies the compatibility for others. Please
see the commits for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ Do this to start a GOV.UK web app:
 govuk-docker up collections-publisher-app
 ```
 
+👉 [Replicate data locally](docs/how-tos.md#how-to-replicate-data-locally) (or use the [`app-live` stack](#the-app--stacks)).
+
 👉 [Check the troubleshooting guide if you have a problem.](docs/troubleshooting.md)
 
 ### The `app-*` stacks
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -14,8 +14,7 @@ These are repos that can be started as a some kind of process, such as a web app
    - ⚠ bouncer
       * **TODO: Missing support for a webserver stack**
    - ✅ cache-clearing-service
-   - ⚠ calculators
-      * Web UI doesn't work without the content item being present in the content-store.
+   - ✅ calculators
    - ❌ ckan
       * Has a [separate](https://github.com/alphagov/docker-ckan) Docker project.
    - ✅  collections
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -18,9 +18,7 @@ These are repos that can be started as a some kind of process, such as a web app
       * Web UI doesn't work without the content item being present in the content-store.
    - ❌ ckan
       * Has a [separate](https://github.com/alphagov/docker-ckan) Docker project.
-   - ⚠  collections
-      * You will need to [populate the Content Store database](#mongodb) or run the live stack in order for it to work locally.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -87,6 +87,8 @@ Do this to start a GOV.UK web app:
 govuk-docker up collections-publisher-app
 ```
 
+👉 [Replicate data locally](docs/how-tos.md#how-to-replicate-data-locally) (or use the `app-live` stack).
```
kevindew, line 90: I feel like it'd be helpful if this linked to #the-app--stacks - I know it's only a couple of lines below but it doesn't seem super clear that contextually that this is what that refers to.

docs/compatibility.md
```diff
@@ -46,8 +46,7 @@ These are repos that can be started as a some kind of process, such as a web app
       * **TODO: Missing support for running the worker**
    - ✅ govuk_publishing_components
    - ✅ govuk-developer-docs
-   - ⚠ hmrc-manuals-api
-      * **TODO: Missing support for a webserver stack**
+   - ✅ hmrc-manuals-api
```
kevindew, line 48: I'm not too sure what you mean by the rationale for this not needing a webserver stack? I'd have thought this one was as relevant as other apps that are flagged as not having a webserver stack as the interface to this app isn't exposed so a developer can only run tests?

kevindew (commented):
Great job Ben - this looks a lot better now and it's nice that it is up-to-date. There's just a couple of minor comments I've got.

docs/compatibility.md
benthorner: That was the original reasoning. However, with this app I paused to think: would someone ever need the app stack? This app is quite special: it's an API app that's only called externally. On the other hand, the other apps are either internal dependencies or have a GUI that someone may want to develop against. Bouncer is similar too, actually.

I don't really mind either way. This file is somewhat vague in that the compatibility will always be superficial and certain features may not work. Here I'm pushing to only flag missing pieces that would clearly add value. That said, it wouldn't be hard to add and test an app stack for these apps, so I don't feel strongly either way.

What do you think?

README.md
benthorner: Sure, that's fair. I've added that in 👍.

docs/compatibility.md
kevindew: > would someone ever need the app stack? 

I'd have been expecting that when you're working with this API you want a version running that you can send requests to - through postman, curl or similar. I recall when I worked a lot with the Publishing API I'd be spending a lot of my time with postman open. Otherwise any experimentation with it has to be done through the test suite, which seems unusual.

My logic for thinking it should have `**TODO: Missing support for a webserver stack**` is that this text implies that the interface to a running server of the app isn't available - it doesn't seem specific to whether that interface is graphical or not. So precedent of the tick would indicate a server of the app is available.

benthorner: > when you're working with this API you want a version running that you can send requests to

This is what I'm sceptical of. But I haven't worked with it myself, so it's just a hunch.

>So precedent of the tick would indicate a server of the app is available.

That's true: it makes less sense looking at it in reverse like that. I've removed the commit, as it's not a big deal either way: someone can just go and do the work to add the missing piece when it's eventually needed.

kevindew: Cool thanks - I guess the real problem is we have an app that virtually no-one works with 😞

kevindew (approved):

Issue #176, opened 2019-08-27
DNS resolution strategy incompatible with Rails 6


ChrisBAshton:
Arises from #174.

Applications running [Rails 6 have "Host Authorization" middleware](https://www.fngtps.com/2019/rails6-blocked-host/) to prevent against DNS rebinding attacks. This means that only requests from `0.0.0.0`, `::`, and `localhost` are allowed by default.

To get the custom alias domains (such as http://collections-publisher.dev.gov.uk/) to work, applications will either need to:

1. Explicitly add the local domain to its `config.hosts` (as has been done in [link-checker-api](https://github.com/alphagov/link-checker-api/pull/284)), or
2. Clear the entire whitelist (`config.hosts.clear`), preferably on local dev only. This lets through requests for all hostnames.

issyl0:
Commenting on this so I get notifications about follow up. (I'm leading on Platform Health's "upgrade all the apps to Rails 6" mission this quarter.) Good spot! I'll add a card to our backlog to do this (or similar) for all our apps.

benthorner:
For greater flexibility, I'm changing everything to do `config.hosts.clear` in development, which is perfectly safe and is less coupled to our particular domain. @issyl0 could you update the Platform Health card?

issyl0:
This is great, thanks @benthorner!

kevindew:
We think this will be resolved as all apps are now on Rails 6. Can reopen if not.

PR #456, opened 2021-02-04
Add web server stack for contacts admin

ChrisBAshton:
This is to facilitate https://trello.com/c/cIv5x8wB/2348-5-add-ability-to-unpublish-with-a-redirect-in-contacts-app.

Contacts Admin doesn't work perfectly with govuk-docker: notably, you currently need to run Publishing API in another shell. I'm not sure why that's the case, and am hoping it is something we can fix here, rather than have to document in https://github.com/alphagov/contacts-admin/pull/838.

### DIFF
--- a/projects/contacts-admin/docker-compose.yml
+++ b/projects/contacts-admin/docker-compose.yml
@@ -29,3 +29,22 @@ services:
       TEST_DATABASE_URL: "mysql2://root:root@mysql-5.5/contacts_test"
       REDIS_URL: redis://redis
 
+  contacts-admin-app: &contacts-admin-app
+    <<: *contacts-admin
+    depends_on:
+      - mysql-5.5
+      - redis
+      - publishing-api-app
+      - nginx-proxy
+      - content-store-app
+      - whitehall-app
+    environment:
+      DATABASE_URL: "mysql2://root:root@mysql-5.5/contacts_development"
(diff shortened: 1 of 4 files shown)


brucebolt (approved):

callumknights (approved):

projects/contacts-admin/docker-compose.yml
```diff
@@ -29,3 +29,22 @@ services:
...
+      - redis
+      - publishing-api-app
+      - nginx-proxy
+      - content-store-app
+      - whitehall-app
+    environment:
+      DATABASE_URL: "mysql2://root:root@mysql-5.5/contacts_development"
+      TEST_DATABASE_URL: "mysql2://root:root@mysql-5.5/contacts_test"
+      REDIS_URL: redis://redis
+      VIRTUAL_HOST: contacts-admin.dev.gov.uk
```
benthorner, line 45: I'm a bit late to the party, but trying this out locally breaks due to [this issue](https://github.com/alphagov/govuk-docker/issues/176). Do we need a separate PR to fix that?

ChrisBAshton: That PR is here: https://github.com/alphagov/contacts-admin/pull/838

benthorner: Ah, I see - I didn't clock that it was hidden in there, thanks!

PR #457, opened 2021-02-08
Modify account dev URL to serve from SecureContext

huwd:
[Account Team Trello](https://trello.com/c/QTDcQIui/616-https-er-ize-local-dev-setup)

The account manager prototype contains UI to allow a user to log in and
manage accounts.

As part of that login journey we have been prototyping webauthn
integration. This has revealed there are a range of secure-only browser
APIs that are only available over a HTTPs connection or in another
secure context.

We looked into adding HTTPs certificates however, an easier win seemed
to be modifying the domain to serve from a `.localhost` domain which is
regarded as a secure context[1].

This may have some advantages over self signed certificates in terms of
simplicity of setup[2], but that requires a bit more of a look.

So as a result we move the URL for only the account manager from a
`dev.gov.uk` account to a `.localhost` domain.

[1]: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts
[2]: https://medium.com/@bramus/on-secure-contexts-in-firefox-https-for-local-development-and-a-potential-nice-gesture-by-cea4ab4903a

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -123,4 +123,4 @@ services:
           - www-origin.dev.gov.uk
           - www.dev.gov.uk
           - attributes.login.service.dev.gov.uk
-          - www.login.service.dev.gov.uk
+          - www.login.service.localhost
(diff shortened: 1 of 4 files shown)


kevindew:
This looks pragmatic but relatively far reaching. Would we be able to hold fire on any merging until we confirm that Accounts is using webauthn (or something else that requires secure-context) in production? It feels like something quite experimental and applying this change would probably have dnsmasq consequences to resolve (which could be more confusing with a more generic domain - you may want `*.govuk.localhost` for a stronger prefix)

A simpler approach you could take is to expose a port for the service (for example adding `ports: [ "33242:3000" ]` to the app definition. This would make the service available on both the `*.dev.gov.uk` hostname and on `localhost:33242`. Trying with Chrome on my machine this works with `navigator.credentials` returning a `CredentialsContainer` object.

PR #460, opened 2021-02-18
Remove account-manager / finder-frontend JWT env vars

barrucadu:
These are no longer needed as finder-frontend is calling an
authenticated endpoint in the account-manager.

---

[Trello card](https://trello.com/c/nfoKVqOK/629-use-the-new-non-posty-jwt-mechanism-in-the-transition-checker)


huwd (approved):
nicely remembered, thanks

PR #462, opened 2021-02-26
Document retrying HTTP requests / rake tasks

ChrisBAshton:
Sometimes these fail because Publishing API isn't fully ready at
the time of action. We've sometimes documented this gotcha in the
application repo, but it is better to centralise the documentation
here. See
https://github.com/alphagov/contacts-admin/pull/845#discussion_r583457007


MuriloDalRi (approved):

benthorner:
I'm late to the party, but this looks good to me - thanks @ChrisBAshton !

PR #465, opened 2021-03-10
Add new account-api app

barrucadu:
We now have three account-related apps in govuk-docker:

1. govuk-account-manager-prototype
2. govuk-attribute-service-prototype
3. account-api

(1) and (2) are running in production on the PaaS, and are prototypes
of the cross-government authentication and attribute storage systems,
with the GOV.UK Transition Checker as the only user currently.

Responsibility for the cross-gov side of accounts is moving to the
Digital Identity (DI) programme, so those apps will be retired as soon
as DI have viable replacements.

(3) is a GOV.UK-internal app specified in [RFC 134][] which lets
different GOV.UK microservices make use of the accounts and
attributes.

When (3) is set up, finder-frontend will talk to it, and not to the
other apps.

[RFC 134]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-134-govuk-wide-session-cookie-and-login.md

---

To make this work we need to add `config.hosts.clear` to the `config/environments/development.rb` of account-api, like this: https://github.com/alphagov/email-alert-api/blob/6d5ed70353cf260910a1f7d7788939600c205ab4/config/environments/development.rb#L55-L56

---

[Trello card](https://trello.com/c/ZeAKl2SM/651-set-up-a-new-app)

### DIFF
--- a/projects/account-api/Makefile
+++ b/projects/account-api/Makefile
@@ -1,0 +1,3 @@
+account-api: bundle-account-api
+	$(GOVUK_DOCKER) run $@-lite bin/rails db:setup
+	$(GOVUK_DOCKER) run $@-lite env RAILS_ENV=test bin/rails db:setup
--- a/projects/account-api/docker-compose.yml
+++ b/projects/account-api/docker-compose.yml
@@ -1,0 +1,42 @@
+version: '3.7'
+
+volumes:
+  account-api-tmp:
+
+x-account-api: &account-api
+  build:
+    context: .
+    dockerfile: Dockerfile.govuk-base
+  image: account-api
+  stdin_open: true
+  tty: true
+  volumes:
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - root-home:/root
+    - account-api-tmp:/govuk/account-api/tmp
+  working_dir: /govuk/account-api
+
+services:
+  account-api-lite:
+    <<: *account-api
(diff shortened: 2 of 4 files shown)


projects/account-api/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+    - ${GOVUK_ROOT_DIR:-~/govuk}:/govuk:delegated
+    - root-home:/root
+    - account-api-tmp:/govuk/account-api/tmp
+  working_dir: /govuk/account-api
+
+services:
+  account-api-lite:
+    <<: *account-api
+    depends_on:
+      - memcached
```
kevindew, line 23: Is this required? I notice it isn't in the app definition.

barrucadu: Whoops

projects/account-api/Makefile
```diff
@@ -0,0 +1,3 @@
+account-api: bundle-account-api
+	$(GOVUK_DOCKER) run $@-lite bin/rails db:setup
+	$(GOVUK_DOCKER) run $@-lite env RAILS_ENV=test bin/rake db:setup
```
kevindew, line 3: Minor - we're mixing bin/rails and bin/rake here which is unnecessary. Sticking with bin/rails for both would be best.

kevindew (approved):
LGTM

PR #466, opened 2021-03-11
Add account-api to finder-frontend-app-account deps, and account-manager to account-api-app deps

barrucadu:
We're moving the communication with the account manager to account-api, so we now have these service dependencies.

---

[Trello card](https://trello.com/c/Y8z5rsJl/653-move-oauth-sign-in-logic-from-the-transition-checker-to-the-new-app)

### DIFF
--- a/projects/finder-frontend/Makefile
+++ b/projects/finder-frontend/Makefile
@@ -1,2 +1,2 @@
-finder-frontend: bundle-finder-frontend content-store static search-api router email-alert-api govuk-account-manager-prototype
+finder-frontend: bundle-finder-frontend account-api content-store static search-api router email-alert-api govuk-account-manager-prototype
 	$(GOVUK_DOCKER) run $@-lite yarn
--- a/projects/finder-frontend/docker-compose.yml
+++ b/projects/finder-frontend/docker-compose.yml
@@ -62,6 +62,7 @@ services:
     depends_on:
       - router-app
       - nginx-proxy
+      - account-api-app
       - email-alert-api-app
       - static-app
       - memcached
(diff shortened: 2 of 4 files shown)


projects/finder-frontend/Makefile
```diff
@@ -1,2 +1,2 @@
-finder-frontend: bundle-finder-frontend content-store static search-api router email-alert-api govuk-account-manager-prototype
+finder-frontend: bundle-finder-frontend account-api content-store static search-api router email-alert-api govuk-account-manager-prototype
```
AlanGabbianelli, line 1: Do we still need `govuk-account-manager-prototype` here or can we remove it since it's already in the `account-api`'s Makefile?

barrucadu: There's a test that the dependencies in the Makefile match the dependencies in the docker-compose file.

projects/finder-frontend/docker-compose.yml
```diff
@@ -62,6 +62,7 @@ services:
     depends_on:
       - router-app
       - nginx-proxy
+      - account-api-app
```
AlanGabbianelli, line 65: Similar question as above: since we now have `account-api-app` as a dependency, which in turn depends on `govuk-account-manager-prototype-app`, can we remove `govuk-account-manager-prototype-app` from this list?

projects/finder-frontend/Makefile
AlanGabbianelli: Can it be removed from both and still work as an implicit dependency that comes from `account-api`? If so, is there any value in still keeping it explicit?

barrucadu: Currently finder-frontend still does directly call the account manager in places, so I think there's value in keeping it as an explicit dependency until those bits are removed.

AlanGabbianelli: I see, thanks for clarifying.

projects/finder-frontend/docker-compose.yml
AlanGabbianelli: You answered this in the previous comment.

AlanGabbianelli (approved):

PR #469, opened 2021-03-25
Update finder-frontend account stack to govuk account

huwd:
[Trello](https://trello.com/c/43FM4pwv/673-switch-to-the-new-sign-in-out-endpoints)

We have migrated parts of the GOV.UK Account out of finder-frontend
(where it was specific to the brexit checker) to frontend (where it acts
as a global login).

We've also removed the feature flag from finder-frontend, the account is
assumed to always be on.

### DIFF
--- a/projects/finder-frontend/docker-compose.yml
+++ b/projects/finder-frontend/docker-compose.yml
@@ -62,17 +62,13 @@ services:
     depends_on:
       - router-app
       - nginx-proxy
+      - frontend-app-account
       - account-api-app
       - email-alert-api-app
       - static-app
       - memcached
-      - govuk-account-manager-prototype-app
     environment:
       ASSET_HOST: finder-frontend.dev.gov.uk
       VIRTUAL_HOST: finder-frontend.dev.gov.uk
       BINDING: 0.0.0.0
       MEMCACHE_SERVERS: memcached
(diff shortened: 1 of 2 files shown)


barrucadu (commented):
With the other changes, finder-frontend no longer directly depends on govuk-account-manager-prototype, so it can be removed from the Makefile and docker-compose.yml.

projects/finder-frontend/docker-compose.yml
```diff
@@ -62,6 +62,7 @@ services:
     depends_on:
       - router-app
       - nginx-proxy
+      - frontend-app
```
barrucadu, line 65: With #474, let's make this `frontend-app-account`

projects/finder-frontend/docker-compose.yml
```diff
@@ -75,4 +75,3 @@ services:
       GOVUK_ACCOUNT_OAUTH_CLIENT_ID: transition-checker-id
       GOVUK_ACCOUNT_OAUTH_CLIENT_SECRET: transition-checker-secret
       PLEK_SERVICE_ACCOUNT_MANAGER_URI: "http://www.login.service.dev.gov.uk"
```
barrucadu, line 77: We can also remove these env vars

barrucadu (approved):

PR #471, opened 2021-03-26
Use `db:setup` instead of `db:create`

1pretz1:
Allows us to seed the database which is a requirement for running the
application locally.

Trello:


1pretz1:
Closing in favour of https://github.com/alphagov/govuk-docker/pull/472.

PR #479, opened 2021-04-19
Add support for Bouncer web stack

benthorner:

This app is different to others on GOV.UK:

- It shares a DB with transition, so the DB name must match [1].

- It's purpose is to redirect third-party domains to transitioned
alternatives. In order to support testing, the app service has a
'bouncer-redirect' domain, in addition to the standard 'bouncer'
domain for testing routes like the healthcheck for the app [2].

- It uses a 'mr-sparkle' gem to support auto reloading (this is a
bare Rack app, so doesn't have this Rails-like feature itself).

[1]: https://github.com/alphagov/govuk-docker/blob/fd419bbee5e42a49e24f77f45ae534101c1cd657/projects/transition/docker-compose.yml#L27
[2]: https://github.com/alphagov/bouncer/blob/ec8022198d0be786c0d68427e4b366134cd2cc67/lib/bouncer/app.rb#L12

### DIFF
--- a/projects/bouncer/docker-compose.yml
+++ b/projects/bouncer/docker-compose.yml
@@ -20,3 +20,16 @@ services:
     environment:
       TEST_DATABASE_URL: "postgresql://postgres@postgres-9.6/transition_test"
       DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: "true"
+
+  bouncer-app:
+    <<: *bouncer
+    depends_on:
+      - postgres-9.6
+      - nginx-proxy
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres-9.6/transition"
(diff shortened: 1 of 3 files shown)


projects/bouncer/docker-compose.yml
```diff
@@ -20,3 +20,16 @@ services:
...
+    depends_on:
+      - postgres-9.6
+      - nginx-proxy
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres-9.6/transition"
+      VIRTUAL_HOST: bouncer.dev.gov.uk,bouncer-redirect.dev.gov.uk
+      BINDING: 0.0.0.0
+    expose:
+      - "8080"
+    command: bundle exec mr-sparkle --force-polling
```
ChrisBAshton, line 35: Great that you've been able to maintain the auto-reloading functionality too! 👏

ChrisBAshton (approved):
Great stuff.

PR #486, opened 2021-05-07
Build Licence Finder search on make and fix misconfigured dependencies

kevindew:
This adds the `search:index` task as part of the make process for Licence Finder. This resolves a problem where search won't work until it has run (as the indexes didn't exist).

This also resolves some mistakes in the Licence Finder dependencies.


benthorner (approved):
Looks good, thanks for fixing the dependency too, I didn't realise that was missing as well!

PR #492, opened 2021-05-26
Remove `app-account` stack and add an `account-api-app-live` project

barrucadu:
See commit messages.


projects/collections/docker-compose.yml
```diff
@@ -49,14 +49,3 @@ services:
       PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
       VIRTUAL_HOST: collections.dev.gov.uk
       BINDING: 0.0.0.0
-
-  collections-app-account:
```
huwd, line 53: I know it's pretty minor, but wouldn't `account-api` be needed for the logged in header?
I see we've just removed this here,

So presumably the thing would be to have two separate terminals one with:
`govuk-docker up finder-frontend-app` and the other with `govuk-docker up collections-app` if you needed the full brexit checker journey?

projects/frontend/docker-compose.yml
```diff
@@ -47,17 +50,3 @@ services:
       PLEK_SERVICE_STATIC_URI: assets.publishing.service.gov.uk
       VIRTUAL_HOST: frontend.dev.gov.uk
       BINDING: 0.0.0.0
-
```
huwd, line 50: shouldn't `frontend-app-live` gain `account-api-app-live` as a `depends_on`?

Otherwise I'm not sure what happpens if you try to spin up `govuk-docker frontend-app` and it tries to bring up `account-api-app` when we really need `account-api-app-live` to get real content IDs?

projects/collections/docker-compose.yml
barrucadu: No, [collections doesn't use account-api at all](https://github.com/alphagov/collections/blob/main/app/controllers/concerns/account_concern.rb)

projects/frontend/docker-compose.yml
barrucadu: Sure, I'll add that

projects/collections/docker-compose.yml
huwd: great! ok!

huwd (approved):
nice one!
There's a couple of places i think we mention `app-accounts` in the [team manual](https://github.com/alphagov/govuk-account-team-manual/) I think.

Might need to update there, or I'll go back over them after I've wrapped the `frontend` work.

Issue #491, opened 2021-05-25
Foreign key constraint failure in collections-publisher-lite db:setup


aldavidson:
On a first-ever run of `make collections-publisher`, I get this error:

```
docker-compose -f [...] run collections-publisher-lite bin/rake db:setup
Creating govuk-docker_collections-publisher-lite_run ... done
Created database 'collections_publisher_development'
Created database 'collections_publisher_test'
/home/al/govuk/govuk-docker/exe/govuk-docker run collections-publisher-lite env RAILS_ENV=test bin/rake db:setup
docker-compose -f [...] run collections-publisher-lite env RAILS_ENV=test bin/rake db:setup
Creating govuk-docker_collections-publisher-lite_run ... done
Database 'collections_publisher_test' already exists
rake aborted!
ActiveRecord::InvalidForeignKey: Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails
/govuk/collections-publisher/db/schema.rb:25:in `block in <top (required)>'
/govuk/collections-publisher/db/schema.rb:13:in `<top (required)>'
Caused by:
Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails
/govuk/collections-publisher/db/schema.rb:25:in `block in <top (required)>'
/govuk/collections-publisher/db/schema.rb:13:in `<top (required)>'
Tasks: TOP => db:schema:load
(See full trace by running task with --trace)
```

I've worked around it using @kevindew 's suggestion in [this Slack thread](https://gds.slack.com/archives/CAB4Q3QBW/p1621953048057800?thread_ts=1621951685.053200&cid=CAB4Q3QBW) -

1. run `govuk-docker run collections-publisher-lite bin/rails db:drop`
2. comment out the second `db:setup` invocation for the test environment on [line 3 of the collections-publisher Makefile](https://github.com/alphagov/govuk-docker/blob/2fe4ee96cfd4fdae9a042727cefb79b85889b8f1/projects/collections-publisher/Makefile#L3)
3. re-run `make collections-publisher`

- but it would be good to fix the issue for the next new starter

kevindew:
Thanks for raising this @aldavidson and welcome to GOV.UK 🎉 

[...]

PR #494, opened 2021-05-28
Switch to db:prepare from db:setup

benthorner:
Fixes: https://github.com/alphagov/govuk-docker/issues/491

In 1ec59d4 we simplified the command to initialise a DB for an app,
since Rails 6 requires we run it for the dev and test environments
separately. The only command swallowed any errors, which masked an
issue that 'db:setup' isn't idempotent for MySQL DBs.

This switches to 'db:prepare', which has roughly the same behaviour
as the old command (including running any seeds), but with the added
bonus of not swallowing any genuine errors. Tested with Collections
Publisher (MySQL) and Content Publisher (Postgres).

### DIFF
--- a/projects/account-api/Makefile
+++ b/projects/account-api/Makefile
@@ -1,3 +1,3 @@
 account-api: bundle-account-api content-store govuk-account-manager-prototype
-	$(GOVUK_DOCKER) run $@-lite bin/rails db:setup
-	$(GOVUK_DOCKER) run $@-lite env RAILS_ENV=test bin/rails db:setup
+	$(GOVUK_DOCKER) run $@-lite bin/rails db:prepare
+	$(GOVUK_DOCKER) run $@-lite env RAILS_ENV=test bin/rails db:prepare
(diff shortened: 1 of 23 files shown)


kevindew:
Thanks for implementing this. It works well on Postgres and MySQL dbs.

Unfortunately db:prepare is not a task that is available for Mongoid DB's so it will fail on those apps (Asset Manager, Publisher, Specialist Publisher are quick ones that come to mind). I think we can get away with that using `db:setup` though (so long as our seed files are idempotent, which they tend to be - much to my surprise).

By the way I noticed Content Store wasn't included here and it seems that it has (as well as Maslow) a slightly unconventional DB setup, that, at a quick glance, looked like it could be changed to the normal one.

benthorner:
Oops, good spot @kevindew! I've amended the commit so that all the repos using Mongoid continue to use `db:setup`. It's a shame we can't use the same task everywhere, as there's always a risk of someone copying the wrong one if they add a project in future. I'm surprised by just how many of our apps use MongoDB.

> I noticed Content Store wasn't included here and it seems that it has (as well as Maslow) a slightly unconventional DB setup

Indeed it can. I've added a commit to do that. You're right that most of our repos do have idempotent seeds. In fact, after looking through them, I can't find one that doesn't: many have explicit conditionals; a couple use `create` instead of `create!` so fail silently; even authenticating proxy seems to work, [despite using `save!` at the end](https://github.com/alphagov/authenticating-proxy/blob/main/db/seeds.rb).

PR #495, opened 2021-05-28
Add PGHOST environment variable for Mapit

brucebolt:
This was removed in #482, but we now get an error when making Mapit:

```
createuser: could not connect to database template1: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

Therefore restoring this environment variable so Mapit can be built.

[Trello card](https://trello.com/c/IwakiikF)


benthorner (approved):
Oops, thanks @brucebolt ⭐. I remember testing without it, but I think that was just the repo tests, and not `make` commands, so what you say makes sense 👍.

PR #498, opened 2021-06-08
Make in-browser Jasmine tests available for GOV.UK Publishing Components

injms:
The GOV.UK Publishing Components gem has JavaScript tests that use Jasmine. This can be run in a CI environment and in the browser. This pull requests adds the ability to run the Jasmine server using GOV.UK Docker.

Using a non-`localhost` domain seems to result in certain Jasmine tests failing - so this has had to use Jasmine's usual `localhost:8888` as the virtual host instead of `govuk-publishing-components`.

To run the server:

```sh
govuk-docker up govuk_publishing_components-jasmine
```

and then visit `http://localhost:8888`

### DIFF
--- a/projects/govuk_publishing_components/docker-compose.yml
+++ b/projects/govuk_publishing_components/docker-compose.yml
@@ -20,6 +20,12 @@ services:
   govuk_publishing_components-lite:
     <<: *govuk_publishing_components
 
+  govuk_publishing_components-jasmine:
+    <<: *govuk_publishing_components
+    ports:
+      - "8888:8888"
+    command: bundle exec rake app:jasmine
+
   govuk_publishing_components-app:
     <<: *govuk_publishing_components
     depends_on:


projects/govuk_publishing_components/docker-compose.yml
```diff
@@ -20,6 +20,14 @@ services:
   govuk_publishing_components-lite:
     <<: *govuk_publishing_components
 
+  govuk_publishing_components-jasmine:
+    <<: *govuk_publishing_components
+    environment:
+      VIRTUAL_HOST: localhost # Jasmine doesn't like non localhost domains
+    expose:
+      - "8888"
```
kevindew, line 28:
```suggestion
    ports:
      - "8888:8888"
```

Setting VIRTUAL_HOST requires a dependency on nginx-proxy (that's the app that picks up the VIRTUAL_HOST env var) I imagine this worked for you without setting that in depends_on because nginx-proxy was already running in the background.

I think it'd be better to use ports rather than expose though as we'll be to access the Jasmine tests via http://localhost:8888 which is the same URL that Jasmine will suggest and won't require someone to have the extra knowledge of how to access the tests.

kevindew (commented):
Thanks for putting this together, I'm a fan of the web view for the Jasmine tests. 

I've made a suggestion of how we can run this via a port so it matches the URL Jasmine itself suggests.

A thing that came to mind was that if/when we add this for multiple apps they'd likely be using the same port which could cause errors if someone tried to run the tests for one app while running another. But this seems both unlikely and an existing problem for running Jasmine tests.

projects/govuk_publishing_components/docker-compose.yml
injms: Super - thanks for this. Agree about using Jasmine's default port, I just didn't know that it could be done. I've added this to the commit 👍

kevindew (approved):
Great stuff 👍

benthorner:
This looks good to me too 👍. I agree it's quicker / easier to run Jasmine tests in the browser, so I'm all for adding this functionality. My guess is that people have previously been doing this on their host machine instead.

On a more major note, we should document this, as it sets a new precedent that we may want to reuse in other projects. We could list it as a new [kind of stack](https://github.com/alphagov/govuk-docker#usage), and/or add [a how-to](https://github.com/alphagov/govuk-docker/blob/main/docs/how-tos.md) about running Jasmine tests interactively. We should also update the README for GOV.UK Publishing Components, as the current `bin/rake app:jasmine` won't work.

On a more minor note, it's a shame we can't make the approach a bit easier - someone has to know about the special way of doing it. I had a think about possible alternatives, but couldn't come up with anything that much better:

1. We could expose a port on the `-lite` stack. This would mean you could do `govuk-docker-run bin/rake app:jasmine`, like it says in the README. The problem with this is that it might surprise people if the port clashes with something on the host, which is more likely as people use the `-lite` stack more frequently than your proposed `-jasmine` one.

2. We could use nginx-proxy on the `-lite` stack. This would also mean `govuk-docker-run bin/rake app:jasmine` works, but you'd  need to go to `govuk-publishing-components.dev.gov.uk` instead of localhost. One problem is that this would clash with any running `-app` stack. The non-localhost domain may also cause issues.

3. We could avoid a special stack entirely, and have a how-to for the equivalent command: `govuk-docker run -p 8888:8888 govuk_publishing_components-lite bin/rake app:jasmine`. This is closer to the README, and is maybe a bit clearer / less magical than having a dedicated stack, since you can directly see the command.

I lean a bit towards the first suggestion, just because it means the README for the repo can remain agnostic to whether someone is using GOV.UK Docker or not. To get around the port clash issue, we could tell Jasmine to use a more unusual port, and update the README to match.

On the other hand, the approach you've proposed works. As long as we document it clearly, then I agree this is good to go. I'm also interested if any of my suggestions above sound appealing.

PR #503, opened 2021-07-15
Add govuk-content-schemas-app

ChrisBAshton:
This is to enable `govuk-docker-run` to work on govuk-content-schemas, so that we can remove the startup.sh script.

Sister PR: https://github.com/alphagov/govuk-content-schemas/pull/1065
Trello: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

### DIFF
--- a/spec/integration/make/dependencies_spec.rb
+++ b/spec/integration/make/dependencies_spec.rb
@@ -3,19 +3,19 @@ require "spec_helper"
 RSpec.describe "Make dependencies" do
   ProjectsHelper.all_projects.each do |project_name|
     it "mirrors docker-compose.yml for #{project_name}" do
-      expect(compose_dependencies(project_name))
-        .to match_array(makefile_dependencies(project_name))
+      expect(app_dependencies_in_compose_file(project_name))
+        .to match_array(app_dependencies_in_makefile(project_name))
     end
   end
 
-  def compose_dependencies(project_name)
+  def app_dependencies_in_compose_file(project_name)
     project_stacks = ComposeHelper.services(project_name).values
     dependencies = project_stacks.flat_map { |s| s["depends_on"].to_a }
     dependencies = compose_remove_stack_from_service_name(dependencies)
     (dependencies & ProjectsHelper.all_projects) - [project_name]
   end
 
-  def makefile_dependencies(project_name)
+  def app_dependencies_in_makefile(project_name)
     MakefileHelper.dependencies(project_name) &
       compose_remove_stack_from_service_name(ComposeHelper.app_services.keys)
   end
--- a/spec/integration/make/dependencies_spec.rb
+++ b/spec/integration/make/dependencies_spec.rb
@@ -11,6 +11,14 @@ RSpec.describe "Make dependencies" do
       expect(app_dependencies_in_compose_file(project_name))
         .to match_array(app_dependencies_in_makefile(project_name))
     end
+
+    it "has only valid dependencies in the #{project_name} Makefile" do
+      app_dependencies = MakefileHelper.dependencies(project_name)
+        .reject { |dep| dep =~ /^(bundle|clone)/ }
+        .map { |dep| "#{dep}-app" }
+
+      expect((app_dependencies - ComposeHelper.all_services.keys)).to eq([])
+    end
   end
 
   def app_dependencies_in_compose_file(project_name)
--- a/spec/integration/make/dependencies_spec.rb
+++ b/spec/integration/make/dependencies_spec.rb
@@ -1,13 +1,15 @@
 require "spec_helper"
 
 RSpec.describe "Make dependencies" do
+  # rubocop:disable Lint/ConstantDefinitionInBlock
   # Some dependencies are only needed by the dependant app for their static files.
-  dependencies_that_do_not_need_to_be_running = %w[govuk-content-schemas]
+  DEPENDENCIES_THAT_DO_NOT_NEED_TO_BE_RUNNING = %w[govuk-content-schemas].freeze
+  # rubocop:enable Lint/ConstantDefinitionInBlock
 
   ProjectsHelper.all_projects.each do |project_name|
     it "mirrors docker-compose.yml for #{project_name}" do
-      expect(app_dependencies_in_compose_file(project_name) - dependencies_that_do_not_need_to_be_running)
-        .to match_array(app_dependencies_in_makefile(project_name) - dependencies_that_do_not_need_to_be_running)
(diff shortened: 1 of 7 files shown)


projects/content-store/Makefile
```diff
@@ -1,2 +1,2 @@
-content-store: bundle-content-store router-api govuk-content-schemas
+content-store: bundle-content-store router-api govuk-content-schemas-app
```
benthorner, line 1: This won't work / doesn't need to change.

(Similarly for publishing-api.)

benthorner (approved):
Works when combined with the other PR mentioned in the description 👍.

Just one more major comment that the build is broken, but I'll approve now.

projects/content-store/Makefile
ChrisBAshton: Ah, I was getting test failures without this change 🤔

benthorner: Are you sure? I wonder if it was something unrelated.

Conversely, I get `make` failures _with_ this change.

ChrisBAshton: I've realised my mistake now, I needed to add the new dependency to docker-compose.yml, not the Makefile 🤦 Could you double-check my changes please?

projects/publishing-api/docker-compose.yml
```diff
@@ -34,6 +34,7 @@ services:
   publishing-api-app: &publishing-api-app
     <<: *publishing-api
     depends_on:
+      - govuk-content-schemas-app
```
benthorner, line 37: I don't think we need these, as Content Schemas just need to be present, not started, which is what the Makefiles ensure.

Does that make sense?

projects/content-store/Makefile
benthorner: Hmm...nope 😄. I'll resolve this comment, but I've raised another...

projects/publishing-api/docker-compose.yml
ChrisBAshton: If I omit them, I get test failures:

```
Failures:

  1) Make dependencies mirrors docker-compose.yml for publishing-api
     Failure/Error:
       expect(compose_dependencies(project_name))
         .to match_array(makefile_dependencies(project_name))
     
       expected collection contained:  ["govuk-content-schemas"]
       actual collection contained:    []
       the missing elements were:      ["govuk-content-schemas"]
     # ./spec/integration/make/dependencies_spec.rb:6:in `block (3 levels) in <top (required)>'

Finished in 0.32347 seconds (files took 0.36969 seconds to load)
1081 examples, 1 failure
```

Any ideas?

benthorner: Ahh, I see. 

Now that `govuk-content-schemas` has an `-app` stack, [it's getting picked up by `makefile_dependencies`](https://github.com/alphagov/govuk-docker/blob/main/spec/integration/make/dependencies_spec.rb#L18). The expectation is that if an app needs another "app" in its Makefile, then it also needs the app to be running when it starts. I think the schemas are the one exception to this, because it's just the files that matter.

Optional: we could clarify a few of the names in that file:

- `compose_dependencies` => `app_dependencies_in_compose_file`
- `makefile_dependencies` => `app_dependencies_in_makefile`

Thinking about it that way, we can see that the schemas aren't an "app dependency" and we could tweak that method to exclude it as a special case.

Does that help?

spec/integration/make/dependencies_spec.rb
```diff
@@ -1,10 +1,13 @@
...
 
 RSpec.describe "Make dependencies" do
+  # Some dependencies are only needed by the dependant app for their static files.
+  dependencies_that_do_not_need_to_be_running = %w[govuk-content-schemas]
+
   ProjectsHelper.all_projects.each do |project_name|
     it "mirrors docker-compose.yml for #{project_name}" do
-      expect(app_dependencies_in_compose_file(project_name))
-        .to match_array(app_dependencies_in_makefile(project_name))
+      expect(app_dependencies_in_compose_file(project_name) - dependencies_that_do_not_need_to_be_running)
```
benthorner, line 9: Agree with subtracting in both places: this stops someone accidentally adding it as a dependency in their Compose file.

Minor: could we make this part of the methods instead? One of them already does a subtraction. "app" is a vague enough term that the methods should still make sense.

benthorner (approved):
Thanks for fixing the tests!

(The following is just an optional follow-up suggestion in case you're interested.)

One of the previous iterations of this PR showed up that we're missing a test to check the dependencies in the Makefiles are valid. The `govuk-content-schemas-app` dependency passed [the existing parity test](https://github.com/alphagov/govuk-docker/blob/main/spec/integration/make/dependencies_spec.rb) because it didn't correspond to a project name, so got ignored in the same way `bundle-govuk-content-schemas` gets ignored. We could test that after removing "allowed prefixes" (`bundle-` and `clone-`), all Makefile dependencies match a project name.

spec/integration/make/dependencies_spec.rb
ChrisBAshton: Good idea 👍 I did try this initially but had some issues, I think because I was wrapping the variable in `[]` 🤦 I've pushed up a commit for this, and a new test for the Makefiles.

spec/integration/make/dependencies_spec.rb
```diff
@@ -1,26 +1,30 @@
 require "spec_helper"
 
 RSpec.describe "Make dependencies" do
+  # rubocop:disable Lint/ConstantDefinitionInBlock
   # Some dependencies are only needed by the dependant app for their static files.
-  dependencies_that_do_not_need_to_be_running = %w[govuk-content-schemas]
+  DEPENDENCIES_THAT_DO_NOT_NEED_TO_BE_RUNNING = %w[govuk-content-schemas].freeze
```
benthorner, line 6: Sorry I'm late reviewing this. I found a slighter tidier approach if you want to review: https://github.com/alphagov/govuk-docker/pull/505

spec/integration/make/dependencies_spec.rb
```diff
@@ -11,6 +11,14 @@
       expect(app_dependencies_in_compose_file(project_name))
         .to match_array(app_dependencies_in_makefile(project_name))
     end
+
+    it "has only valid dependencies in the #{project_name} Makefile" do
+      app_dependencies = MakefileHelper.dependencies(project_name)
+        .reject { |dep| dep =~ /^(bundle|clone)/ }
+        .map { |dep| "#{dep}-app" }
```
benthorner, line 18: Optional: you could avoid this by instead comparing with `ProjectsHelper.all_projects`.

PR #505, opened 2021-07-20
Fix linting issue to avoid disabling Cop

benthorner:
Previously we used a constant but we can achieve the same thing with
a "let" statement, which doesn't violate any linting rules.

I think the reason the original variable stopped working is that the
reference in the test was in the same scope (of the "describe" block),
whereas the methods are run in some other scope, which appears to be
the same as that populated by the "let".

(Using a constant always worked because constants are global, hence
the linting issue.)

### DIFF
--- a/spec/integration/make/dependencies_spec.rb
+++ b/spec/integration/make/dependencies_spec.rb
@@ -1,10 +1,8 @@
 require "spec_helper"
 
 RSpec.describe "Make dependencies" do
-  # rubocop:disable Lint/ConstantDefinitionInBlock
   # Some dependencies are only needed by the dependant app for their static files.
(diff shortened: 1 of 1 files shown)


spec/integration/make/dependencies_spec.rb
```diff
@@ -21,18 +19,19 @@
     end
   end
 
+
```
ChrisBAshton, line 22:
```suggestion
```

ChrisBAshton (approved):
LGTM 👍 one small linting issue to fix.

spec/integration/make/dependencies_spec.rb
benthorner: Thanks for spotting, I've amended with that 👍.

PR #507, opened 2021-08-24
Remove unneeded dependencies from Dockerfiles

Tetrino:
The dependencies in question were removed from the remote repos, and are no longer required for Chromium to build.

Removing them outright is safe to do, and does not affect any development environment.


Tetrino:
@kevindew This doesn't have to be pulled in, but as far as I can tell through looking, Chromium just doesn't need these anymore, nor does it affect our environment.

Dockerfile.govuk-base
```diff
@@ -6,7 +6,7 @@
 FROM buildpack-deps
 
 # Install chrome and its dependencies
-RUN apt-get update -qq && apt-get install -y libxss1 libappindicator1 libindicator7
+RUN apt-get update -qq && apt-get install -y
```
kevindew, line 9:
```suggestion
```

We can remove this line (same with the other Dockerfiles changed here) as it is no longer installing any software with apt

kevindew (commented):
Thanks very much @Tetrino for digging into this, apologies I didn't notice this earlier

I too had the problem that `docker builder prune` didn't work, as that only removed dangling images. I think `docker image prune -a` will do the trick (but does nuke everything be warned). 

Have you tested some projects can still run Chrome ok? I gave this ago myself and I'm pretty confident it works. I was able to run some JS Content Publisher feature tests (I did hit a different annoying issue in Content Publisher where image copying fails `identify-im6.q16: insufficient image data in file `/tmp/mini_magick20210908-143-1b7oecp.jpg' @ error/jpeg.c/ReadJPEGImage_/1166.` 🤔  but I think that is unrelated)

Tetrino:
Just raising that I've not forgotten about this, I am snowed under with some pretty hefty work and as this isn't affecting most people I've put it to the side just for now.

PR #511, opened 2021-09-22
Fix `make release`

ChrisBAshton:
Running `make release` was failing on this error:

```
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateUsers < ActiveRecord::Migration[4.2]
```

This has been noticed in other apps before, e.g. [in contacts-admin](https://github.com/alphagov/govuk-docker/issues/381). The fix in that case was to call `db:setup`. However, we changed all of our `db:setup` calls to `db:prepare` in #494.

I could have followed the advice of the error message by editing the migrations' code in `release`, as was done in Content Tagger [in 2018](https://github.com/alphagov/content-tagger/pull/639). However, most recently in [July 2020](https://github.com/alphagov/signon/pull/1457), this approach was declined in Signon. Instead, the issue in that case was resolved by a careful series of deployments. See https://github.com/alphagov/signon/pull/1458#issuecomment-655415079.

> Rails maintains a schema_migrations table based on the filenames under the db/migrations directory. This lets it keeps track of which migrations have to be run.
>
> We had deployed a branch with migrations removed, set up the database, and then tried deploying the app with migrations. This didn’t work, since rails thought the migrations were new/down and tried to run them (as above), which isn’t what we wanted.

By explicitly executing `db:schema:load` before running `db:prepare`, we're telling Rails that all migrations have been run at the time of creating the database. Since we have no data in our development DB that we have to worry about migrating to the latest schema, we can safely run this step.

### DIFF
--- a/projects/release/Makefile
+++ b/projects/release/Makefile
@@ -1,4 +1,5 @@
 release: bundle-release
+	$(GOVUK_DOCKER) run $@-lite bin/rake db:schema:load
 	$(GOVUK_DOCKER) run $@-lite bin/rake db:prepare
 	$(GOVUK_DOCKER) run $@-lite env RAILS_ENV=test bin/rake db:prepare
 	$(GOVUK_DOCKER) run $@-lite yarn


kevindew:
I’d imagine you hit the problem you did because you had an empty database. As, If I remember right, prepare does db:setup if no db or db:migrate if there is one. So you’d fallen into a path that shouldn’t happen.

I can’t remember exactly what schema_load does (whether it nukes the db or not in particular) but, as I understand, would not be appropriate for a scenario where someone already has a populated database.

ChrisBAshton:
Ah, thanks for the insight. It shouldn't have happened, but does seem to be a fairly common quirk - definitely worth adding a section to the [Troubleshooting](https://github.com/alphagov/govuk-docker/blob/main/docs/troubleshooting.md) docs for next time a developer comes across this.

I've switched back to the `main` branch of govuk-docker and ran `make release` but all is well now, so I can't reproduce the issue anymore. I wonder what the most appropriate fix would be, next time? Should we advise developers to temporarily add the `db:schema:load` to their Makefile (as we know that works), or tell them to manually run `govuk-docker-run bundle exec rails db:setup` (untested, but should work, if I've understood your comment)?

kevindew:
Yup, adding something to the troubleshooting sounds good if this something people regularly get stuck on. It's hard to have a fully idempotent approach in the make file that can cope with empty system and already provisioned system.

I'd probably suggest that the best thing someone should do if they get stuck in this situation is run `db:drop` to nuke the db and get to a state where make would work when run again. There's a couple of quirks to consider though when providing guidance: only appropriate for ActiveRecord based projects and that you may need to do also do it on the test database.

benthorner:
I was wondering about a plausible sequence of steps to reproduce this. Could it have happened if `db:prepare` quit or errored between `db:create` and `db:schema:load`? Agree with suggesting `db:drop` in such cases 🚀.

PR #515, opened 2021-11-10
Update readme to make clear that the commands are using an example app name

owenatgov:
## What
Updates the startup docs replacing any instance of an example app used in a command with `[app-name]` along with an explainer that app-name is the app you're trying to run.

## Why
Currently, the readme uses collections-publisher as an example app for explaining how to use the commands to build and run. We've had a number of folks join who are trying to use docker before they fully grasp our app setup, leading to them thinking they have to run these commands against the example app, primarily collections-publisher, specifically and not understanding why they can't run the app they want to. This change spells it out in a bit more detail that the commands being used here are for every app and depend on what app you want to build/run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,12 @@ Running GOV.UK applications can be resource intensive. To give Docker more resou
 Do this the first time you work on a project:
 
 ```sh
+make [app-name]
+```
+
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👍

PR #523, opened 2021-12-02
Update Link Checker API to use PostgreSQL 13

ollietreend:

[Trello card](https://trello.com/c/zVdVJqFM/62-work-out-how-much-effort-is-required-to-upgrade-postgresql-databases-in-each-of-our-applications)


brucebolt (commented):
Looks good @ollietreend.

Could you add a comment (like we've done in https://github.com/alphagov/govuk-docker/pull/521) to mark that the version in use differs from that in production.

brucebolt (approved):

PR #524, opened 2021-12-02
Update Publishing API to use PostgreSQL 13

ollietreend:

[Trello card](https://trello.com/c/zVdVJqFM/62-work-out-how-much-effort-is-required-to-upgrade-postgresql-databases-in-each-of-our-applications)


beccapearce (approved):
👏🏻 👏🏻 👏🏻

PR #534, opened 2021-12-06
Upgrade MySQL to version 8

davidgisbey:
This is the version we are planning to upgrade MySQL to in production, so we need to update our development environment too.

trello card: https://trello.com/c/btcm7JyB/83-work-out-how-much-effort-is-required-to-upgrade-mysql-databases-in-each-of-our-applications

### DIFF
--- a/projects/signon/docker-compose.yml
+++ b/projects/signon/docker-compose.yml
@@ -22,21 +22,21 @@ services:
   signon-lite:
     <<: *signon
     depends_on:
-      - mysql-5.5
+      - mysql-8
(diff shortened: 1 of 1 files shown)


beccapearce (commented):
Looks good, but seems to have a line change in contacts-admin which doesn't seem related to this PR. Think we should remove it from the PR

davidgisbey:
Apologies! Have removed the line change in the contacts admin docker compose file.

beccapearce (approved):
LGTM

PR #535, opened 2021-12-06
Update Content Publisher to use PostgreSQL 13

ollietreend:

[Trello card][1]

**Note:** the test suite for this app doesn't currently pass in the docker environment. But the test failures seem to be unrelated to the database upgrade, as evidenced by the fact that [they pass in CI][2] with the same database upgrade.

We have a separate [Trello card][3] to resolve the failing tests in the docker environment.

[1]: https://trello.com/c/zVdVJqFM/
[2]: https://github.com/alphagov/content-publisher/pull/2431
[3]: https://trello.com/c/QZOaYQCH/

### DIFF
--- a/projects/content-publisher/docker-compose.yml
+++ b/projects/content-publisher/docker-compose.yml
@@ -22,17 +22,18 @@ services:
   content-publisher-lite:
     <<: *content-publisher
     depends_on:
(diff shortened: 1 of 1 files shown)


davidgisbey (approved):
lgtm

Issue #537, opened 2021-12-07
Fixture file based tests fail due to Linux Kernel issue


kevindew:
People have experienced problems running the test suites of Content Publisher and Whitehall (I'm not aware of other apps but there could be more) using govuk-docker. It transpires this issue is a problem with the Linux kernel, in particular the version that Docker-for-mac is using currently (on my machine it's `5.10.47-linuxkit`) but I'm recording this here so the GOV.UK issue is documented and can be tracked to a degree.

The evidence of the problem is tests that fail like so:

```
     MiniMagick::Invalid:
       `identify /tmp/mini_magick20211206-1-7orhba.jpg` failed with error:
       identify-im6.q16: insufficient image data in file `/tmp/mini_magick20211206-1-7orhba.jpg' @ error/jpeg.c/ReadJPEGImage_/1166.
```

or

```
Error:
ImageUploaderTest#test_should_store_all_the_versions_of_a_bitmap_image_in_asset_manager:
CarrierWave::ProcessingError: Failed to manipulate with MiniMagick, maybe it is not an image? Original Error: `convert /govuk/whitehall/tmp/1321009871-957757668381030-0168-8098/s960/minister-of-funk.960x640.jpg -auto-orient -resize 960x640^ -gravity Center -background rgba(255,255,255,0.0) -extent 960x640 /tmp/image_processing20111111-428-11fn1kn.jpg` failed with error:
convert-im6.q16: insufficient image data in file `/govuk/whitehall/tmp/1321009871-957757668381030-0168-8098/s960/minister-of-funk.960x640.jpg' @ error/jpeg.c/ReadJPEGImage_/1166.
convert-im6.q16: no images defined `/tmp/image_processing20111111-428-11fn1kn.jpg' @ error/convert.c/ConvertImageCommand/3229.
```

The source of the issue is a [relatively obscure set of circumstances](https://github.com/docker/for-linux/issues/1015): a file is stored on a docker mount (i.e. a test fixture), it is then copied to be a temporary file (i.e. a test faking the upload of a file) and then is copied again (i.e copying the uploaded file for processing).

The issue is discussed (at length) in: https://github.com/docker/for-linux/issues/1015. It looks like versions of Linux Kernel > 5.10 are fixed and it appears [that a backported fix for 5.10 is imminent](https://github.com/docker/for-linux/issues/1015#issuecomment-985799437), however it's completely unclear how long it will take for any fixes to make their way into a Docker-for-mac releases.

[...]

PR #539, opened 2021-12-07
Mount /tmp as a tmpfs volume to work around #537

ollietreend:
This commit is a workaround for the issue #537. It allows the Content Publisher and Whitehall test suites to pass in the GOV.UK Docker development environment.

Once docker/for-linux#1015 has been fixed, this workaround will no longer be needed.

### DIFF
--- a/projects/content-publisher/docker-compose.yml
+++ b/projects/content-publisher/docker-compose.yml
@@ -17,6 +17,9 @@ x-content-publisher: &content-publisher
     - content-publisher-tmp:/govuk/content-publisher/tmp
     - content-publisher-node-modules:/govuk/content-publisher/node_modules
   working_dir: /govuk/content-publisher
+  # Mount /tmp as a tmpfs volume. This is a workaround until docker/for-linux#1015 is resolved.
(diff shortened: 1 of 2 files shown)


kevindew:
Nice work @ollietreend ⭐  - great to hear there's an alternative option to the rather grim ones I presented in: https://github.com/alphagov/govuk-docker/issues/537

I don't know too much about tmpfs to know what the trade-offs are here. Is the /tmp mount localised to an individual container? i.e. has an ephemeral lifespan that matches the container; or does it persist? does it get shared between containers at all?

ollietreend:
@kevindew as I understand it, the [`tmpfs` option](https://docs.docker.com/compose/compose-file/compose-file-v3/#tmpfs) creates a RAM-backed filesystem mount which is localised to each individual container, and isn't persisted or shared between different instances of containers. So completely isolated and ephemeral.

I've just had a quick play with some containers running in parallel and that's the behaviour I observed.

kevindew:
That sounds great - let's give it a go. Could you write a reply to the GitHub issue just to indicate there's a workaround?

PR #540, opened 2021-12-07
Use Postgres 13 for Docker development

1pretz1:
Uses Postgres 13 instead of 9.6. All tests pass and the app runs
successfully. Content Data Admin only used one table (users).

Trello:

* https://trello.com/c/QQf4LjT6/14-content-data-admin

* https://trello.com/c/yaBLaC1O/957-upgrade-content-data-admin-postgres-96-13


hannako (approved):
LGTM

PR #543, opened 2021-12-08
Update locations-api to use postgres-13

MuriloDalRi:
We'll need to update production to use this version as well.

https://trello.com/c/IQPrLHtP/2809-upgrade-postgres-in-locations-api-3


ChrisBAshton (approved):
✅

PR #544, opened 2021-12-08
Implement restoring dumps for postgres-13 and mysql-8

barrucadu:
Restoring postgres-9.6 dumps into postgres-13 doesn't quite work without a `DROP SCHEMA IF EXISTS "public"` query - but adding that query in makes restoring to postgres-9.6 fail!  I guess the fundamental problem here is `pg_restore` making assumptions about the output of `pg_dump`, so I've had to add a conditional here.

MySQL just worked.

I've tested this by restoring dumps for email-alert-api, link-checker-api, collections-publisher, and search-admin.  All were able to load data successfully, and it was visible from a rails console.

This approach of finding the container name will work for elasticsearch and mongo too, but since we don't have different database versions to test, I didn't go ahead an implement aht.

---

[Trello card](https://trello.com/c/B8k5OtGe/29-configure-govuk-docker-to-replicate-databases-to-correct-engine-version)

### DIFF
--- a/bin/replicate-mysql.sh
+++ b/bin/replicate-mysql.sh
@@ -42,16 +42,17 @@ fi
 echo "stopping running govuk-docker containers..."
 govuk-docker down
 
-govuk-docker up -d mysql-5.5
-trap 'govuk-docker stop mysql-5.5' EXIT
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Great job @barrucadu - nice thinking with the use of Ruby for parsing the YAML and getting past the schema error. 

I guess we may find PostgreSQL has future surprises in store for us once the infrastructure is creating dumps from a newer Postgres but 🤞

PR #550, opened 2021-12-17
Add docs for mysql `caching_sha2_password` default

huwd:
This stumped me for a couple of days.

Since the mysql 8.0 update I had been unable to `make whitehall` seeing errors when i tried to run: `govuk-docker run whitehall-lite bundle exec rails db:prepare`.

The stack trace was as follows:

```
ActiveRecord::ConnectionNotEstablished: Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or dire
ctory                                                                                                                                                                                                              
/govuk/whitehall/app/models/edition/translatable.rb:19:in `block in <module:Translatable>'                                                                                                                         
/govuk/whitehall/app/models/edition.rb:23:in `include'                                                                                                                                                             
/govuk/whitehall/app/models/edition.rb:23:in `<class:Edition>'                                                                                                                                                     
/govuk/whitehall/app/models/edition.rb:3:in `<main>'                                                                                                                                                               
/govuk/whitehall/app/models/publicationesque.rb:9:in `<main>'                                                                                                                                                      
/govuk/whitehall/app/models/publication.rb:8:in `<main>'                                                                                                                                                           
/govuk/whitehall/app/helpers/admin/edition_routes_helper.rb:4:in `<module:EditionRoutesHelper>'                                                                                                                    
/govuk/whitehall/app/helpers/admin/edition_routes_helper.rb:1:in `<main>'                                                                                                                                          
/govuk/whitehall/config/initializers/teaspoon.rb:2:in `<main>'                                                                                                                                                     
/govuk/whitehall/config/environment.rb:5:in `<main>'                                                                                                                                                               
                                                                                                                                                                                                                   
Caused by:                                                                                                                                                                                                         
Mysql2::Error::ConnectionError: Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or directory   
/govuk/whitehall/app/models/edition/translatable.rb:19:in `block in <module:Translatable>'                                                                                                                         
/govuk/whitehall/app/models/edition.rb:23:in `include'                                                                                                                                                             
/govuk/whitehall/app/models/edition.rb:23:in `<class:Edition>'                                                                                                                                                     
/govuk/whitehall/app/models/edition.rb:3:in `<main>'                                                                                                                                                               
/govuk/whitehall/app/models/publicationesque.rb:9:in `<main>'                                                                                                                                                      
/govuk/whitehall/app/models/publication.rb:8:in `<main>'                                                                                                                                                           
/govuk/whitehall/app/helpers/admin/edition_routes_helper.rb:4:in `<module:EditionRoutesHelper>'                                                                                                                    
/govuk/whitehall/app/helpers/admin/edition_routes_helper.rb:1:in `<main>'                                                                                                                                          
/govuk/whitehall/config/initializers/teaspoon.rb:2:in `<main>'                                                                                                                                                     
/govuk/whitehall/config/environment.rb:5:in `<main>'                                                                                                                                                               
Tasks: TOP => db:prepare => db:load_config => environment                                                                                                                                                          
(See full trace by running task with --trace)
```

This seems to be related to a change in [MySQL password default config][1] in v8.
This documents a solution, (with massive thanks to @ChrisBAshton and @brucebolt)

We may want to find a more sustainable solution as part of the MySQL docker setup. 
[I have created an issue here for that](https://github.com/alphagov/govuk-docker/issues/551).

[1]: https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -113,3 +113,19 @@ To check if you're using compose V2 and turn it off:
 - go to "Experimental Features"
 - check if the "Use Docker Compose V2 release candidate" checkbox is checked
 - uncheck it if it is checked and save
+
+## Cannot `rails db:prepare` or start console due to "Plugin caching_sha2_password could not be loaded"
+
+In MySQL 8.0 `caching_sha2_password` was made the default over the previous `mysql_native_password`.
+
+This can lead to the following error when ActiveRecord is attempting to connect to the database, for example when running `rails db:prepare` or trying to bring up a rails console.
+
+```
+ActiveRecord::ConnectionNotEstablished: Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or directory
+```
+
+A workaround is to get MySQL to fall back to using `mysql_native_password` as follows:
+
+- Check that you can see `govuk-docker_mysql-8_1` when running `govuk-docker ps`, if not you will need to start a service that uses mysql (for example Whitehall).
+- Bring up a mysql console inside the container: `docker exec -it govuk-docker_mysql-8_1 mysql --user=root --password=root`
+- Alter the way the root user identifies itself. `ALTER USER 'root' IDENTIFIED WITH mysql_native_password BY 'root';`


docs/troubleshooting.md
```diff
@@ -113,3 +113,28 @@ To check if you're using compose V2 and turn it off:
...
+
+This can lead to the following error when ActiveRecord is attempting to connect to the database, for example when running `rails db:prepare` or trying to bring up a rails console.
+
+```
+ActiveRecord::ConnectionNotEstablished: Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or directory
+```
+
+A current workaround is to get mysql to fall back to using `mysql_native_password` as follows:
+
+- Check that you can see `govuk-docker_mysql-8_1` when running `govuk-docker ps`, if not you will need to start a service that uses mysql.
```
alex9smith, line 138: Could you add an example of an app that uses MySQL?

alex9smith (approved):
LGTM, one non-blocking request

docs/troubleshooting.md
```diff
@@ -113,3 +113,28 @@ To check if you're using compose V2 and turn it off:
 - go to "Experimental Features"
 - check if the "Use Docker Compose V2 release candidate" checkbox is checked
 - uncheck it if it is checked and save
+
+## Cannot rails db:prepare or start console due to "Plugin caching_sha2_password could not be loaded"
```
ChrisBAshton, line 117:
```suggestion
## Cannot `rails db:prepare` or start console due to "Plugin caching_sha2_password could not be loaded"
```

docs/troubleshooting.md
```diff
@@ -113,3 +113,28 @@ To check if you're using compose V2 and turn it off:
...
+
+In December 2021 we began work to update our database versions.
+One target of this work was to upgrade from MySQL v5.5 to v8.
+For example [see whitehall here](https://github.com/alphagov/govuk-docker/pull/546)
+
+MySQL 8.0 contains a plugin "[Caching SHA-2 Pluggable Authentication](https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html)" that implements SHA-256 hashing for user account passwords.
+ner
```
ChrisBAshton, line 124:
```suggestion
```

docs/troubleshooting.md
```diff
@@ -113,3 +113,28 @@ To check if you're using compose V2 and turn it off:
...
+
+`caching_sha2_password` requires a secure connection or an unencrypted connection that supports password exchange using an RSA key pair, which `govuk-docker` is not currently set up to provide.
+
+This can lead to the following error when ActiveRecord is attempting to connect to the database, for example when running `rails db:prepare` or trying to bring up a rails console.
+
+```
+ActiveRecord::ConnectionNotEstablished: Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or directory
+```
+
+A current workaround is to get mysql to fall back to using `mysql_native_password` as follows:
```
ChrisBAshton, line 136: We can make it less wordy:

```suggestion
A workaround is to get MySQL to fall back to using `mysql_native_password` as follows:
```

ChrisBAshton (commented):
Thanks a lot for getting this documented!

I'm not sure it needs quite such a detailed introduction (perhaps some of the context can be added to the commit message itself, rather than the doc), as most devs will just want to dive in and copy and paste the fixes. But happy to leave it in if it's useful 👍

huwd:
Great thanks both, will get those amended today

docs/troubleshooting.md
huwd: woops

docs/troubleshooting.md
```diff
@@ -113,3 +113,19 @@ To check if you're using compose V2 and turn it off:
...
+
+```
+ActiveRecord::ConnectionNotEstablished: Plugin caching_sha2_password could not be loaded: /usr/lib/x86_64-linux-gnu/mariadb19/plugin/caching_sha2_password.so: cannot open shared object file: No such file or directory
+```
+
+A workaround is to get MySQL to fall back to using `mysql_native_password` as follows:
+
+- Check that you can see `govuk-docker_mysql-8_1` when running `govuk-docker ps`, if not you will need to start a service that uses mysql (for example Whitehall).
+- Bring up a mysql console inside the container: `docker exec -it govuk-docker_mysql-8_1 mysql --user=root --password=root`
+- Alter the way the root user identifies itself. `mysql> ALTER USER 'root' IDENTIFIED WITH mysql_native_password BY 'root';`
```
ChrisBAshton, line 131: Nitpick (as we're already in the console at this point):

```suggestion
- Alter the way the root user identifies itself. `ALTER USER 'root' IDENTIFIED WITH mysql_native_password BY 'root';`
```

ChrisBAshton (approved):
LGTM 👍  one non-blocking comment if you fancy addressing before merging. Shouldn't need  a re-review from me.

docs/troubleshooting.md
huwd: Ah yeah, thought about that at the time, was shooting for "don't punch this into bash"- But you are right on reflection, a dev just wants a copy and paste and this'll get in the way.

PR #552, opened 2021-12-17
Fix: New line syntax in make router

huwd:
Before this commit on a linux machine I was encountering errors with
running:

```
make router
```

The error referenced illegal characters, apparently due to the `\` at
the end of each line of mongo db initiate configuration not being parsed
correctly.

I tested this by moving the entire statement onto a single (ugly) line,
after which it worked.

The solution here is to swap double and single quotes in the statement.
I'm unclear why I've found this to be an issue on my machine but
colleagues have not on MacOs.

However this should work on both operating systems so I am suggesting it
here.


chao-xian (approved):
Running on macOS is fine too so no harm in this change

PR #555, opened 2022-01-13
Pull Postgres dumps from new DB Admin instances

ollietreend:
This changes the Postgres replication script to expect database dump files to be present in a new location now that we've implemented [RFC-143] and have one RDS instance per application.

Each Postgres server (RDS instance) is named with the app it serves.
> e.g. the app `content-tagger` uses RDS instance `content-tagger-postgres`

Database dumps are now located inside a directory named after the RDS instance.
> e.g. `content-tagger` database dumps will now be located in the directory `s3://govuk-integration-database-backups/content-tagger-postgres/`

The script will pick the most recent timestamped database dump from that instance's directory.

### Fix for Support API

The Support API uses a non-conventional database name (`support_contacts_production` rather than `support_api_production`). The script previously tried to handle that as a special case (although it didn't actually work as expected). Now the script only relies on knowing the RDS instance name rather than the database name, so it no longer needs this exception to be hardcoded.

### Fix for Transition/Bouncer

The Transition/Bouncer database previously wasn't possible to import because its dumps existed outside the hardcoded `postgres-backend` directory in the S3 bucket. With the change to hostname-based directories, this should implicitly resolve the limitation of not being able to replicate this database.

### Local database names

In the docker environment, most databases are conventionally named after their application.

> e.g. the database used by Content Tagger is called `content-tagger`:
>
> https://github.com/alphagov/govuk-docker/blob/e3958a7c273316a0d28fcd9dd985cfa5063c0569/projects/content-tagger/docker-compose.yml#L28

This isn't always the case, though. In particular, the [Transition](https://github.com/alphagov/govuk-docker/blob/e3958a7c273316a0d28fcd9dd985cfa5063c0569/projects/transition/docker-compose.yml#L28) and [Bouncer](https://github.com/alphagov/govuk-docker/blob/e3958a7c273316a0d28fcd9dd985cfa5063c0569/projects/bouncer/docker-compose.yml#L31) apps share a database called `transition`.

Previously the script assumed the database was named after the app. But this causes issues in the case of Transition/Bouncer. I've therefore tweaked the script to read the application's database name directly from the `DATABASE_URL` environment variable in the Docker Compose config.

### Clean up Postgres 9.6 conditional

I've removed a conditional that did some special magic when restoring Postgres 9.3 dumps into a Postgres 13 instance. This shouldn't be needed any more since all environments are now running on Postgres 13.

Trello ticket: https://trello.com/c/RKe0EO4x/80-configure-govuk-docker-to-replicate-data-using-new-db-admin-machines

[RFC-143]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md

### DIFF
--- a/bin/replicate-postgresql.sh
+++ b/bin/replicate-postgresql.sh
@@ -81,4 +81,4 @@ database="$(govuk-docker config | ruby -ryaml -e "puts YAML::load(STDIN.read).di
 
 govuk-docker run "$postgres_container" /usr/bin/psql -h "$postgres_container" -U postgres -c "DROP DATABASE IF EXISTS \"${database}\""
 govuk-docker run "$postgres_container" /usr/bin/createdb -h "$postgres_container" -U postgres "$database"
-pv "$archive_path" | govuk-docker run "$postgres_container" /usr/bin/pg_restore -h "$postgres_container" -U postgres -d "$database" --no-owner
+pv "$archive_path" | govuk-docker run "$postgres_container" /usr/bin/pg_restore -h "$postgres_container" -U postgres -d "$database" --no-owner --no-privileges
--- a/bin/replicate-postgresql.sh
+++ b/bin/replicate-postgresql.sh
@@ -1,10 +1,31 @@
 #!/usr/bin/env bash
 
 function try_find_file {
+  app="$1"
+
+  # Work out the database hostname
+  case "$app" in
+    "bouncer"|"transition")
+      # Bouncer and Transition share a database with a non-standard hostname (ending with "postgresql" not "postgres")
+      db_hostname="transition-postgresql"
+      ;;
+    "content-data-api")
+      # Content Data API has a non-standard hostname (ending with "postgresql" not "postgres")
+      db_hostname="content-data-api-postgresql"
+      ;;
+    *)
+      db_hostname="${app}-postgres"
+      ;;
+  esac
+
   set +e
-  out="$(aws s3 ls "s3://${bucket}/postgresql-backend/" | grep "${1}_production.gz" | sed 's/.* //' | sort | tail -n1)"
+  # Find the most recent database dump from that host
+  out="$(aws s3 ls "s3://${bucket}/${db_hostname}/" | grep "\.gz$" | sed 's/.* //' | sort | tail -n1)"
   set -e
-  echo "$out"
+
+  if [[ -n "$out" ]]; then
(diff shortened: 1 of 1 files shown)


bin/replicate-postgresql.sh
```diff
@@ -66,16 +76,9 @@ until govuk-docker run "$postgres_container" /usr/bin/psql -h "$postgres_contain
...
-if [[ "$postgres_container" != "postgres-9.6" ]]; then
-  # Restoring a dump created with postgres 9.6 into postgres 13 gives
-  # an error about the "public" schema already existing.  The SQL dump
-  # includes a CREATE SCHEMA line, so I don't know why it's fine with
-  # older postgres.  We can drop this conditional and line when we're
-  # only replicating data dumped with a newer pg_dump.
-  govuk-docker run "$postgres_container" /usr/bin/psql -h "$postgres_container" -U postgres -d "$database" -c 'DROP SCHEMA IF EXISTS "public"'
-fi
-pv "$archive_path" | govuk-docker run "$postgres_container" /usr/bin/pg_restore -h "$postgres_container" -U postgres -d "$database" --no-owner
+pv "$archive_path" | govuk-docker run "$postgres_container" /usr/bin/pg_restore -h "$postgres_container" -U postgres -d "$database" --no-owner --no-privileges
```
ChrisBAshton, line 84: It took me a minute to figure out what had changed about this line. Can you explain why `--no-privileges` was added? Should it maybe be encapsulated in its own commit?

ollietreend: Yes – this probably should be its own commit. I'll split it out. Thanks for the nudge! 😄

When restoring Postgres 13 dumps into a Postgres 13 docker container, I was getting an error:

```
pg_restore: error: could not execute query: ERROR: role "rdsadmin" does not exist
```

I believe `pg_restore` was trying to reinstate some permissions on an RDS-specific user role which doesn't exist in the Docker environment. I think this must be a new behaviour (possibly something new in RDS) because it doesn't look like this was necessary with Postgres 9.6.

Essentially, our interest is in importing the database schema and data – not the various user roles and permissions that exist in RDS.

Just to be clear, this seems to be independent of the conditional `DROP SCHEMA IF EXISTS "public"` that was needed when moving between Postgres versions.

ChrisBAshton (approved):
Thanks for splitting it out into two commits 👍

PR #558, opened 2022-01-17
Set TEST_DATABASE_URL for collections-publisher

barrucadu:
The app does not use the DATABASE_URL env var to generate the test
one: https://github.com/alphagov/collections-publisher/blob/main/config/database.yml


kevindew (approved):
Nice catch

PR #564, opened 2022-01-26
Put dnsmasq config files in /opt/homebrew/etc

KludgeKML:
### Why
On M1 machines, even when you can get an app running in govuk-docker, you can't currently access it via the .dev.gov.uk hostname in your browser. 

### What's the problem
We're using dnsmasq to resolve those hostnames into 127.0.0.1, which the nginx-docker proxy then picks up and routes to the backing app. The existing bin/setup script puts the configuration files for dnsmasq in /usr/local/etc. But on M1 machines /usr/local/etc doesn't exist, and homebrew dnsmasq looks in /opt/homebrew/etc, where the dnsmasq config files and directories already exist. This change updates those paths in bin/setup.

### Any other notes
This problem wasn't caught by Doctor, which only checks:
- is dnsmasq running
- does the first config file (/etc/resolver/dev.gov.uk) exist?
..in this case, both of those were true, but without the other two config files nothing would work. We could perhaps update doctor to make an actual dns request (via dig maybe?) to check that the configuration is working as expected.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+brew_root=$(brew --prefix)
+
 echo -e "✅ Installing dependencies via Homebrew...\n"
 brew bundle --file="$(dirname $0)/../Brewfile"
 echo
@@ -12,12 +14,12 @@ echo -e "You may need to enter your root password...\n"
 sudo mkdir -p /etc/resolver
 echo -e "nameserver 127.0.0.1\nport 53" | sudo tee /etc/resolver/dev.gov.uk >/dev/null
 
-echo -e "✅ Configuring dnsmasq [/usr/local/etc/dnsmasq.conf]\n"
-echo "conf-dir=/usr/local/etc/dnsmasq.d,*.conf" > /usr/local/etc/dnsmasq.conf
+echo -e "✅ Configuring dnsmasq [$brew_root/etc/dnsmasq.conf]\n"
+echo "conf-dir=$brew_root/etc/dnsmasq.d,*.conf" >> $brew_root/etc/dnsmasq.conf
 
-echo -e "✅ Configuring dnsmasq [/usr/local/etc/dnsmasq.d/development.conf]\n"
-mkdir -p /usr/local/etc/dnsmasq.d
-echo "address=/dev.gov.uk/127.0.0.1" > /usr/local/etc/dnsmasq.d/development.conf
+echo -e "✅ Configuring dnsmasq [$brew_root/etc/dnsmasq.d/development.conf]\n"
+mkdir -p $brew_root/etc/dnsmasq.d
+echo "address=/dev.gov.uk/127.0.0.1" > $brew_root/etc/dnsmasq.d/development.conf
 
 echo "✅ Restarting dnsmasq"
 echo -e "You may need to enter your root password...\n"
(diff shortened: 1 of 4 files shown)


Tetrino (approved):
Looks great for now - hopefully we can do some architecture sniffing in the future but let's get this running first.

bin/setup
```diff
@@ -12,12 +12,11 @@ echo -e "You may need to enter your root password...\n"
 sudo mkdir -p /etc/resolver
 echo -e "nameserver 127.0.0.1\nport 53" | sudo tee /etc/resolver/dev.gov.uk >/dev/null
 
-echo -e "✅ Configuring dnsmasq [/usr/local/etc/dnsmasq.conf]\n"
-echo "conf-dir=/usr/local/etc/dnsmasq.d,*.conf" > /usr/local/etc/dnsmasq.conf
+echo -e "✅ Configuring dnsmasq [/opt/homebrew/etc/dnsmasq.conf]\n"
```
ollietreend, line 15: I think it should be possible to make this cross-architecture compatible by using `brew --prefix` instead of hardcoding the path:

```suggestion
echo -e "✅ Configuring dnsmasq [$(brew --prefix)/etc/dnsmasq.conf]\n"
```

KludgeKML: Ooh, interesting. I'd assumed we'd do a bit of cross-architecture sniffing when we merged into full main, but this is very low-cost, worth doing now. I'll update and confirm it works on this machine, if someone else can then try it on an intel machine.

KludgeKML:
@ollietreend 's change made, confirmed works on M1 machine.

kevindew:
Great @KludgeKML could we branch this to go into the main branch instead of the M1 one? I imagine we only want to use that branch for anything that isn't interoperable.

bin/setup
```diff
@@ -12,12 +12,11 @@ echo -e "You may need to enter your root password...\n"
 sudo mkdir -p /etc/resolver
 echo -e "nameserver 127.0.0.1\nport 53" | sudo tee /etc/resolver/dev.gov.uk >/dev/null
 
-echo -e "✅ Configuring dnsmasq [/usr/local/etc/dnsmasq.conf]\n"
-echo "conf-dir=/usr/local/etc/dnsmasq.d,*.conf" > /usr/local/etc/dnsmasq.conf
+echo -e "✅ Configuring dnsmasq [$(brew --prefix)/etc/dnsmasq.conf]\n"
+echo "conf-dir=$(brew --prefix)/etc/dnsmasq.d,*.conf" >> $(brew --prefix)/etc/dnsmasq.conf
 
-echo -e "✅ Configuring dnsmasq [/usr/local/etc/dnsmasq.d/development.conf]\n"
-mkdir -p /usr/local/etc/dnsmasq.d
```
ollietreend, line 19: Doesn't this directory still need to be created? Or was it already created by something else – and therefore unnecessary here?

The directory won't be auto-created if it doesn't already exist:

```bash
$ echo "hello" > missing-dir/file.txt
zsh: no such file or directory: missing-dir/file.txt
```

KludgeKML: Oh, that's a good point. It _doesn't_ need to be created in the /opt/homebrew/etc case (there's already one there). But I can put in the mkdir -p line safely for Intel systems, as you say.

KludgeKML:
@kevindew I can either change this PR to main, or I can close it and reopen on main (maybe with the 3 commits squished into one?). What's the usual preferred option?

kevindew:
@KludgeKML I'd change the base of this PR to main (can be done by hitting edit in the top right corner if you're not already aware).

ollietreend (approved):
This works on my Intel-based Mac 👍🏻

PR #568, opened 2022-02-02
Adds docker as a prerequisite

DilwoarH:
## What

Updates ReadMe doc to include docker as a prerequisite

## Why

It may be an obvious thing to do but should be explicitly stated so users are not confused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ The aim of govuk-docker is to make it easy to develop any GOV.UK app. It achieve
 
 - [brew](https://brew.sh/)
 - [git](https://git-scm.com)
+- [Docker](https://www.docker.com/get-started)
 - A `govuk` directory in your home directory
 
 **Next, add the following line to your bash/zsh config.**


ChrisBAshton (approved):

ChrisBAshton:
(Welcome back 👋 )

kevindew:
As I understand this isn't needed because the Brewfile installs docker: https://github.com/alphagov/govuk-docker/blob/ff55aa1698eccc8ab0ad926c50dae2c6b9516b71/Brewfile#L8 and by changing this we create an inconsistency between some dependencies listed here and others in the Brewfile.

DilwoarH:
> As I understand this isn't needed because the Brewfile installs docker:
> 
> https://github.com/alphagov/govuk-docker/blob/ff55aa1698eccc8ab0ad926c50dae2c6b9516b71/Brewfile#L8
> 
> and by changing this we create an inconsistency between some dependencies listed here and others in the Brewfile.

This is a good point, I'll close the PR

PR #569, opened 2022-02-02
Force docker builds to use linux/amd64 platform

DilwoarH:
This change forces the builds to create images for linux/amd64 platform.

M1 Macs use a arm archtecture and this is generally not *currently* compatible with the packages we need to install, ie. Chrome.

This change allows builds to successfully complete.


DilwoarH:
> @DilwoarH I suspect we can just close this now, right? 

Yes

PR #572, opened 2022-02-04
Remove warnings about db versions in production

beccapearce:
We added warnings about the database version in docker not matching 
those in production whilst we completed an upgrade. The upgrade has now 
been done so these warnings can be removed.


ChrisBAshton (approved):
⭐

PR #573, opened 2022-02-04
Make replication scripts work with Docker Compose v2

ollietreend:
This commit changes the database replication scripts to make them compatible with Docker Compose v2. This doesn't affect backwards compatibility with Docker Compose v1 – this project should now be version agnostic.

The `docker-compose run` command allocates a TTY by default. However this was causing problems when piping commands and files in to the container. Adding the `-T` flag tells Docker not to allocate a TTY, making the command non-interactive, which seems to resolve the problem.

I'm not entirely sure why Docker Compose v1 was forgiving of us excluding this flag, but it seems that v2 needs it to be set.

---

I also experienced some race-condition type issues where `docker down` failed to remove some containers because they were already being removed.

Output like this would be quite common:
```
Error response from daemon: removal of container {container_id} is already in progress
```

I don't understand exactly why this happened, but it seems that asking each individual `docker run` to remove itself with the `--rm` flag has stopped those error messages.

### DIFF
--- a/bin/replicate-mysql.sh
+++ b/bin/replicate-mysql.sh
@@ -55,16 +55,16 @@ govuk-docker down
 
 mysql_container="$(govuk-docker config | ruby -ryaml -e "puts YAML::load(STDIN.read).dig('services', '${app}-lite', 'depends_on').keys.select { |k| k.start_with? 'mysql-' }")"
 govuk-docker up -d "$mysql_container"
-trap 'govuk-docker stop ${mysql_container}' EXIT
+trap 'govuk-docker down' EXIT
 
 echo "waiting for mysql..."
-until govuk-docker run "$mysql_container" mysql -h "$mysql_container" -u root --password=root -e 'SELECT 1' &>/dev/null; do
+until govuk-docker run --rm -T "$mysql_container" mysql -h "$mysql_container" -u root --password=root -e 'SELECT 1' &>/dev/null; do
   sleep 1
 done
 
 # Extract the local database name from the app's DATABASE_URL environment variable
 database="$(govuk-docker config | ruby -ryaml -e "puts YAML::load(STDIN.read).dig('services', '${app}-app', 'environment', 'DATABASE_URL').split('/').last")"
 
-govuk-docker run "$mysql_container" mysql -h "$mysql_container" -u root --password=root -e "DROP DATABASE IF EXISTS \`${database}\`"
-govuk-docker run "$mysql_container" mysql -h "$mysql_container" -u root --password=root -e "CREATE DATABASE \`${database}\`"
-pv "$archive_path" | gunzip | govuk-docker run "$mysql_container" mysql -h "$mysql_container" -u root --password=root "$database"
+govuk-docker run --rm -T "$mysql_container" mysql -h "$mysql_container" -u root --password=root -e "DROP DATABASE IF EXISTS \`${database}\`"
+govuk-docker run --rm -T "$mysql_container" mysql -h "$mysql_container" -u root --password=root -e "CREATE DATABASE \`${database}\`"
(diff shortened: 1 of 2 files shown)


kevindew:
Nice one, thanks for digging into this. Before looking deeply I wondered if we need to amend the elasticsearch and MongoDB scripts too?

ollietreend:
@kevindew I did look at the Mongo and ES scripts, but they seem to take different approaches to getting content into the container which aren't affected by the move to Docker Compose v2.

MySQL and Postgres are the only scripts which use `pv` to pipe data in – which seems to be what was causing problems.

kevindew:
Great stuff, code looks good to me. I've still not been able to replicate the problem so I think it's best to wait until Monday so @KludgeKML can try it out before merging.

Oddly even when I have Docker Compose v2 it seems to work for me on main branch 🤷  :

```
➜  govuk-docker git:(main) git pull
Already up to date.
➜  govuk-docker git:(main) govuk-docker version
docker-compose -f [...] version
Docker Compose version v2.1.1
➜  govuk-docker git:(main) gds aws govuk-integration-readonly ./bin/replicate-postgresql.sh service-manual-publisher
Touch your YubiKey...
Replicating postgres for service-manual-publisher
download: s3://govuk-integration-database-backups/service-manual-publisher-postgres/2022-02-04T05:00:01-service-manual-publisher_production.gz to replication/postgresql/service_manual_publisher_production.dump.gz
stopping running govuk-docker containers...
docker-compose -f [...] down
[+] Running 1/0
 ⠿ govuk-docker  Warning: No resource found to remove                      0.0s
docker-compose -f [...] config
docker-compose -f [...] up -d postgres-13
[+] Running 2/2
 ⠿ Network govuk-docker_default          Created                           0.0s
 ⠿ Container govuk-docker-postgres-13-1  Started                           0.4s
waiting for postgres...
docker-compose -f [...] config
docker-compose -f [...] run postgres-13 /usr/bin/psql -h postgres-13 -U postgres -c DROP DATABASE IF EXISTS "service-manual-publisher"
DROP DATABASE
docker-compose -f [...] run postgres-13 /usr/bin/createdb -h postgres-13 -U postgres service-manual-publisher
docker-compose -f [...] run postgres-13 /usr/bin/pg_restore -h postgres-13 -U postgres -d service-manual-publisher --no-owner --no-privileges
11.6MiB 0:00:02 [4.50MiB/s] [================================>] 100%
docker-compose -f [...] stop postgres-13
[+] Running 5/5
 ⠿ Container govuk-docker-postgres-13-1                 Stopped            0.3s
 ⠿ Container govuk-docker_postgres-13_run_13ced638217f  Stopped            0.1s
 ⠿ Container govuk-docker_postgres-13_run_1d5636efc1b6  Stopped            0.0s
 ⠿ Container govuk-docker_postgres-13_run_78c9bd69cc3b  Stopped            0.1s
 ⠿ Container govuk-docker_postgres-13_run_e90a719f2638  Stopped            0.1s
```

ollietreend:
> Oddly even when I have Docker Compose v2 it seems to work for me on main branch 🤷

Oh, weird! I've been able to reproduce it on my Intel machine with Docker Compose v2 on the `main` branch.

Makes sense to see if it solves the problem for @KludgeKML 👍🏻

KludgeKML:
Looking good so far - I've synced service-manual-publisher successfully, and publishing-api is running now (it's about 30 mins just to import, but it's actually got to the import, so that's good).

KludgeKML:
publishing-api replication worked well, so looking good.

ollietreend:
Great – thanks for testing it @KludgeKML 🌟

Can you please approve the PR so it can be merged in?

KludgeKML (approved):
Works well for me.

PR #576, opened 2022-02-14
[WIP] Change how chrome is installed

DilwoarH:
Currently chrome is installed by downloading a specific amd64 version. This breaks on M1 macs as they do not have the same amd architecture.

To resolve this chromium is installed but with the additional support for both amd and arm architectures.


ChrisBAshton:
Superseded by #577

PR #579, opened 2022-03-09
[DO NOT MERGE] Remove mongo 2.6 container, replace 2.6 references with 3.6

KludgeKML:
## WHAT

- Remove the mongo 2.6 container
- Fixup router and router-api to use the 3.6 container.

## WHY

M1 laptops can't run mongo 2.6 in arm64 mode, and since router and router-api are due to be moved soon to a Mongo 3.6-compliant DocumentDB cluster, they won't be using it in production. The other apps that currently use 2.6 (imminence / authenticating proxy) already use mongo 3.6 in govuk-docker, so there's precedent for the docker image preceding production in the versions of backing apps used.

### DIFF
--- a/bin/replicate-mongodb.sh
+++ b/bin/replicate-mongodb.sh
@@ -23,7 +23,6 @@ case "$app" in
   "router"|"draft-router")
     hostname=router_backend
     database="${app//-/_}"
-    mongo_version=2.6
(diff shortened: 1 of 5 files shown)


KludgeKML:
Obsolete.

PR #585, opened 2022-04-12
Fixes/optimisations for `bin/doctor`

ChrisBAshton:
See commits for details.


lib/govuk_docker/doctor/checkup.rb
```diff
@@ -77,7 +77,7 @@ def run_state_message
     end
 
     def running_as_different_user?
-      @running_as_different_user ||= system "ps aux | grep `pgrep #{service_name}` | grep -v `whoami` 1>/dev/null"
+      @running_as_different_user ||= system "ps aux | grep `pgrep #{service_name} || echo '#{service_name}_not_running'` | grep -v `whoami` 1>/dev/null"
```
kevindew, line 80: Nice work sussing this out.

This did take me rather a long time to understand what it is was doing which made me wonder if there was a simpler option. 

It seems the complicated aspect is that there's two checks going on - whether process is running and whether it's run by a the user? could the just be separated out i.e: `pgrep #{service_name}` tells you it's running and `pgrep -u "$(whoami)" #{service_name}` tells you whether the current user is running it.

kevindew (approved):
Looks good to me 👍

PR #587, opened 2022-04-14
Document issue installing `dnsmasq` (incompatibility with SSH)

ChrisBAshton:
My new Macbook has my username separated with a dot, i.e.
`christopher.ashton`. My SSH username is set as `christopherashton`.
For SSH commands to work, I'd set `export USER="christopherashton"`
in my `~/.zshrc`, but setting this variable caused the govuk-docker
`bin/setup` script to fail - specifically on the `dnsmasq`
postinstall step.

I am [not the only developer who has had this issue](https://github.com/alphagov/govuk-connect/issues/72),
and it is likely to catch people out again, so worth documenting
here.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,5 +1,18 @@
 # Troubleshooting
 
+## `bin/setup` fails to succeed
+
+If `bin/setup` fails on installing `dnsmasq`, make sure you're not overriding the `USER` env variable anywhere.
+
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
Looks good to me (although as you say in the linked issue, the workaround here isn't ideal if we can fix it with a different env var)

ChrisBAshton:
I've added a note to update the govuk-docker docs after changing that behaviour 👍

kevindew:
To be fair, changing $USER is likely a problematic idea in many situations. You might be hard pressed to document all the places it could break our software.

PR #594, opened 2022-04-26
Check on bundle if branch is main and up to date

KludgeKML:
##WHAT

Adds makefile target branch-checks-% which fetches the repo, works out 
if the head is main or master, then warns the user if they're not 
currently on that branch or if it's not up to date, giving them the 
option of fixing those cases, ignoreing them, or stopping the make.

##WHY

https://trello.com/c/BEkh2YsT/261-warn-govuk-docker-users-when-theyre-on-outdated-or-non-default-branches


ChrisBAshton (approved):
Works like a charm - thank you! I did consider whether we should have tests, but manual testing is probably best here, and there is [precedent for adding scripts without tests](https://github.com/alphagov/govuk-docker/pull/220/files), so happy to approve 👍

KludgeKML:
Yeah, I was wondering. I mean, we have checks for Doctor, but the doctor script is just a wrapper around ruby code, so we actually test the ruby rather than the script. Perhaps I should rubify this code as well, allow for a bit of testing at least.

(I should also probably change at least the messaging, because currently up-to-date is a bit misleading - it's checking if your current HEAD matches the origin, which isn't necessarily up to date - eg if you've rebased onto main and are actually technically ahead of origin that would make you more up to date, but the script suggests you aren't. I think we still want to warn people in that case, but the warning should be more precise).

Issue #589, opened 2022-04-19
Initial installation assumes rbenv but not setup


kevindew:
We have an issue that the [installation instructions](https://github.com/alphagov/govuk-docker#installation) expects a `bundle install` step before `bin/setup`, on a new machine this will use system Ruby and on GDS current dev machines this won't work due to the bundle lockfile version. Example error:

```
/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe’: Could not find ‘bundler’ (2.1.4) required by your /Users/matt.hill/govuk/govuk-docker/Gemfile.lock. (Gem::GemNotFoundException)
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:2.1.4`
	from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems.rb:302:in `activate_bin_path’
	from /usr/bin/bundle:23:in `<main>'
```

[We install `rbenv` as part of `bin/setup`](https://github.com/alphagov/govuk-docker/blob/66422629be24a2e10ebd289cd50844729461c70b/Brewfile#L2) however rbenv is not configured nor is a Ruby version installed so this doesn't really help anyone progress with a Ruby installation.

Some options I could see to proceed with this:

1) Drop `bundle install` from initial install step, you don't actually need it until you run `govuk-doctor` or the tests. This seems a bit funky as you'll have parts of the repo that might not work and need extra instructions to set them up
2) Remove `rbenv` from Brewfile and suggest a dev sets that up first with a link to rbenv instructions - this is probably the lowest effort approach.
3) Set-up rbenv as part of `bin/setup` - this is most consistent with current approach - this would mean us having to modify a ~/.zshrc file run source on it and then `rbenv install` and `bundle install`. This one is a bit of a pain for being invasive (changing the zshrc file feels like a confirm action) or assuming someone uses zsh

I'm somewhat of the view that 2 is the best option where 3 just seems like more things for us to maintain and to break, but I feel like I'm a little out of step with how far we want `bin/setup` to go (it already seems to go pretty far)

AshGDS:
Hi @kevindew 

Thanks for this.

[...]

Issue #435, opened 2020-11-17
"bin/doctor" requires Ruby and Gems to be installed


benthorner:
## Background

We refer to this utility in [our troubleshooting guidance](https://github.com/alphagov/govuk-docker/blob/679d5bf335a5f0d5aab7f7e17f6aacac90f435d5/docs/troubleshooting.md#diagnose-common-issues-when-setting-up-govuk-docker), but using it requires someone to install Rbenv, Ruby and Gems. This can take a long time if someone doesn't already have the [required](https://github.com/alphagov/govuk-docker/blob/master/.ruby-version) version of Ruby installed on their system. Having to do all these steps also makes it likely the user will need support with the support utility 😆.

## Suggestion

Since [most of the checks](https://github.com/alphagov/govuk-docker/blob/500aa5750b6b0f8ee2bd473f1b802ddb5341df44/lib/govuk_docker/doctor/checkup.rb) are just simple shell commands, we could rewrite the script [in bash](https://github.com/alphagov/govuk-docker/blob/500aa5750b6b0f8ee2bd473f1b802ddb5341df44/bin/govuk-docker#L1) to avoid the need for Ruby / Gems. Alternatively, we could raise a PR, or conduct an informal poll, to remove the utility: we should prioritise fixing problems at source, and empowering people to fix harder problems themselves, with guidance.

PR #608, opened 2022-06-20
Use GOV.UK Docker to test GOV.UK Docker

benthorner:
Fixes https://github.com/alphagov/govuk-docker/issues/589.

This adds a new stack so that GOV.UK Docker has its own Ruby
environment to run Ruby commands in.

As part of doing this, I've also tweaked the guidance so that
all the relevant setup instructions are inline with the one use
case that remains for a local Ruby environment - bin/doctor -
which should itself be removed in time [^1].

[^1]: https://github.com/alphagov/govuk-docker/issues/435


kevindew (approved):
This is a nice idea. Thanks for your work putting it together. Changes look good to me.

PR #611, opened 2022-07-18
Add flag to always update branch from origin

KludgeKML:
If you just want the latest version of whatever branch it's annoying to have the build keep stopping to prompt you, so now setting `GOVUK_DOCKER_UPDATE_BRANCH=always` (either permanently in your environment or temporarily at the point you call make) will always update from origin whatever branch you're on without bothering you.

### DIFF
--- a/bin/branch_checks.sh
+++ b/bin/branch_checks.sh
@@ -40,13 +40,20 @@ fi
 
 origin_head_commit=$(git -C "${govuk_root_dir}/${app}" rev-parse "@{u}")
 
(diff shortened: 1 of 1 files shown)


chao-xian (approved):

ollietreend:
Great work @KludgeKML! This very feature was on a personal to-do list of mine – so it's great to see it added 🚀

KludgeKML:
Thanks! If you have time, one improvement I think would be to add an option to the menus so you can move to always updating even if you forget the env var. But I'll get to that eventually.

PR #618, opened 2022-08-23
Add Whitehall worker

brucebolt:
This allows us to start a Whitehall worker process, to locally run the contents of the queue.

The process won't start by default when the Whitehall app is launched in govuk-docker, as it is not needed in most occasions.

When required, the following can be used from the Whitehall directory to launch the worker:

```
govuk-docker-up worker
```


BeckaL (approved):
LGTM, looks equivalent to what we've got in other docker composes

PR #622, opened 2022-10-14
Add Redis as a dependency of Imminence

1pretz1:
Trello:
https://trello.com/c/HvagSfkT/1571-ps-14-imminence-causes-dropped-names-on-local-lookup-services


KludgeKML (approved):
Looks good to me, matches other apps with worker dependencies.

PR #623, opened 2022-10-20
Retire mapit, since it's no longer used.

KludgeKML:
- replace imminence dependency on mapit with locations-api.
- add imminence dependency on local-links-manager.

https://trello.com/c/LMRiT7yz/1581-retire-mapit-and-mapit-scripts


projects/imminence/Makefile
```diff
@@ -1,3 +1,3 @@
-imminence: bundle-imminence mapit
+imminence: bundle-imminence locations-api
```
ChrisBAshton, line 1: Ah, does Imminence have a dependency on Locations API? I thought you'd discovered it should only have a dependency on Local Links Manager instead?

KludgeKML: Updated.

brucebolt (approved):
This makes me very happy!

PR #627, opened 2022-11-18
Upgrade Node.js to version 16

MuriloDalRi:
Node.js 12 is no longer supported and 14 is only supported until April: https://github.com/nodejs/Release#release-schedule

I've tested a few apps on version 16 without any issues.

https://trello.com/c/PMsAXSoT/3029-review-govuks-usage-of-node-2

### DIFF
--- a/Dockerfile.govuk-base
+++ b/Dockerfile.govuk-base
@@ -12,7 +12,7 @@ RUN apt-get update -qq && apt-get install -y chromium chromium-driver
 ENV GOVUK_TEST_CHROME_NO_SANDBOX 1
 
 # Install node / yarn
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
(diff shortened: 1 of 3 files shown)


ops20 (approved):
LGTM

ChrisBAshton (commented):
From the PR description it sounds like a good next step would be to `make` these 3 apps. If they work, then we can try bumping to v16 and opening a PR for that. Feels like we can do that rather than merging this intermediary PR.

MuriloDalRi:
> From the PR description it sounds like a good next step would be to `make` these 3 apps. If they work, then we can try bumping to v16 and opening a PR for that. Feels like we can do that rather than merging this intermediary PR.

I'll try that.

MuriloDalRi:
@ChrisBAshton it's ready for review again

PR #630, opened 2022-12-06
Replace MongoDB with Postgres for Imminence

KludgeKML:
Updates Imminence to use PG+PostGIS rather than MongoDB. 

Related PRS:
https://github.com/alphagov/imminence/pull/766

Trello:
https://trello.com/c/R6Lb3vYa/1611-move-imminence-away-from-mongodb-onto-postgres

### DIFF
--- a/projects/imminence/docker-compose.yml
+++ b/projects/imminence/docker-compose.yml
@@ -20,25 +20,24 @@ services:
   imminence-lite:
     <<: *imminence
     depends_on:
-      # In production this uses Mongo 2.6, however there is not a published Mongo 2.6 image compatible with ARM64
-      - mongo-3.6
+      - postgres-14-postgis
       - redis
     environment:
-      MONGODB_URI: "mongodb://mongo-3.6/imminence"
-      TEST_MONGODB_URI: "mongodb://mongo-3.6/imminence-test"
+      DATABASE_URL: "postgis://postgres@postgres-14-postgis/imminence"
+      TEST_DATABASE_URL: "postgis://postgres@postgres-14-postgis/imminence-test"
       REDIS_URL: redis://redis
 
   imminence-app: &imminence-app
     <<: *imminence
     depends_on:
-      - mongo-3.6
+      - postgres-14-postgis
       - local-links-manager-app
       - locations-api-app
       - nginx-proxy
       - redis
       - imminence-worker
     environment:
-      MONGODB_URI: "mongodb://mongo-3.6/imminence"
+      DATABASE_URL: "postgis://postgres@postgres-14-postgis/imminence"
       VIRTUAL_HOST: imminence.dev.gov.uk
       BINDING: 0.0.0.0
       REDIS_URL: redis://redis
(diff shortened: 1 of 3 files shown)


projects/imminence/docker-compose.yml
```diff
@@ -17,28 +18,39 @@ x-imminence: &imminence
   working_dir: /govuk/imminence
 
 services:
+  imminence-pg:
```
kevindew, line 21: Does this need to be imminence specific - could it be postgres-14-postgis ? and be with the other db services?

projects/imminence/docker-compose.yml
```diff
@@ -17,28 +18,39 @@ x-imminence: &imminence
...
+    image: ghcr.io/baosystems/postgis:14-3.2
```
kevindew, line 22: Do we need to use github container registry for this? I'm somewhat guessing you've had to go this way to get arm64 and amd64 containers but not sure. I think all our other images are dockerhub ones currently.

projects/imminence/docker-compose.yml
```diff
@@ -17,28 +18,39 @@ x-imminence: &imminence
...
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
+      POSTGRES_PASSWORD: pgsecret
```
kevindew, line 25: Do we need a password? I think we don't for our existing postgres container.

kevindew (commented):
It's super exciting to see this transition to postgres.

Sorry for adding an early review, I know it's still a draft. I've just added some questions suggestions about the DB for consistency with existing approaches.

projects/imminence/docker-compose.yml
KludgeKML: It could probably be with the others, I'll move it and see what falls out.

projects/imminence/docker-compose.yml
KludgeKML: Hmm. At the time I was initially making this repo, this was the only one available that would work. But it may be that it's changed now. I'll check what's available - it's this version of the image that requires the password, too, so it's possible that if we change it we might not need the password.

projects/imminence/docker-compose.yml
KludgeKML: We needed it for the old container, but if we're okay with the new one, it's unnecessary.

projects/imminence/docker-compose.yml
KludgeKML: The official Postgis container only comes in AMD64 versions, so is unusable for local dev on M1 machines. I've replaced this with a dockerhub container that seems reasonably up-to-date, but it's "just some person". I guess that's probably not too bad a thing, given that most of the open source packages we use are just some person, but I don't know if other people might have a more solid opinion than mine, which is just "eh... it's not ideal, but I guess it's okay?"

KludgeKML: https://hub.docker.com/r/nickblah/postgis

kevindew: Yeah I think it's in the realm of "eh... it's not ideal, but I guess it's okay?"

I'd add a comment above it explaining we're not using official one because of lack of AMD64 support.

I notice there is https://hub.docker.com/r/kartoza/postgis which seems to have a couple of orders magnitude more downloads and AMD64 images (but may have missed something that makes it unsuitable)

KludgeKML: Not sure why I didn't use that one, I saw it, my brain just glossed over it. But yeah, that also works and seems more stable, so I've updated to use that one.

KludgeKML:
We're now on PG in production, so this PR's time has come.

richardTowers (approved):

PR #634, opened 2023-01-03
Update Ruby to 3.2 and .github cleanup

kevindew:
This updates this project to the latest version of Ruby (3.2.0) to reflect that Ruby 2.7 will be reaching end-of-life in the coming months.

I did debate whether to go for a lower version of Ruby (3.1.3) since Ruby 3.2.0 is very new but as we don't update this very frequently I felt it was best to go to the latest version. The key advantage of an earlier Ruby version is that devs would likely have that already installed, but Ruby 3.1.3 is also very new so that also doesn't have that advantage and it seemed odd to update this to 3.1.2 when 3.1.3 is out.


spec/govuk_docker/doctor/checkup_spec.rb
```diff
@@ -20,9 +20,9 @@
   context "when the repository is up-to-date" do
     it "should report that it is up-to-date" do
       subject = described_class.new(
-        service_name: service_name,
+        service_name:,
```
ChrisBAshton, line 23: The [hash value omission feature](https://bugs.ruby-lang.org/issues/14579) seems to have been [added in ruby 3.1.0](https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/), so this commit should probably be `Fix rubocop violations for Ruby >= 3.1`?

.github/dependabot.yml
```diff
@@ -4,5 +4,3 @@ updates:
   directory: "/"
   schedule:
     interval: daily
-    time: "03:00"
-  open-pull-requests-limit: 10
```
ChrisBAshton, line 8: Good catch - looks like it was only applied through automation in 448868500934f923f4e93bc935366991e230b7e5.

ChrisBAshton (approved):
Nice one ⭐

spec/govuk_docker/doctor/checkup_spec.rb
kevindew: Oh yes, good catch. That's what I meant 😆

PR #639, opened 2023-02-01
Replace MongoDB with Postgres for Authenticating Proxy

KludgeKML:
Updates Authenticating Proxy to use PG rather than MongoDB.

Related PRS:
https://github.com/alphagov/authenticating-proxy/pull/350

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ volumes:
   rabbitmq:
   postgres-12:
   postgres-13:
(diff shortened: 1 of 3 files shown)


AgaDufrat (approved):
LGTM and it was already removed from the replication script in https://github.com/alphagov/govuk-docker/pull/640/commits/c998b4d9811c59b642f4d5389781128eebad3227

PR #642, opened 2023-02-22
Fixed postgres password for postgis container

KludgeKML:
The kartoza image used for postgres+postgis development on imminence requires a password.


1pretz1 (approved):
Nice!

PR #643, opened 2023-02-28
Bump router golang version to 1.20.2

Tetrino:
Matches dependabot version upgrade.

### DIFF
--- a/projects/router/Dockerfile
+++ b/projects/router/Dockerfile
@@ -1,1 +1,1 @@
-FROM golang:1.19.1
+FROM golang:1.20.2


.ruby-version
```diff
@@ -1 +1 @@
-3.2.0
```
ChrisBAshton, line 1: This looks like it slipped in. Probably should be a separate PR (especially as it's a downgrade)!

Tetrino: Good catch. It shouldn't be needed at all. It'll sit here anyway until the k9s move is done.

chao-xian (approved):

PR #661, opened 2023-05-17
Add ability to run some frontends apps against integration data

brucebolt:
This allows us to run Collections, Finder Frontend and Government Frontend against the data present in the integration environment.

We sometimes use integration to test the running of rake tasks or data migrations, so it is useful to have a local way of testing frontend changes against that data.

The integration app can be started using:

```
govuk-docker-up app-integration
```


ChrisBAshton:
Nice! 🎉 this is worthy of a note in the tech weeknotes btw :)

ChrisBAshton:
I've raised https://github.com/alphagov/govuk-developer-docs/pull/4020

PR #667, opened 2023-06-14
Fix replicate-mongodb.sh to work on linux

chrisroos:
The `\d` in the regexp passed to grep isn't supported in the GNU version of grep. I understand that the equivalent `[[:digit:]]` should work in all implementations.

The `-z` option to `tar` results in the following error on linux:

```
gzip: stdin: not in gzip format
tar: Child died with signal 13
tar: Error is not recoverable: exiting now
```

The `-z` option filters the output through gzip but I don't believe this is necessary because we're explicitly `gunzip`ing the dump before passing it to `tar`.

With these two changes the script runs without error on my linux development environment.

### DIFF
--- a/bin/replicate-mongodb.sh
+++ b/bin/replicate-mongodb.sh
@@ -49,7 +49,7 @@ if [[ -e "$archive_path" ]]; then
   echo "Skipping download - remove ${archive_path} to force a new download on the next run"
 else
   mkdir -p "$archive_dir"
-  remote_file_name=$(aws s3 ls "s3://${bucket}/${hostname}/" | grep "\d-$database.gz" | tail -n1 | sed 's/^.* .* .* //')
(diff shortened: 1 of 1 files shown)


chrisroos:
I'd appreciate someone testing these changes on a Mac to make sure I haven't broken anything there. You should be able to test with:

```
gds aws govuk-integration-readonly --assume-role-ttl 3h ./bin/replicate-mongodb.sh content-store
```

KludgeKML (approved):
Looks good to me, checked on M1 Mac (Ventura 13.4)

beccapearce:
This ran on my mac, monterey 12.6.6

chrisroos:
Awesome. Thanks @KludgeKML and @beccapearce :+1:

PR #669, opened 2023-06-20
fix mongodb replication script

kornysietsma:
change to use current location of asset-manager backup file

### DIFF
--- a/bin/replicate-mongodb.sh
+++ b/bin/replicate-mongodb.sh
@@ -33,6 +33,10 @@ case "$app" in
     hostname=mongo-api
     database="${app//-/_}_production"
     ;;
+  "asset-manager")
+    hostname=mongo-normal
+    database=govuk_assets_production
+    ;;
   *)
     hostname=mongo
     database="${app//-/_}_production"


ChrisBAshton (approved):
Thanks for fixing this!

For context (for future devs who may look back at this), we arrived at this by:

> The previous data replication process looked at a [proprietary mongodb bucket](https://github.com/alphagov/govuk-docker/blob/ff55aa1698eccc8ab0ad926c50dae2c6b9516b71/bin/replicate-mongodb.sh#L47-L55) that was created via a separate mechanism to the usual govuk-env-sync mechanism (which drove all the other apps’ data replication)
> According to the [comment](https://trello.com/c/mOYRWDOc/239-modify-govuk-docker-so-that-it-uses-the-govuk-env-sync-backups-3#comment-63e1375a4e990ca580f36bb1), that bucket only contained content_store_production, draft_content_store_production, imminence_production, and licence_finder_production (no asset manager DB).
> As ‘Asset manager’ has been ported to DocumentDB, its backup is created via govuk-env-sync under [govuk-integration-database-backups/mongo-normal](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups?region=eu-west-1&prefix=mongo-normal/&showversions=false) (“govuk_assets_production.gz”).
> It sounds like we need an additional case [in replicate-mongodb.sh](https://github.com/alphagov/govuk-docker/blob/cf7083421c0a6839128878a264b78ca0be28def3/bin/replicate-mongodb.sh#L18-L40)

chrisroos:
Thanks @kornysietsma :+1:

chrisroos (approved):

PR #674, opened 2023-06-30
Enable production-like logging in development

lauraghiorghisor-tw:
We want to allow production-like logging behaviour in development. As part of this behaviour we are setting GOVUK-Request-Id header.

[Trello card](https://trello.com/c/niLCNflj/117-story-enable-production-like-logging-in-dev-environments)

### DIFF
--- a/nginx-proxy.conf
+++ b/nginx-proxy.conf
@@ -12,6 +12,13 @@ map $http_x_forwarded_host $proxy_x_forwarded_host {
 
 proxy_set_header X-Forwarded-Host $proxy_x_forwarded_host;
 
+map $http_govuk_request_id $govuk_request_id {
+  default $http_govuk_request_id;
+  ''      $request_id;
+}
+
+proxy_set_header GOVUK-Request-Id $govuk_request_id;
+
 # Required for external assets of frontend apps
 add_header 'Access-Control-Allow-Origin' '*';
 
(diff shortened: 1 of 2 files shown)


nginx-proxy.conf
```diff
@@ -12,6 +12,13 @@ map $http_x_forwarded_host $proxy_x_forwarded_host {
 
 proxy_set_header X-Forwarded-Host $proxy_x_forwarded_host;
 
+map $http_govuk_request_id $govuk_request_id {
+  default $http_govuk_request_id;
+  ''      "$pid-$msec-$remote_addr-$request_length";
```
kevindew, line 17:
```suggestion
  ''      $request_id;
```

Since govuk_request_id is not something we expect to be parsed, we may as well just use nginx's built in type to keep things simple. This can help signify the format is not special and that this is not code we need to keep in sync with the production nginx perl code.

kevindew (commented):
Change looks good, just got a suggestion to make it simpler. 

What I'm not so clear on is how we expect people to make use of it, that could be useful information in the commit. I'm expecting that by default this won't be logged by nginx or Rails (though will be by Sidekiq) so I wasn't sure if we planned further changes so that this would be available even in developer logs, or if this is meant to be an advanced feature that you can only make use of for Rails apps if you turn on the env var to get the JSON logging?

nginx-proxy.conf
lauraghiorghisor-tw: We have now enabled the env var in the `helm-charts` [#1190](https://github.com/alphagov/govuk-helm-charts/pull/1190) so the logs are available in dev environment, in json format.

kevindew: Ah so you'd not expect people to use this without using the production JSON logs? This is information that shouldn't be in dev logs?

It'd be good to include this context in the commit message. Do we want to add a "how-to" for using it: https://github.com/alphagov/govuk-docker/blob/main/docs/how-tos.md ?

kevindew (approved):
Thanks 👍

PR #677, opened 2023-07-13
Add troubleshooting step for app not making

deborahchua:


AgaDufrat:
Merging as Deb is on annual leave

PR #679, opened 2023-07-20
Remove licence-finder as it is now retired

KludgeKML:
(replaced by find-a-licence inside finder-frontend)

https://trello.com/c/1VXQW5I1/1620-retire-original-licence-finder-application


leenagupte (approved):
🎉

PR #685, opened 2023-09-05
Fix static URL for assets

JessyTW:
In local dev assets in Whitehall are currently served using a static domain
e.g.http://static.dev.gov.uk/government/uploads/system/uploads/image_data/file/15/Screenshot_2023-04-30_at_22.15.58.png. 
It should be served from asset manager instead.
i.e. http://asset-manager.dev.gov.uk. 
This will enable correct rendering correct of assets.


ChrisBAshton (approved):
Assuming you've tested this locally 👍 the code looks fine to me.

PR #688, opened 2023-09-14
Configure RabbitMQ for search APIs

csutter:
- Change `search-api` to use `RABBITMQ_URL` instead of deprecated individual environment variables
- Add `RABBITMQ_URL` and RabbitMQ dependency to `search-api-v2`


JonathanHallam (approved):
👍

PR #695, opened 2023-10-20
Set up Google environment for `search-api-v2`

csutter:
This allows an engineer working on `search-api-v2` using GOV.UK Docker to authenticate on their host machine using `gcloud` and set an environment variable for their datastore in the dev GCP environment.

- Add mounting of `~/.config/gcloud` into the `search-api-v2` containers
- Add mirroring of local `DISCOVERY_ENGINE_DATASTORE` environment variable into containers
- Add checks to the Makefile to ensure user setup is correct
- Except `prerequisites` Make tasks from project name validation

### DIFF
--- a/projects/search-api-v2/Makefile
+++ b/projects/search-api-v2/Makefile
@@ -1,2 +1,22 @@
-search-api-v2: bundle-search-api-v2
+search-api-v2: prerequisites-search-api-v2 bundle-search-api-v2
 	$(GOVUK_DOCKER) run $@-lite bundle exec rake document_sync_worker:create_queue
+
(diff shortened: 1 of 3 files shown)


robinjam:
Does this break anything if the user _doesn't_ have the gcloud CLI configured on their local machine?

(LGTM if not)

csutter:
It should only fail if the user doesn't have the gcloud CLI, **and** they specifically try to run `make search-api-v2`.

If they are interacting with any other project through GOV.UK Docker it shouldn't affect them 🙂

robinjam (approved):

PR #702, opened 2023-12-08
Add postgres container with email-alert-api role

KludgeKML:
- allows container to run with role that matches the hygiene script
- remove obsolete comment

Not sure if there's a smoother way to do this.


sengi (approved):
Yeah I think it'd be easier to just create the user — the script only cares that the user exists.

sengi:
Thanks for fixing this — sorry for the breakage; keep forgetting govuk-docker is still a thing 🙈

PR #714, opened 2024-01-22
Change Support app to work with dart-sass

floehopper:
When [I moved the Support app from `libsass` to `dart-sass`][1] following [these instructions][2], I created a new `bin/dev` file which runs two processes in `procfile.dev`, one of which runs the rails server on port `3070` instead of port `3000`. This updates the `docker-compose.yml` file accordingly.

[1]: https://github.com/alphagov/support/pull/1253
[2]: https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html


chrisroos (approved):
Looks good to me, @floehopper :+1:

PR #715, opened 2024-01-25
Update s3 bucket and file matching pattern for mongo replication

AgaDufrat:
We no longer have mongo-normal direcorty in the
[govuk-integration-database-backups](https://s3.console.aws.amazon.com/s3/buckets/govuk-integration-database-backups?region=eu-west-1&bucketType=general&prefix=shared-documentdb/&showversions=false) s3 bucket. Those are now in shared-documentdb.

The file matching also stopped returning file name. It was not updated to match the end of the expression with `\>`

### DIFF
--- a/bin/replicate-mongodb.sh
+++ b/bin/replicate-mongodb.sh
@@ -30,11 +30,11 @@ case "$app" in
     database="${app//-/_}_production"
     ;;
   "asset-manager")
-    hostname=mongo-normal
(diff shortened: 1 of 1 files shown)


brucebolt:
Thanks for raising this PR @AgaDufrat!  It looks like https://github.com/alphagov/govuk-docker/pull/732 has been raised since which has been merged and solved the issue.

PR #732, opened 2024-03-10
Updates mongo replication script for new S3 locations and format

andy-collon:
https://trello.com/c/v2WB7RzB/187-504-error-when-searching-and-filtering-in-mainstream-publisher-5631736

These updates are mostly what I needed to get the script to replicate publisher locally, I also removed content store as it's no longer on mongo and licensify as there doesn't appear to be a backup in S3 right now. These are the missing lines covering licensify, updated for what I think the instance should be;
```
    "licensify")
    instance=licensify-documentdb
    database=licensify
    ;;
```


anatron (approved):
Looks good!

PR #733, opened 2024-03-11
Adds sidekiq worker service to govuk-chat

AlexAvlonitis:
## Description

Adds sidekiq worker service to govuk-chat
- Adds redis and postgres as dependencies for starters

## Trello card

https://trello.com/c/B3ykgWFN/1191-background-worker-to-generate-an-answer-from-python-chat-api

### DIFF
--- a/projects/govuk-chat/docker-compose.yml
+++ b/projects/govuk-chat/docker-compose.yml
@@ -30,12 +30,24 @@ services:
   govuk-chat-app:
     <<: *govuk-chat
     depends_on:
+      - govuk-chat-worker
       - nginx-proxy
       - postgres-13
     environment:
       DATABASE_URL: "postgresql://postgres@postgres-13/govuk-chat"
+      REDIS_URL: redis://redis
       VIRTUAL_HOST: govuk-chat.dev.gov.uk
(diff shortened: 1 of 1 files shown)


projects/govuk-chat/docker-compose.yml
```diff
@@ -30,12 +30,22 @@ services:
...
       VIRTUAL_HOST: govuk-chat.dev.gov.uk
       BINDING: 0.0.0.0
     expose:
       - "3000"
     command: bin/dev
+
+  govuk-chat-worker:
+    <<: *govuk-chat
+    depends_on:
+      - redis
```
davidgisbey, line 48: do we need to add postgres as a dependency and pass it a database url like [here](https://github.com/alphagov/govuk-docker/blob/34dfd9d94c2d1d3e31ab6b3ad10b81deee7cce80/projects/content-publisher/docker-compose.yml#L55-L63)?

AlexAvlonitis: Good point, we probably need to access postgres from the workers. I'll add it

davidgisbey (approved):
lgtm!

PR #740, opened 2024-04-02
Integrate a message queue with govuk-chat

kevindew:
Trello: https://trello.com/c/q6AqNW2t/1265-spike-into-what-is-needed-to-consume-publishing-api-queue-for-search-ingestion

GOV.UK Chat will be listening to the Publishing API message queue for published messages.

This change depends on https://github.com/alphagov/govuk-chat/pull/69

I've added a rabbitmq dependency for lite so that rabbit is available for set-up commands.


AlexAvlonitis (approved):
looks great!

PR #743, opened 2024-04-05
Updates router Dockerfile base image to golang 1.22

andy-collon:
We had to make a few changes to get router to make, as part of https://trello.com/c/6YIezKsO/218-spike-how-to-run-the-entire-publishing-suite-locally

- Update the version of golang in the Dockerfile to match the router repo (a dependency was failing initially as it itself had a dependency on framework code in 1.21, then with the 1.22.0 update the version in the Dockerfile, 1.20.2, didn't recognise the new version number format)
- Force docker-compose to rebuild, picking up the version change in the Dockerfile - otherwise it won't

The rebuild itself is barely an inconvenience when the images are cached so shouldn't impact local dev times at all.


AgaDufrat (approved):
LGTM

PR #745, opened 2024-04-24
Update `search-admin` config to reflect app changes

csutter:
- Remove `search-api` dependency
- Remove environment variable for `database_cleaner` configuration (this gem has been removed as we're not using Cucumber anymore and RSpec system tests run in transactions so don't need it)
- Change command in `app` stack to run conventional `bin/dev` that has been added to the application
- Add explicit port to `app` stack now that the application is run through Foreman
- Add `TEST_DATABASE_URL` to app stack to allow running tests on `app` stack too for convenience


AgaDufrat (approved):
LGTM

PR #748, opened 2024-05-01
Use postgres-16 for govuk-chat

AlexAvlonitis:
- Adds postgres-16 in the list of databases.
- Govuk-chat will be using postgres-16 in all the environments (integration, staging, prod). Updated the dev environment to reflect that.

Trello Card: https://trello.com/c/ewV5ilJ6/1229-initial-terraform-for-integration-environment


kevindew (approved):
Nice one, just worked fine for me locally

PR #750, opened 2024-05-03
Add OpenSearch as a govuk-chat-worker dependency

kevindew:
This process now requires access to the search database.


peteglondon (approved):
👍

PR #763, opened 2024-07-12
Replicate data from staging, not integration

ChrisBAshton:
Since the env sync was ported over to govuk-helm-charts, not all apps have backups saved to s3 on integration (e.g. note the lack of `backup` operation [here](https://github.com/alphagov/govuk-helm-charts/blob/cae5e10baf0f67423bbebf66454da737da787d37/charts/db-backup/values.yaml#L767-L773)). This means that the data replication scripts fail, as they're looking for a bucket that does not exist.

In contrast, all data is backed up to s3 on staging, so it should be safe to pull from there by default. The only downside is that integration access is given to devs by default, but that access to staging and production required Production Admin access. This is something that can hopefully be ironed out in the move to using govuk-helm-charts for local development, over the coming months. For now it seems better to fix the script for prod-admin users than to have it broken for everyone.

NB I've also removed the unnecessary duplication of replication examples where one used a readonly role and one used a poweruser role. We should use principle of least privilege, so only show the readonly example.

### DIFF
--- a/bin/replicate-mongodb.sh
+++ b/bin/replicate-mongodb.sh
@@ -33,7 +33,7 @@ case "$app" in
     ;;
 esac
 
-folder="govuk-integration-database-backups/$instance"
+folder="govuk-staging-database-backups/$instance"
 
 archive_dir="${replication_dir}/mongodb"
 archive_file="${database}.gz"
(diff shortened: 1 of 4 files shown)


kevindew:
This feels like something that'll make it harder for any new starters and also creates an increased risk of sensitive data being on the dev machines given staging dbs aren't scrubbed.

Can we not just fix the integration backups? 

Failing that could we have an env var of the environment that defaulted to integration, and if necessary, someone could change that to pull from staging/prod for a db backup (Is there a reason we'd actually want staging over prod?)

ChrisBAshton:
Agreed. Superseded by https://github.com/alphagov/govuk-helm-charts/pull/2200

PR #766, opened 2024-07-30
Fix broken Elasticsearch 6 container

csutter:
The underlying version of Ubuntu for our custom ES6 container has changed, and it now includes a non-root user by default, with the same ID as the one which we were previously creating. This was causing the image build to fail.

- Remove custom non-root user setup and replace with default non-root user


kevindew (approved):
Ooh nice.

PR #770, opened 2024-08-09
Use dartsass for publisher css compilation

mtaylorgds:
### What
Change docker compose for Mainstream Publisher to match compilation of CSS using dart-sass. See https://github.com/alphagov/publisher/pull/2278.

### Why
We're switching these apps from libsass to dart-sass, and this change is needed to support that.

### DIFF
--- a/projects/publisher/docker-compose.yml
+++ b/projects/publisher/docker-compose.yml
@@ -37,6 +37,7 @@ services:
       - publishing-api-app
       - link-checker-api-app
       - publisher-worker
+      - publisher-css
     environment:
       MONGODB_URI: "mongodb://mongo-3.6/publisher"
       REDIS_URL: redis://redis
@@ -44,7 +45,11 @@ services:
       BINDING: 0.0.0.0
     expose:
       - "3000"
-    command: bin/rails s --restart
+    command: bin/dev web
+
+  publisher-css:
+    <<: *publisher
+    command: bin/dev css
 
   publisher-worker:
     <<: *publisher
@@ -56,4 +61,4 @@ services:
     environment:
       MONGODB_URI: "mongodb://mongo-3.6/publisher"
       REDIS_URL: redis://redis
-    command: bundle exec sidekiq -C ./config/sidekiq.yml
+    command: bin/dev worker


projects/publisher/docker-compose.yml
```diff
@@ -44,7 +44,7 @@ services:
       BINDING: 0.0.0.0
     expose:
       - "3000"
-    command: bin/rails s --restart
+    command: bin/dev
```
kevindew, line 47: Something that is a bit weird about this approach (and appreciate horse has bolted for a bunch of our repos) is that bin/dev is going to be managing multiple processes which isn't typical behaviour for docker. It raises quite a wtf for apps like this where there is also a worker process - why have that managed differently in docker to css?

I'd suggest you take the approach we've done in govuk-chat and have each individual process be it's own docker process:

https://github.com/alphagov/govuk-docker/blob/9604d0bcd2158b0fb97be4d12240e72d0b9aeb16/projects/govuk-chat/docker-compose.yml#L50-L54

davidtrussler: I have tried this approach locally and found that it does not run Sass in watch mode (as described [here](https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html#watch-mode)). I don't know if that is inherent in the approach or if it requires a change in this file to run.

kevindew: it will be running the exact same process

davidtrussler: What I am seeing in the console as I start publisher is shown below, comparing the original changes with what is suggested here, which suggests the command `Execute dartsass:watch` is not being called.
|Screenshot 2024-08-12 at 12 36 00|Screenshot 2024-08-12 at 12 30 15|
|-|-|
|Screenshot 2024-08-12 at 12 21 34|Screenshot 2024-08-12 at 12 21 50|

kevindew: For the latter you need to execute publisher-css as a separate process and/or flag it as a dependency of the process - as per how the worker is configured, https://github.com/alphagov/govuk-docker/blob/9fbfaeaf2ff991e398f055a5721d57f6090262ce/projects/publisher/docker-compose.yml#L39

The former example you've got there is unconventional for docker as it's managing 2 processes

davidtrussler: OK, that's making sense (and also working as expected 😄) so I have updated accordingly.

kevindew: Great 😄 you'll just need to edit:

https://github.com/alphagov/govuk-docker/blob/1de6d855646e9a124a7b4b306b3e58e657d40033/projects/publisher/docker-compose.yml#L33-L39

to include publisher-css, which will mean that css is automatically started when web is started.

davidtrussler: Oh yeah, have done that as well. 👍

kevindew (approved):

PR #773, opened 2024-08-16
Added note about preventing unexpected DB schema diffs

ryanb-gds:
This has become a bit of an annoying issue for the Whitehall team over time, and has sometimes led to changes to the schema.rb file not being committed when they should have been.

Trello: https://trello.com/c/Ai59Yd1F

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -205,3 +205,19 @@ This may resolve by removing orphan containers which were created in a previous 
 ```
 docker-compose down -v --rmi all --remove-orphans
 ```
+
+
+## Rails database schema changed unexpectedly when running `rake db:migrate`
+
+It is common to find that running `rake db:migrate` results in unexpected changes to the `schema.rb` file. This is caused by slight differences between databases running in GOV.UK Docker and the production databases on AWS.
+
+Unfortunately, to resolve the issue, the only solution is to drop the database in GOV.UK docker and create a new one.
+
+Once you have taken a backup, you can remove and re-create your local database using the following commands:
+
+```bash
+rake db:drop
+rake db:create
+rake db:schema:load
+rake db:schema:migrate // optional, but if you are adding a new migration this will now update the schema without creating a horrible diff
+```


docs/troubleshooting.md
```diff
@@ -205,3 +205,19 @@ This may resolve by removing orphan containers which were created in a previous
 ```
 docker-compose down -v --rmi all --remove-orphans
 ```
+
+
+## Rails database schema changed unexpectedly when running `rake db:migrate`
+
+It is common to find that running `rake db:migrate` results in unexpected changes to the `schema.rb` file. This is caused by slight differences between databases running in GOV.UK Docker and the production databases on AWS.
+
+Unfortunately, to resolve the issue, the only solution is to drop the database in GOV.UK docker and create a new one. This is annoying if you've built up a useful corpus of development data, so you might want to take a backup containing just the data. For MySQL you can do this using the [--no-create-info](https://dev.mysql.com/doc/refman/8.4/en/mysqldump-definition-data-dumps.html) option with MySQL dump, for Postgres you can use the `--data-only` option for PostgreSQL dump.
```
ChrisBAshton, line 214: Can you include the exact command you used to create the data backup? As I'd still have to go and Google how to do a data dump and then figure out where to pass this parameter.

ChrisBAshton (approved):
LGTM, one small suggestion below

docs/troubleshooting.md
ryanb-gds: I could, but i haven't actually ever done this, so it's entirely theoretical. I suppose I should probably google it on behalf of everyone else instead of being lazy 😅 

The other point I hadn't really thought about is that for this to be that useful you would probably need to back up all the databases (publishing app, publishing API, content store, router API, asset manager) so it probably isn't going to be worth it. Though maybe you could just republish all the content from Whitehall 🤔

ChrisBAshton: In that case I'd probably delete everything after the first sentence - someone else can document it at the point of use if they end up taking a backup 👍 

(NB I'd think it's easier to drop the DB and then reimport it from the local tar download in govuk-docker, assuming that's not been deleted by the user since they last downloaded a backup from AWS)

ChrisBAshton (approved):

PR #775, opened 2024-08-21
Add a how to guide for how to publish content from Whitehall using GOV.UK Docker

ryanb-gds:
This is very similar to the guide for testing finders on GOV.UK Docker, but with some of the steps removed.

Trello: https://trello.com/c/Ai59Yd1F

### DIFF
--- a/docs/how-tos/publishing-content-on-whitehall.md
+++ b/docs/how-tos/publishing-content-on-whitehall.md
@@ -1,0 +1,33 @@
+# Publishing Content from Whitehall for Developing Government Frontend
+
+This guide explains how to use GOV.UK Docker to develop features for government frontend, and how to publish content to test your
+work with locally. The instructions below assume that you have "made" all of the relevant projects before starting.
+
+1. Start the Whitehall app: `govuk-docker up whitehall-app`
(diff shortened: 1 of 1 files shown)


docs/how-tos/publishing-content-on-whitehall.md
```diff
@@ -0,0 +1,19 @@
+# Publishing Content from Whitehall for Developing Government Frontend
+
+This guide explains how to use GOV.UK Docker to develop features for government frontend, and how to publish content to test your
+work with locally. The instructions below assume that you have "made" all of the relevant projects before starting.
+
+1. Start the Whitehall app: `govuk-docker up whitehall-app`
+2. Start the government-frontend app: `govuk-docker up government-frontend-app`
+3. Create a router backend for government frontend: `curl http://router-api.dev.gov.uk/backends/government-frontend -X PUT \
```
ChrisBAshton, line 8: Do the backticks work when spread over multiple lines? I'd be tempted to pull this out into triple-backtick blocks instead (indented so that you don't 'reset' the list counter).

ryanb-gds: They do work, but the triple backtick style is much nicer for multiline commands. Good shout!

PR #780, opened 2024-09-02
Only use loopback interface for nginx

unoduetre:
## What

Currently docker allows any host on the network to connect to the nginx proxy and consequently allows anyone on the network to access the apps while in development. This is because nginx is currently configured to bind on all interfaces.

This PR changes that so it only binds to 127.0.0.1 (the loopback interface).

## Why

https://gds.slack.com/archives/CAB4Q3QBW/p1725265458430239


kevindew (commented):
Thanks for this. It does feel like we should be able to set this with a global option in the network bit, I'd want us to rule that out before committing this.

I also get an error for the nginx part - this could be my local machine of course:

```
Error response from daemon: Ports are not available: exposing port TCP 192.168.1.208:80 -> 0.0.0.0:0: failed to connect to /var/run/com.docker.vmnetd.sock: is vmnetd running?: dial unix /var/run/com.docker.vmnetd.sock: connect: no such file or directory
```

As mentioned on Slack, we're not under pressure to resolve this instantly and can wait for further comms from @jameskirk36 before proceeding.

unoduetre:
> Thanks for this. It does feel like we should be able to set this with a global option in the network bit, I'd want us to rule that out before committing this.
> 
> I also get an error for the nginx part - this could be my local machine of course:
> 
> ```
> Error response from daemon: Ports are not available: exposing port TCP 192.168.1.208:80 -> 0.0.0.0:0: failed to connect to /var/run/com.docker.vmnetd.sock: is vmnetd running?: dial unix /var/run/com.docker.vmnetd.sock: connect: no such file or directory
> ```
> 
> As mentioned on Slack, we're not under pressure to resolve this instantly and can wait for further comms from @jameskirk36 before proceeding.

Thanks. If someone could provide a more general solution that would automatically apply to all ports, I also think that'd be better and this PR could be closed.

I checked and it works for me e.g. for frontend-app-live: `govuk-docker up frontend-app-live` starts the app and I can access it through `frontend.dev.gov.uk`. Maybe restarting docker and removing the container nginx-proxy could help? What app did you check?

kevindew:
That was government-frontend, after first checking on govuk-chat. Yeah I tried restarting docker. 

I imagine I just need to look into the error more, I imagine it might be something on my host machine perhaps if it's localised to just me.

hannako:
I got the same error as @kevindew  
```no container to killError response from daemon: Ports are not available: exposing port TCP 127.0.0.1:80 -> 0.0.0.0:0: failed to connect to /var/run/com.docker.vmnetd.sock: is vmnetd running?: dial unix /var/run/com.docker.vmnetd.sock: connect: no such file or directory```

I stopped and killed all the containers (and images and volumes!) before running `make government-frontend` and then `govuk-docker-up app-live` from the government frontend repo.

AgaDufrat:
@unoduetre are you still working on this? Can we revert it to draft or close?

PR #781, opened 2024-09-03
Add an app-live to docker compose for smart answers

syed-ali-tw:
Some of the dev urls that are required to run smart answers locally don't work and we would need to configure live urls hence with this PR we have added an app-live to docker compose to allow us to run locally using some live urls.


mtaylorgds (approved):
LGTM

PR #783, opened 2024-09-24
Retire special route publisher

KludgeKML:
Special-route-publisher tasks are taken over by publishing-api as of: https://github.com/alphagov/publishing-api/pull/2881, so update documentation and remove special-route-publisher as an app in govuk-docker.

https://trello.com/c/9rOxWNnD/396-retire-special-route-publisher


ChrisBAshton (approved):
Thanks for this ⭐

PR #785, opened 2024-10-10
Add option to skip branch checks on `make`

csutter:
Sometimes I just want to quickly rebuild a project (for example if things have gotten into a bad state somehow) without the build scripts obsessively checking for updates to dependent repositories. This is especially useful when I have several dependent repositories intentionally checked out to a different branch and don't want to have to confirm that each time, or when I am on a train and the internet connection is too spotty to reliably update local repositories.

This adds an environment variable flag `SKIP_BRANCH_CHECKS` that when set will skip the branch checks on `make` invocations.

### DIFF
--- a/bin/branch_checks.sh
+++ b/bin/branch_checks.sh
@@ -7,6 +7,11 @@ if [[ "$#" == "0" ]]; then
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Nice - I've wanted this feature too!

PR #789, opened 2024-10-30
Run Publishing API alongside Collections

brucebolt:
We are adding GraphQL queries into Collections, which are made directly to Publishing API.

Therefore we need to run Publishing API at the same time as Collections when developing locally.

[Trello card](https://trello.com/c/jDF1nSQ0)

### DIFF
--- a/projects/collections/Makefile
+++ b/projects/collections/Makefile
@@ -1,2 +1,2 @@
-collections: bundle-collections content-store router static
+collections: bundle-collections content-store publishing-api router static
 	$(GOVUK_DOCKER) run $@-lite yarn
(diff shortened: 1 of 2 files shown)


yndajas (approved):
Is it worth clarifying in the commit message that this is for a GraphQL _proof of concept/pilot_, just in case anyone reads this out of context and gets overly excited/concerned? I suppose this only really affects local (and CI test) environments though

brucebolt:
That's a good idea.  I've added that into the commit message here.

I'll also make the same clear in the GDS API Adapters PR, that adds the GraphQL method.

PR #794, opened 2024-11-14
Upgrade version of PostgreSQL used by Publishing API

brucebolt:
We were using an outdated image, which contained a version of `pg_restore` that was not compatible with the dumps of production data.

Therefore upgrading the version of PostgreSQL image we are using to allow replication of real data.


brucebolt:
I'm not sure it's the same version of Postgres, but it's definitely a version of `pg_restore` that is compatible with the version of `pg_dump` we're using.

PR #799, opened 2024-11-19
Add env values for local run to allow moving away from Rails.secrets

syed-ali-tw:
Rails.application.secrets is deprecated in favor of Rails.application.credentials and and will be removed in Rails 7.2.

This generates alot of warning noise with test runs.

We dont necessarily need Rails.application.credentials and we can just replace the values to be read from environment variables.

### DIFF
--- a/projects/publisher/docker-compose.yml
+++ b/projects/publisher/docker-compose.yml
@@ -27,6 +27,8 @@ services:
       MONGODB_URI: "mongodb://mongo-3.6/publisher"
       TEST_MONGODB_URI: "mongodb://mongo-3.6/publisher-test"
       REDIS_URL: redis://publisher-redis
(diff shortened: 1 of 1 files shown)


MuriloDalRi (approved):

AgaDufrat:
@syed-ali-tw Do you still want to merge this? Persoanly I think this PR should be closed and the relevant changes made in the Publisher repo.

You don’t need to set the SECRET_KEY_BASE. In development and test the value is randomly generated and stored in a temporary file in `tmp/local_secret.txt`. 
The `ENV SECRET_KEY_BASE_DUMMY=1` is set in
https://github.com/alphagov/govuk-ruby-images/blob/main/builder.Dockerfile#L16

AgaDufrat:
@syed-ali-tw Can we close this?

syed-ali-tw:
> @syed-ali-tw Can we close this?

Hey @AgaDufrat , not sure why I was not notified about your message, yes, I will close this now, this has been addressed in Publisher

PR #801, opened 2024-11-21
Add search-api V1 to frontend

leenagupte:
# What's changed?

Add search-api V1 config to frontend

# Why?

search-api V1 is being called from the take part pages and the new landing pages, so should be added here to aid local development.

search-api V1 config has already been added to the startup script and the in app.json (for Heroku) files frontend.


projects/frontend/docker-compose.yml
```diff
@@ -47,6 +48,7 @@ services:
       GOVUK_PROXY_STATIC_ENABLED: "true"
       PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
       PLEK_SERVICE_STATIC_URI: https://assets.publishing.service.gov.uk
+      PLEK_SERVICE_SEARCH_API_URI: https://www.gov.uk/api
```
sihugh, line 51: If you're setting this you probably don't need to add the `Makefile` or app dependency bits?

leenagupte: I thought I'd add it for consistency, in case someone does try to run all the things locally. 😬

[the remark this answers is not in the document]
leenagupte: Oops! Good catch!

brucebolt: I think the dependency will be required, if you run the app without specifying live or integration (i.e. to use the local apps only).

PR #803, opened 2024-11-27
Search API v2: Add `GOOGLE_CLOUD_PROJECT_ID`

csutter:
This environment variable is now available on the deployed environments, adding it here as well.

see https://github.com/alphagov/govuk-helm-charts/pull/2802


nimalank7 (approved):
LGTM!

PR #805, opened 2024-11-28
Document the "get back to clean slate" option

ChrisBAshton:
We've [discussed adding this before](https://github.com/alphagov/govuk-docker/pull/434#pullrequestreview-533335033), and shied away from it in favour of more focussed troubleshooting. But sometimes folks just want a quick and easy task to reset things - so let's add this in.


KludgeKML (approved):
LGTM

PR #816, opened 2025-02-10
Add `signon-app` to Whitehall/Mainstream

pezholio:
Whitehall (and Mainstream shortly) will soon need access to Signon to fetch user information about updated content blocks (see https://github.com/alphagov/publisher/pull/2513 and https://github.com/alphagov/whitehall/pull/9754)


mtaylorgds (approved):
LGTM

PR #820, opened 2025-02-19
Search Admin: Add Publishing API token

csutter:
This is now required by the app to run, but can be an arbitrary value in development.

### DIFF
--- a/projects/search-admin/docker-compose.yml
+++ b/projects/search-admin/docker-compose.yml
@@ -30,6 +30,7 @@ services:
       TEST_DATABASE_URL: "mysql2://root:root@mysql-8/search_admin_test"
       REDIS_URL: redis://search-admin-redis
       DISCOVERY_ENGINE_DEFAULT_COLLECTION_NAME: "projects/780375417592/locations/global/collections/default_collection"
+      PUBLISHING_API_BEARER_TOKEN: example
 
   search-admin-app: &search-admin-app
     <<: *search-admin-base
@@ -46,6 +47,7 @@ services:
       BINDING: 0.0.0.0
       PORT: 3000
       DISCOVERY_ENGINE_DEFAULT_COLLECTION_NAME: "projects/780375417592/locations/global/collections/default_collection"
+      PUBLISHING_API_BEARER_TOKEN: example
     expose:
       - "3000"
     command: bin/dev


projects/search-admin/docker-compose.yml
```diff
@@ -46,6 +47,7 @@ services:
       BINDING: 0.0.0.0
       PORT: 3000
       DISCOVERY_ENGINE_DEFAULT_COLLECTION_NAME: "projects/780375417592/locations/global/collections/default_collection"
+      PUBLISHING_API_BEARER_TOKEN: example
```
leenagupte, line 50: Similar question here. Is this needed when the app already "depends on" `publishing-api-app`?

I'm comparing it to the set-up for the other publishing apps. e.g. https://github.com/alphagov/govuk-docker/blob/main/projects/collections-publisher/docker-compose.yml#L40

projects/search-admin/docker-compose.yml
```diff
@@ -30,6 +30,7 @@ services:
       TEST_DATABASE_URL: "mysql2://root:root@mysql-8/search_admin_test"
       REDIS_URL: redis://search-admin-redis
       DISCOVERY_ENGINE_DEFAULT_COLLECTION_NAME: "projects/780375417592/locations/global/collections/default_collection"
+      PUBLISHING_API_BEARER_TOKEN: example
```
leenagupte, line 33: I think the `lite` stack is only used to run the tests, so I'm not sure if this is required.

leenagupte (commented):
I have a couple of inline questions.

csutter:
> I have a couple of inline questions.

Both are fair questions – the underlying reason is the same: [Search Admin `fetch`es the value of critical configuration environment variables on boot](https://github.com/alphagov/search-admin/blob/main/config/application.rb#L49-L53), so the app would refuse to start if the environment variable is missing.

I'm a big fan of this pattern compared to just doing `ENV["blah"]` deep down in the codebase, because it declares all "critical" configuration without which the app can't run in one place, and it's particularly useful in our k8s environment:

If someone were to accidentally change the configuration of the app in a breaking way, for example by removing or misspelling this env variable in the Helm chart, a new pod would refuse to come up and the old pod would continue running with the previous configuration (instead of a new pod happily starting and then breaking at a random point later down the line when, in this instance, someone would try to publish a recommended link from Search Admin).

The tradeoff here is that you then need to make sure this environment variable is set in circumstances where it otherwise wouldn't be needed: for `lite` stacks, local development, and for [Github Actions config](https://github.com/alphagov/search-admin/blob/main/.github/workflows/rspec.yml#L19).

I think this is okay though, because it keeps this behaviour consistent across environments - having a way to avoid this locally would introduce different code paths with all the fun and games that brings.

leenagupte (approved):
Thanks for explaining. That all makes sense. 💯

PR #823, opened 2025-02-24
Add postgres url along with mongo to support mongo and postgres in parallel

syed-ali-tw:
We want to be able to run mongo and postgres in parallel for Mainstream Publisher app while we move Mainstream to postgress.

### DIFF
--- a/projects/publisher/docker-compose.yml
+++ b/projects/publisher/docker-compose.yml
@@ -22,16 +22,20 @@ services:
     shm_size: 512mb
     depends_on:
       - publisher-redis
+      - postgres-16
       - mongo-3.6
     environment:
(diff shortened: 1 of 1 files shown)


projects/publisher/docker-compose.yml
```diff
@@ -22,16 +22,20 @@ services:
     shm_size: 512mb
     depends_on:
       - publisher-redis
+      - postgres-13
```
baisa, line 25: We should probably be using new version (17) of Postgres?

syed-ali-tw: ah yes, good point

Tetrino (approved):

PR #831, opened 2025-03-12
Update developer docs config

pezholio:
This adds a `yarn` step to Developer Docs now we have JS dependencies (see https://github.com/alphagov/govuk-developer-docs/pull/5019), as well as adding a `NO_CONTRACTS` env var to skip some of Middleman's magic typechecking, which slows the build down.


ChrisBAshton (approved):
LGTM

PR #852, opened 2025-05-23
Use Debian Bookworm

MuriloDalRi:
**As a** Developer
**I want** govuk-docker to be on a supported version of Debian **so that** we continue to get updated packages such as Chromium and stop having problems with local development

https://github.com/alphagov/govuk-infrastructure/issues/2036

### DIFF
--- a/Dockerfile.govuk-base
+++ b/Dockerfile.govuk-base
@@ -3,12 +3,18 @@
 # project and make any changes to it there.
 
 # Install packages for building ruby
-# TODO: Change from bullseye to stable tag once we've resolved problems with Ruby and SSL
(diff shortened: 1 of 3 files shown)


AgaDufrat:
I tried to test it locally but I'm getting an error when bundle runs on this branch (`main` branch is fine for me):
```
Could not load OpenSSL.
You must recompile Ruby with OpenSSL support.
```

MuriloDalRi:
> I tried to test it locally but I'm getting an error when bundle runs on this branch (`main` branch is fine for me):
> 
> ```
> Could not load OpenSSL.
> You must recompile Ruby with OpenSSL support.
> ```

What command were you trying to run? Did you rebuild your containers from scratch?

PR #879, opened 2025-08-20
On GitHub Actions dependency updates: convert tags to commit hashes

AP-Hunt:
We want all of our GitHub Actions actions dependencies to be pinned to a specific commit hash, but Dependabot will only pin to a tag. To work around this we introduce this workflow, which runs whenever Dependabot raises a relevant PR, and uses a tool called pinact to convert the tags to the commit id to which they point at that moment in time.


AP-Hunt:
Closing as unnecessary. Dependabot will bump pinned hashes when it sees them, instead of tags, and we already have PRs in all of the repositories to change to hashes.

PR #882, opened 2025-08-28
Add `content-block-manager` app

pezholio:
This adds the in-progress [`content-block-manager`](https://github.com/alphagov/content-block-manager) app to govuk-docker. The app is currently running as an engine in Whitehall, but we’re in the process of moving it to it’s own app.


ChrisBAshton (approved):
Thanks for this - exciting to see!

PR #883, opened 2025-08-29
Add search-api-v2-quality stack for running evaluations rake tasks locally.

emmalowe:
The goal of the PR is to allow us to easily run evaluations related rake tasks locally e.g.

`govuk-docker run search-api-v2-quality bundle exec rake quality:setup_sample_query_sets`


projects/search-api-v2/docker-compose.yml
```diff
@@ -66,5 +66,15 @@ services:
       DISCOVERY_ENGINE_DEFAULT_LOCATION_NAME: "projects/780375417592/locations/global"
     command: bin/rake document_sync_worker:run
 
+  search-api-v2-quality:
```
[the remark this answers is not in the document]
emmalowe: I don't think it will work for the document_sync_worker rake tasks though (that seems to need rabbitmq, and is probably why we have the specific stack for that), so it's not completely generic. If we wanted to use "search-api-v2-task-runner" as the name, I should probably document how to run the document_sync_worker rake tasks specifically in the README and say task-runner is for the rest. Thoughts?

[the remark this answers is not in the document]
leenagupte: I think you'd need to use the new stack to run the user-events tasks as they call DiscoveryEngine and most likely also need the product_id etc.

Documenting the `document_sync_worker` as well would be great.

projects/search-api-v2/docker-compose.yml
```diff
@@ -66,5 +66,15 @@ services:
       DISCOVERY_ENGINE_DEFAULT_LOCATION_NAME: "projects/780375417592/locations/global"
     command: bin/rake document_sync_worker:run
 
+  search-api-v2-task-runner:
+    <<: *search-api-v2
+    depends_on:
+      - search-api-v2-redis
+    environment:
+      REDIS_URL: redis://search-api-v2-redis
```
hannako, line 74: What does it need Redis for?

emmalowe: Oh good point... probably nothing. The application was just failing because it's expecting a `REDIS_URL` env variable, but setting this to `none` and removing the dependency looks like it works.

If I run `govuk-docker run search-api-v2-task-runner bundle exec rake quality:setup_sample_query_sets` with that change, I get into the `Google::Cloud::AlreadyExistsError:`.

leenagupte: This is a good catch. It looks like only document sync worker needs access to Redis, and that already has it's own stack.

PR #889, opened 2025-09-12
[SCH-1491] Add new search-api-v2-beta-features app

leenagupte:
This app will only be used to run evaluation tasks and therefore doesn't need a web server.


AgaDufrat:
Converting to Draft for the moment until it's ready later in the week. 
Related to https://github.com/alphagov/govuk-infrastructure/pull/2798

PR #907, opened 2025-10-23
Move dnsmasq to port 533

craigjbass:
Which enables GOV.UK dnsmasq to run on devices that have ZScaler ZIA running on them (which exposes a DNS resolver on port 53).


samsimpson1 (approved):
LGTM, thanks for sorting this. I tested it on my CO mac and it seems to be working as intended.

PR #914, opened 2025-11-07
Reconfigure frontend to not use static

andysellick:
## What / why
Change `frontend` to not rely on `static` anymore.

- app was modified to not use static in October: https://github.com/alphagov/frontend/pull/4928
- updating the config here to reflect that


KludgeKML (approved):
LGTM!

PR #915, opened 2025-11-12
Reconfigure smart-answers to not use static

andysellick:
## What / why
Change `smart-answers` to not rely on `static` anymore.

- app was modified to not use static in August: https://github.com/alphagov/smart-answers/pull/7304
- updating the config here to reflect that


KludgeKML (approved):
LGTM

PR #957, opened 2026-01-30
Remove `content-block-manager-full`

pezholio:
This was always a convenience service to run all services needed to run the whole stack, but caused some issues when using RubyMine and Docker together, so let’s bin it off (I think it’s only me that uses it anyway!)

I’ll update the app itself to add a convenience script to run the full stack.


jasonBirchall:
closing as it's aged out

PR #986, opened 2026-04-27
Revert "Add `draft-frontend`" and replace functionality.

KludgeKML:
Reverts alphagov/govuk-docker#974, then reinstates the functionality (with a slight change to app name).

974 adds the ability to access the local draft stack through govuk-docker, but introduces two problems:
- it adds a number of dependencies to the "-lite" version of the service, which should be kept dependency free (in particular, this seems to make build times worse)
- it breaks debugging on app-live and app-integration by removing the exposed debugging port. This is to allow the live and draft containers to be run at the same time, but it's a bit of a drastic solution to the problem and essentially disabled a useful feature with no notice. Here we alter the exposed port so that the app-draft stack can still be debugged, but doesn't clash with the port used by app/live/integration. It's assumed here that people probably aren't running live and integration versions at the same time (but they might want local and local draft together)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -170,7 +170,6 @@ services:
           - draft-content-store.dev.gov.uk
           - draft-origin.dev.gov.uk
           - draft-router.dev.gov.uk
-          - draft-frontend.dev.gov.uk
           - email-alert-api.dev.gov.uk
           - email-alert-frontend.dev.gov.uk
           - fact-check-manager.dev.gov.uk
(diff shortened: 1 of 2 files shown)


KludgeKML:
@pezholio could you try out this branch when you have a moment? I believe it will do what you want (ie allow you to run versions of the container pointing to live _and_ draft live at the same time), but since I don't know your exact use case it would be useful for you to confirm.

deborahchua (approved):
LGTM!

PR #989, opened 2026-04-29
Remove router api

KludgeKML:
Router-api has been archived, and is no longer necessary in govuk-docker. Remove the app, and references to it in documentation, scripts, and other apps.


nimalank7 (approved):
LGTM!

PR #994, opened 2026-05-21
Add app-ports port forwarding support for Frontend

KludgeKML:
- This allows us to run local frontend against all the backing services in any environment, based on port forwarding, improving dev feedback loop for postcode lookup pages, licences, etc.

### DIFF
--- a/bin/port-forward
+++ b/bin/port-forward
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -eu
+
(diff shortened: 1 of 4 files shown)


KludgeKML:
Fixes https://github.com/alphagov/frontend/issues/2803

deborahchua (approved):
Haven't been able to try it out myself yet, but looks good to me!

PR #1002, opened 2026-06-16
Ignore cert errors when installing elastic

graycodes:
When running `make search-api` I was getting an error on the line where we fetch elastic. Possibly introduced as part of a recent upgrade to the elastic version, or just a change on the vendor side? Curl is failing on fetching the tarball with this error:
```
curl: (60) SSL certificate problem: unable to get local issuer certificate 
```

I assume this is fine to ignore for the purposes of installing elastic in docker.

### DIFF
--- a/services/elasticsearch-6/Dockerfile
+++ b/services/elasticsearch-6/Dockerfile
@@ -19,7 +19,7 @@ RUN apt update && apt install -y curl
 
 # Get Elasticsearch
 WORKDIR /usr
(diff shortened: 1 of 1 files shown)


graycodes:
The issue here seems to have been caused by the move to using zscaler. 

The fix here is likely going to be to bring in the zscaler pem and include it in this dockerfile (in addition to some others presumably). As such this can be closed for now

PR #1003, opened 2026-06-16
Add ZScaler root ca cert into ES6 dockerfile build

jfharden:
The Dockerfile in question is always built locally, and our laptops use ZScaler to MITM proxy traffic, we need to trust the ZScaler root CA cert for the build of the ES6 container.

Thanks to @graycodes and @kentsanggds for reporting the problem and prior art to fix


graycodes (approved):
Works for me running locally! 🙏🏻 thanks!

PR #1008, opened 2026-06-30
Amend replicate elasticsearch script

stringiest:
The Santa policies on the new DSIT technologist laptops block virtual machines from accessing certain directories. Changing the replicate-elasticsearch.sh script to use a local `/tmp` directory rather than system one.

I've run the amended script locally with this change and managed to restore the elasticsearch data.

### DIFF
--- a/bin/replicate-elasticsearch.sh
+++ b/bin/replicate-elasticsearch.sh
@@ -23,7 +23,7 @@ fi
 
 # temporary config file because ES needs to be configured in advance
 # for filesystem-based snapshots
-cfg_path=$(mktemp '/tmp/govuk-docker-data-sync.XXXXX')
+cfg_path=$(mktemp -d -t govuk-docker-data-sync)
 echo "
   cluster.name: 'docker-cluster'
   network.host: 0.0.0.0


bin/replicate-elasticsearch.sh
```diff
@@ -23,7 +23,7 @@ fi
 
 # temporary config file because ES needs to be configured in advance
 # for filesystem-based snapshots
-cfg_path=$(mktemp '/tmp/govuk-docker-data-sync.XXXXX')
+cfg_path=$(mktemp './tmp/govuk-docker-data-sync.XXXXX')
```
jfharden: Oh interesting, it will have access to macOS' real temp dirs which are not in /tmp.


```suggestion
cfg_path=$(mktemp -d -t govuk-docker-data-sync)
```

This will use a path of

```
/var/folders/md/j4qns4ln205gq69n96cgqsy80000gn/T/govuk-docker-data-sync.XXXXXXXXXX
```
Being a real temp folder this will be cleaned up as you would expect of temp folders

stringiest: Ooh thanks! Will change.

jfharden (approved):

PR #1010, opened 2026-07-02
Content Block Manager set default JWT_AUTH_SECRET

graycodes:
When running locally in Docker, the app requires this value to be set to something.


edavey (approved):
thank you

PR #1015, opened 2026-07-14
Fix debugging in finder-frontend

KludgeKML:
- needs bin/dev to use the procfile which controls debugging
- needs web concurrency 0 set on all environments so that two processes don't start up and clash when they create debug ports.


sairamya93 (approved):
LGTM 👍

PR #1023, opened 2026-08-05
Add 12349 as debug port for places-manager

deborahchua:
This enables VS code debugging for places-manager (see https://github.com/alphagov/places-manager/pull/1675)


KludgeKML (approved):
LGTM

PR #1024, opened 2026-08-06
Upgrade version of PostgreSQL used by Locations API

deborahchua:
Currently we get this error when trying to replicate data for locations-api:

```
pg_restore: error: unsupported version (1.16) in file header)
```


KludgeKML (approved):
LGTM

PR #1026, opened 2026-08-12
Add places-manager-app-ports service

deborahchua:
This allows us to run a version of places-manager against backing services in any environment by using port forwarding. It also reduces the need to have replicated data locally.


KludgeKML (approved):
LGTM!

PR #1028, opened 2026-08-20
Define shared WHITEHALL_ASSET_MANAGER_SHARED_API_KEY [WHIT-3751]

ChrisBAshton:
We need to set up a shared key so that some requests from Whitehall can bypass the usual access-limiting checks on assets. See https://github.com/alphagov/asset-manager/pull/2039

Jira: https://gov-uk.atlassian.net/browse/WHIT-3751


ChrisBAshton:
Superseded by https://github.com/alphagov/asset-manager/pull/2040

PR #1031, opened 2026-08-26
Specialist Publisher: migrate from Mongo to MySQL [WHIT-3578]

ChrisBAshton:
## What

Migrates Specialist Publisher from Mongo to MySQL. See https://github.com/alphagov/specialist-publisher/pull/3828

## Why

Jira: https://gov-uk.atlassian.net/browse/WHIT-3578


leenagupte (approved):
Tried this locally alongside https://github.com/alphagov/specialist-publisher/pull/3828
Worked like a charm! 🥇

PR #1034, opened 2026-09-16
Migrate to OpenSearch

koetsier:
Update the local and replication setup to support Search API using OpenSearch:

* Update docker-compose.yml to use OpenSearch instead of Elasticsearch.
* Update the replication script to use OpenSearch data from integration.


bin/replicate-opensearch.sh
```diff
@@ -70,18 +68,27 @@ indices=$(curl "http://127.0.0.1:9200/_snapshot/snapshots/$snapshot_name" | jq -
...
-    "licence-finder"
+    (map(select(startswith("govuk-"))) | sort | last)
   ] | join(",")
 ')
 
 echo "restoring indices..."
-curl -XPOST "http://127.0.0.1:9200/_snapshot/snapshots/${snapshot_name}/_restore?wait_for_completion=true" \
--H "Content-Type: application/json" \
--d @- <<EOF
+curl -XPOST "http://127.0.0.1:9200/_snapshot/snapshots/${snapshot_name}/_restore" \
```
emmalowe, line 76: Can't we still use `wait_for_completion`? Looks like it is available in open search 3.7:
https://docs.opensearch.org/3.7/api-reference/snapshots/restore-snapshot/

emmalowe (approved):
LGTM. Just added a non-blocking comment/question.

bin/replicate-opensearch.sh
koetsier: Yes we can but there's a new loop that waits for it and gives a bit of feedback when the snapshot is restoring

